PR #23, opened 2016-12-23
fix #9: implement sorting tabs by container

### LINKED ISSUE #9: Implement Container Sort Capability
Clicking the `sort` button in the containers toolbar should rearrange open tabs in order of individual containers followed by non-container tabs.

groovecoder:
The current behavior is to sort the tabs inside their window. No special logic for sorting tabs into/across separate windows.

testpilot-containers-sort-tabs


jonathanKingston:
@groovecoder - I couldn't cause an issue with multiple windows however I suspect there must be a way to cause the id's to be an issue based upon the implementation. We may have to make this window id aware in future.

PR #27, opened 2016-12-30
re-integrate ContextualIdentityService into SDK code

groovecoder:
testpilot-containers-showhide

While we could bundle a WebExtension Experiment with an Embedded WebExtension add-on via [Multiple item extension packaging](https://developer.mozilla.org/en-US/docs/Multiple_Item_Packaging), the `type 256` Experiment add-ons are not supported until Firefox 52, which doesn't hit Release until March 7, 2017.

So, this PR moves the `ContextualIdentityService` module usage back into the SDK `index.js`. It drops the `webextension-experiment/` completely, because [trying to use the `ExtensionAPI` implementation directly from SDK similarly failed](https://gist.github.com/jonathanKingston/7e132231c3257cf9194971a0e2cae70e).


groovecoder:
Found a bug: since I moved `identitiesState` to the `popup.js`, it's lost when the user closes the pop-up. However, only the [WebExtensions Tab](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/Tab) object has access to open/restore tabs with a `cookieStoreId` parameter.

So, I'm wondering if a weekend project will be to move the `identitiesState` object back to `index.js` and wrap it with something like [`react-chrome-redux`](https://github.com/tshaddix/react-chrome-redux) for managing state and UI changes between the separate contexts. Redux + React would mean adding a build chain though 😞

groovecoder:
Okay, fixed the bug by moving `identitiesState` into the SDK `index.js` module to it's preserved thru pop-up close-and-re-open.

@jonathanKingston I think this is ready for review & spot-check.

Issue #13, opened 2016-12-15
Implement tab color bars to match UX spec


johngruen:

TanviHacks:
Is this issue about the tab highlighting overline/underline?

ChenMorpheus:
@johngruen @TanviHacks 
I assumed this ticket is for the purpose Tanvi asked. Please find the attached measurement for details, and let us know if any questions, thanks.

![container-tabbar](https://cloud.githubusercontent.com/assets/18019946/21711113/f38873b8-d427-11e6-8548-9ed3783b83de.png)

jonathanKingston:
Related bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1325902

johngruen:
@jonathanKingston i think we're good here and can focus on that polish bug #96, but i'll wait til you get back

PR #50, opened 2017-01-09
Initial tab restyle work to match experiment bug #13

jonathanKingston:
This implements the same style as mentioned in the email, follow up work should be done to make it have the same style as #13

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,8 @@ const { viewFor } = require("sdk/view/core");
 const webExtension = require('sdk/webextension');
 const windows = require("sdk/windows");
 const windowUtils = require('sdk/window/utils');
+const { attachTo } = require("sdk/content/mod");
+const { Style } = require("sdk/stylesheet/style");
 
 let ContainerService =
 {
@@ -295,6 +297,9 @@ let ContainerService =
     });
 
     button.appendChild(popup);
+    let style = Style({ uri: self.data.url("chrome.css") });
+
+    attachTo(style, viewFor(window));
   },
 };
 
--- a/data/chrome.css
+++ b/data/chrome.css
@@ -1,0 +1,116 @@
+/* containers experiment */
+.tabbrowser-tab[usercontextid] {
+  background-image: linear-gradient(to right, var(--identity-tab-color) 0%, var(--identity-tab-color) 100%) !important;
+  background-size: var(--identity-stroke-background-size) !important;
+  background-repeat: no-repeat !important;
+  background-position: 0 29px !important;
+  --identity-stroke-background-size: auto 2px;
+}
+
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -2,12 +2,23 @@
 const {ContextualIdentityService} = require('resource://gre/modules/ContextualIdentityService.jsm');
 const { Cc, Ci, Cu, Cr } = require('chrome');
 
+Cu.import("resource://gre/modules/Services.jsm");
```
bakulf: This is not needed

index.js
```diff
@@ -2,12 +2,23 @@
...
 var windowUtils = require('sdk/window/utils');
 var tabsUtils = require('sdk/tabs/utils');
 
+const { browserWindows: windows } = require("sdk/windows");
+const { attachTo } = require("sdk/content/mod");
+const { Style } = require("sdk/stylesheet/style");
+let self = require("sdk/self");
+let style = Style({ uri: self.data.url("chrome.css") });
+
+for (let w of windows)
```
bakulf: 1. This should go into init()
2. use {}
3. add a comment about styling windows.

index.js
```diff
@@ -2,12 +2,23 @@
...
 
+Cu.import("resource://gre/modules/Services.jsm");
+
 const tabs = require('sdk/tabs');
 const webExtension = require('sdk/webextension');
 const { viewFor } = require("sdk/view/core");
 var windowUtils = require('sdk/window/utils');
 var tabsUtils = require('sdk/tabs/utils');
 
+const { browserWindows: windows } = require("sdk/windows");
```
bakulf: you make my life easier if you do : `var windows = require("sdk/windows");`
ad then you do:

for (let window in windows.browserWindows) {
  ...

groovecoder: Can it be `const windows = require('sdk/windows');` ? `const` is nice. 😉

index.js
jonathanKingston: Lul I copied this not my style. Moved it anyway :)

index.js
```diff
@@ -11,6 +11,8 @@ const { viewFor } = require("sdk/view/core");
 const webExtension = require('sdk/webextension');
 const windows = require("sdk/windows");
 const windowUtils = require('sdk/window/utils');
+const { attachTo } = require("sdk/content/mod");
```
bakulf: alphabetic order

index.js
```diff
@@ -295,6 +297,9 @@ let ContainerService =
     });
 
     button.appendChild(popup);
+    let style = Style({ uri: self.data.url("chrome.css") });
```
bakulf: this method is meat to be called several times for the same window. We should check that this style is not done multiple time. But this can be done in separate patch.
Please add a comment before this let style ...

PR #53, opened 2017-01-09
Let's use a DocumentFragment instead innerHTML

bakulf:
I also converted some "const" to "let" and some "foo=>" to "foo => "

### DIFF
--- a/webextension/js/popup.js
+++ b/webextension/js/popup.js
@@ -3,31 +3,37 @@ const CONTAINER_HIDE_SRC = '/img/container-hide.svg';
 const CONTAINER_UNHIDE_SRC = '/img/container-unhide.svg';
 
 function showOrHideContainerTabs(userContextId, hasHiddenTabs) {
-  return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


webextension/js/popup.js
```diff
@@ -3,31 +3,37 @@ const CONTAINER_HIDE_SRC = '/img/container-hide.svg';
...
+    hideorshowIcon.src = hasHiddenTabs ? CONTAINER_HIDE_SRC : CONTAINER_UNHIDE_SRC;
+
+    // The new identity is returned.
+    return identity;
   });
 }
 
 if (localStorage.getItem('onboarded2')) {
-  for (const element of document.querySelectorAll('.onboarding')) {
+  for (let element of document.querySelectorAll('.onboarding')) {
```
groovecoder: Why `let` instead of `const`?

bakulf: const is for immutable objects. We are not actually changing element, but to me seems wrong that we change attributes in element when it is const. Maybe because I write too much C++.

groovecoder: worksforme.

PR #56, opened 2017-01-09
eslint updated

bakulf:
What about this setup?


index.js
```diff
@@ -1,32 +1,30 @@
-const XUL_NS = "http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul";
+const XUL_NS = 'http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul';
```
jonathanKingston: I think we should probably use double to match Firefox to be honest.

PR #57, opened 2017-01-09
Add repo and fix SPDX license in package.json

pdehaan:
Sorry, also ran [**fixpack**](http://npm.im/fixpack) to make everything beautifully A-z sorted, which causes a bit of diff-churn.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,30 +1,33 @@
 {
-  "title": "Containers Experiment",
-
   "name": "testpilot-containers",
-
-  "version": "0.0.1",
-
+  "title": "Containers Experiment",
   "description": "Containers works by isolating cookie jars using separate origin-attributes defined visually by colored ‘Container Tabs’. This add-on is a modified version of the containers feature for Firefox Test Pilot.",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,30 +1,33 @@
...
   "keywords": [
     "jetpack"
   ],
-  
-  "hasEmbeddedWebExtension": true,
-
-  "devDependencies": {
-    "jpm": "^1.2.2",
-    "eslint": "^3.12.2"
+  "license": "MPL-2.0",
```
pdehaan: Was "MPL2" which npm was complaining about saying it wasn't a valid SPDX value.

package.json
```diff
@@ -1,30 +1,33 @@
...
+  "main": "index.js",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/mozilla/testpilot-containers.git"
+  },
```
pdehaan: Also added git repository, which added values like "homepage" and bug url.

PR #67, opened 2017-01-10
Tweak package.json lint script to lint add-on

pdehaan:
Noticed we added a devDep for addons-linter, but it wasn't wired to anything.
Now running `npm run lint` will run both ESLint and the addons-linter tools.

If you want to run the scripts individually, you can still use stuff like `$ npm run lint:js -- --fix` to autofix any ESLint errors/warnings.


jonathanKingston:
Yay saved me a job! Thanks

PR #82, opened 2017-01-13
Many little things

bakulf:


webextension/js/popup.js
```diff
@@ -1,150 +1,335 @@
...
 /* global browser, window, document, localStorage */
+
 const CONTAINER_HIDE_SRC = "/img/container-hide.svg";
 const CONTAINER_UNHIDE_SRC = "/img/container-unhide.svg";
 
-function showPanel(panelSelector) {
-  for (let panelElement of document.querySelectorAll(".panel")) {
-    panelElement.classList.add("hide");
+// Let's set it to false before releasing!!!
+const DEBUG = false;
```
pdehaan: Re: "Let's set it to false before releasing!!!" Do we want this to be `true` during development?
Or some flag we can maybe set in about:addons?

webextension/js/popup.js
```diff
@@ -1,150 +1,335 @@
...
 
-  // Show or not the has-tabs section.
-  for (let trHasTabs of document.getElementsByClassName("container-info-has-tabs")) {
-    trHasTabs.hidden = !identity.hasHiddenTabs && !identity.hasOpenTabs;
-    trHasTabs.setAttribute("data-user-context-id", identity.userContextId);
-  }
+  showPanel(panel, currentIdentity = null) {
+    // Invalid panel... ?!?
+    if (!(panel in this._panels)) {
+      throw("Something really bad happened. Unknown panel: " + panel + "\n");
```
pdehaan: Re: http://eslint.org/docs/rules/no-throw-literal

Do we want to throw strings or Error objects?

webextension/js/popup.js
```diff
@@ -1,150 +1,335 @@
...
+    });
+  },
 
-  // Let"s remove all the previous tabs.
-  for (const trTab of document.getElementsByClassName("container-info-tab")) {
-    trTab.remove();
-  }
+  showPreviousPanel() {
+    if (!this._previousPanel) {
+      throw "Current panel not set!";
```
pdehaan: See previous comment re: `throw "Current panel not set!"` vs `throw new Error("Current panel not set!")`.

nit: Also, in some cases we're saying `throw "string"` and in other cases we're using `throw("string")` which made searching for instances a bit tricky.

webextension/js/popup.js
```diff
@@ -1,150 +1,335 @@
...
+          Logic.showPanel(P_CONTAINER_INFO, identity);
+        }
       });
-    }
+    });
 
-    document.getElementById("container-info-table").appendChild(fragment);
-  })
+    document.querySelector(".identities-list").innerHTML = "";
+    document.querySelector(".identities-list").appendChild(fragment);
```
pdehaan: Should we cache `document.querySelector(".identities-list")` so we don't need to do a DOM lookup twice?

webextension/js/popup.js
```diff
@@ -1,150 +1,335 @@
...
+      userContextId: identity.userContextId,
+    }).then(tabs => {
+      log('browser.runtime.sendMessage getTabs, tabs: ', tabs);
+      // For each one, let's create a new line.
+      let fragment = document.createDocumentFragment();
+      for (let tab of tabs) {
+        let tr = document.createElement("tr");
+        fragment.appendChild(tr);
+        tr.classList.add("container-info-tab");
+        tr.classList.add("clickable");
```
pdehaan: Re: https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

It looks like we can pass multiple classes to the `classList` by passing multiple values:

```js
tr.classList.add("container-info-tab", "clickable");
```

webextension/js/popup.js
```diff
@@ -1,150 +1,335 @@
...
+  prepare() {
     let fragment = document.createDocumentFragment();
-    for (const tab of tabs) {
+
+    Logic.identities().forEach(identity => {
+      log('identities.forEach');
       let tr = document.createElement("tr");
       fragment.appendChild(tr);
-      tr.classList.add("container-info-tab");
+      tr.classList.add("container-panel-row");
```
pdehaan: Ref: https://github.com/mozilla/testpilot-containers/pull/82/files#r96049411

I think you can do this to add the two class names at once:
```js
tr.classList.add("container-panel-row", "clickable");
```

webextension/js/popup.js
bakulf: Yes, follow up. By default true as you suggested.

bakulf:
Thanks for the review. Comments applied.

Issue #83, opened 2017-01-13
What about if the user wants to create a container with an empty name?


bakulf:
Should we show an alert? Or we want to show a message?

ChenMorpheus:
I would propose to provide "Container 01" as default name since the user might want to quickly create a container. Besides, we'll have a blinking text cursor to remind the user inputing a name. If the user wants to create more and more containers without inputing names, we can simply add the number to "Container 02", "Container 03", etc. I have concerns that displaying an alert or a message might interrupt the user flow.  What do you think?
![2 1 container create name]

(https://cloud.githubusercontent.com/assets/18019946/21974730/89d5e7be-dc03-11e6-89fb-98691e9dd46b.png)

Issue #70, opened 2017-01-10
Add Travis or Circle CI for running tests and linting


pdehaan:
We should probably add Travis-CI or Circle-CI for building the add-on and running lint checks and running tests on each commit so we can be sure that we aren't accidentally breaking anything.

pdehaan:
Current status:

```sh
$ git rev-parse --short HEAD
3d5b580

$ npm run lint:js
> testpilot-containers@0.0.1 lint:js /Users/pdehaan/dev/github/mozilla/testpilot-containers
> eslint .

/Users/pdehaan/dev/github/mozilla/testpilot-containers/webextension/js/popup.js
   24:5   error  Unexpected console statement         no-console
  106:25  error  Strings must use doublequote         quotes
  112:68  error  Missing semicolon                    semi
  121:27  error  Expected '===' and instead saw '=='  eqeqeq
  122:41  error  Strings must use doublequote         quotes
  122:47  error  Strings must use doublequote         quotes
  200:11  error  Strings must use doublequote         quotes
  254:84  error  'e' is defined but never used        no-unused-vars
  264:82  error  'e' is defined but never used        no-unused-vars
  306:11  error  Strings must use doublequote         quotes

✖ 10 problems (10 errors, 0 warnings)
```

pdehaan:
I think the bare minimum needed (to just support per-commit linting on Travis-CI) is something like:

```yaml
language: node_js

node_js:
  - 7

script:
  - npm run lint
```

I seem to lack the appropriate permissions in this repo to enable Travis-CI (https://travis-ci.org/mozilla/testpilot-containers).

groovecoder:
We are linting with Travis CI now!

PR #84, opened 2017-01-16
Issue #83

bakulf:


webextension/js/popup.js
```diff
@@ -101,6 +101,28 @@ let Logic = {
...
+        let id = parseInt(identity.name.substr(defaultName.length))
+        if (id) {
+          ids.push(id);
+        }
+      }
+    });
+
+    // Here we find the first valid id.
+    for (let id = 1;; ++id) {
+      if (ids.indexOf(id) == -1) {
```
pdehaan: Should probably be `===`. In fact, shouldn't this have failed ESLint linting, per [/.eslintrc.js:9](https://github.com/mozilla/testpilot-containers/blob/3d5b58078db1ddc5635a3d46bf3a17399472ae5b/.eslintrc.js#L9)?

pdehaan: Also, isn't there a slight possibility of a never-ending loop since the `return` statement is in an `if` block?

webextension/js/popup.js
```diff
@@ -101,6 +101,28 @@ let Logic = {
...
   },
+
+  generateIdentityName() {
+    const defaultName = 'Container';
+    let ids = [];
+
+    // This loop populates the 'ids' array with all the already-used ids.
+    this._identities.forEach(identity => {
+      if (identity.name.startsWith(defaultName)) {
+        let id = parseInt(identity.name.substr(defaultName.length))
```
pdehaan: Nit: Add `radix` argument? http://eslint.org/docs/rules/radix
Probably only an issue if the `name` starts with a `0` though, so possibly pretty edge case.

webextension/js/popup.js
pdehaan: > Should probably be `===`. In fact, shouldn't this have failed ESLint linting, per /.eslintrc.js:9?

Actually, it did cause an ESLint error. Ref #70

PR #88, opened 2017-01-16
Add 'npm run once' script

pdehaan:
Add `$ npm run once` script which calls `jpm run` and launches an Aurora/DeveloperEdition browser with a fresh profile.

If you want to use a pre-built profile, just add `$ npm run once -- -p addon_dev`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
   "scripts": {
     "lint": "npm-run-all lint:*",
     "lint:js": "eslint .",
-    "lint:addon": "addons-linter webextension --self-hosted"
+    "lint:addon": "addons-linter webextension --self-hosted",
+    "once": "jpm run -b Aurora"
   }
 }


bakulf:
can we use beta instead?

pdehaan:
> can we use beta instead?

I've never actually gotten this add-on to work on Beta.
If I use `$ ./node_modules/.bin/jpm run -c beta`, I get the following:

add-ons_manager

blank_skitch_document

bakulf:
Right, this is because beta has a security check on signed/unsigned addons.
In order to use beta you have to download an unbranded version of beta:
https://wiki.mozilla.org/Add-ons/Extension_Signing#Unbranded_Builds

bakulf:
Probably it's better to don't land this patch until we decide to support aurora fully.

pdehaan:
Yeah, I'm pretty reluctant to download special unbranded builds to test things, if I can avoid it.
I'd hate to test the add-on in a special mystery build, then we ship the experiment and discover that it works for 0% of the users who are using the vanilla release/beta channels.

jonathanKingston:
Unfortunately @pdehaan that is the way that I think we are all testing it, the only other way would be to build the beta branch yourself too. Both methods are discussed in the README. We will likely need to fix things specifically for 52+ due to patches that have landed in central since then.

groovecoder:
Is this meant to be an aid to QA or a more general aid to anyone who wants to run the add-on?

We should add docs that say **NOT** to use this `jpm run once` for QA purposes, since it will run against 52 (Aurora) which is **NOT** what we're releasing into.

Or we can just close this PR to leave it out completely until we're off the un-branded Beta release.

PR #90, opened 2017-01-17
Promise eslint errors fixed

bakulf:


bakulf:
@pdehaan can you take a look?

pdehaan:
I think this looks fine.
Not sure if we eventually want to add some logging for the errors so we can detect when we're failing... Or if we send them through to Sentry or something. But 👍 for merging this.

PR #91, opened 2017-01-18
many things, all together in 1 big PR

bakulf:


pdehaan:
I had two tabs (hbo.com and cbs.com) and I moved those "Personal" container tabs into a new window. When they reopened, both were about:blank.

pdehaan:
Also seeing an issue where I have a "Work" container tab with 1 tab, http://aol.com. When I hide all the tabs in that container group, then move those hidden tabs to a new window, nothing happens. I think I was expecting all the hidden tabs to become visible again if I moved them to a new window.

bakulf:
This second comment is something that has to be implemented.
About the moving, it works fine locally, I'll try it with aurora/nightly as well.

bakulf:
Can you tell me if you see any error? Errors are shown on stderr (the terminal) and/or in the browser console. Thanks

PR #121, opened 2017-01-27
Fix the moving of tabs in nightly

bakulf:


index.js
```diff
@@ -384,16 +384,11 @@ const ContainerService = {
...
           // Let's move the tab to the new window.
           for (let tab of list) { // eslint-disable-line prefer-const
-            const newTab = newBrowserWindow.gBrowser.addTab("about:blank");
-            newBrowserWindow.gBrowser.swapBrowsersAndCloseOther(newTab, viewFor(tab));
-            // swapBrowsersAndCloseOther is an internal method of gBrowser
-            // an it's not supported by addon SDK. This means that we
-            // don't receive an 'open' event, but only the 'close' one.
-            // We have to force a +1 in our tab counter.
-            ++this._identitiesState[args.userContextId].openTabs;
+            newBrowserWindow.gBrowser.adoptTab(viewFor(tab), pos++, false);
```
groovecoder: Nice ... where can I look up `adoptTab` to learn more about it?

bakulf: The logic of tab browsing in firefox is implemented using XBL. XBL is a magic internal-only technology (at the moment) that allows you to create new XUL elements and define custom methods for them.
In particular, if you check how tabs are implemented, you will find that, there is a magic object called gBrowser. This is the tabbrowser object defined here:
https://dxr.mozilla.org/mozilla-central/source/browser/base/content/tabbrowser.xml#12
If you want to know more, ping me on IRC.

Issue #123, opened 2017-01-28
shouldn't allow containers being opened within PB


kjozwiak:
We shouldn't be allowing containers to be opened within PB..

STR (assuming you've already installed the testing environment):

* launch PB via Hamburger Menu -> New Private Window
* click on the "Container" icon and open several default containers

You'll notice that they're being opened within the PB window which shouldn't be allowed as we block containers from being opened within PB everywhere else.. This also creates several UX/UI issues as demonstrated in the second screenshot.

<img width="1280" alt="pb" src="https://cloud.githubusercontent.com/assets/2602313/22394949/ba669f48-e4fc-11e6-92b1-e249a9ff6c17.png">

<img width="1275" alt="dropdownissue" src="https://cloud.githubusercontent.com/assets/2602313/22394973/6fd7a00c-e4fd-11e6-831e-91798a93a745.png">

bakulf:
@TanviHacks, @johngruen Do we want to disable the popup button or we want to show a panel saying: containers are disabled on PB?

bakulf:
I wrote a patch to disable the addon in privateBrowsing. The button is not clickable and the popup is not shown.

PR #124, opened 2017-01-28
Disable popup if in PrivateBrowsing

bakulf:


bakulf:
This is about issue #123

jonathanKingston:
@bakulf does this work in 'always private mode'?
Recent events in a certain country has lead me to suggest everyone should use this mode.

webextension/background.js
```diff
@@ -0,0 +1,13 @@
+browser.tabs.onCreated.addListener(tab => {
+  if (tab.incognito) {
+    browser.browserAction.disable(tab.id);
```
groovecoder: let's add `browser.browserAction.setTitle("Containers disabled in Private Browsing Mode")` so we can show the user what's happening via the tooltip.

groovecoder:
And yes - I verified that this disables the button for users who are in "always Private Browsing" mode. (I.e., "Never remember history" in Firefox settings).

nightly-always-private

jonathanKingston:
(didn't finish posting this comment until now)
@groovecoder thats a shame :(.
Hopefully the change to enable in that mode won't be too complex. We discussed this for central and said we would do this. 
- I will try and fix in a follow up

bakulf:
PR #133 is about it.

Issue #75, opened 2017-01-11
"Error: this._identitiesState[identity.userContextId] is undefined" after adding custom container


pdehaan:
Found in a6bfaab (via `$ git rev-parse --short HEAD`)

---

### Steps to reproduce:
1. Open the container tabs doorhanger. Everything is kosher.
1. Open the container tabs doorhanger and select "Edit Containers".
1. Add a custom container (mine is named "pppp" and uses default blue fingerprint icon).
1. Open Browser Console and delete any existing logs.
1. Open the container tabs doorhanger.

### Actual results:
- No containers are displayed (only heading and Edit Containers button).
- Error in Browser Console:
  > Error: this._identitiesState[identity.userContextId] is undefined (unknown)
  
  My best (and often horribly mis-informed) guess is that this error is coming from [/index.js:47](https://github.com/mozilla/testpilot-containers/blob/8bcfae7bed536788457b5f9f389607667bc1822c/index.js#L47):

  ```
    // Map of identities.
    ContextualIdentityService.getIdentities().forEach(identity => {
      this._identitiesState[identity.userContextId] = {
        hiddenTabUrls: [],
        openTabs: 0
      };
    });
  ```

---

![preferences_and_testpilot-containers_index_js_at_8bcfae7bed536788457b5f9f389607667bc1822c_ _mozilla_testpilot-containers_and_1__go_to_https___discourse_mozilla-community_org_c_ _blok_and_index2_js_ _extract-xpi](https://cloud.githubusercontent.com/assets/557895/21830993/78e41a28-d755-11e6-9d3e-3f824c761d3a.png)

pdehaan:
... and after deleting my custom container, everything looks/acts normal again:

![preferences_and_testpilot-containers_index_js_at_8bcfae7bed536788457b5f9f389607667bc1822c_ _mozilla_testpilot-containers_and_1__go_to_https___discourse_mozilla-community_org_c_ _blok_and_index2_js_ _extract-xpi](https://cloud.githubusercontent.com/assets/557895/21831025/b1fd5e0a-d755-11e6-9c1f-f0eb14d1f20a.png)

bakulf:
It's important to fix this bug, but let's remember that nightly is not supported

pdehaan:
@bakulf Ah, good tip. Maybe I should re-read the README.
What Firefox channels are currently supported? Should I only be testing in DevEdition?

bakulf:
[...]

PR #126, opened 2017-01-31
Remap tabs in case a userContextId is unknown #75

bakulf:


index.js
```diff
@@ -216,6 +205,25 @@ const ContainerService = {
...
   },
 
+  _createIdentityState() {
+    return {
+      hiddenTabUrls: [],
+      openTabs: 0
+    };
+  },
+
+  _maybeRemapTabs(userContextId) {
```
groovecoder: Can we change to `_remapTabsIfMissing` so it's more obvious what the method is doing everywhere it's called?

groovecoder:
non-blocking comment: `showTabs`, `hideTabs`, and `getTabs` resolve with different types after the remap (`null` for `showTabs` and `hideTabs`, `[]` for `getTabs`).

Is it worth changing their code (and calling code) to resolve with the same type? Would that cut down on any code duplication?

bakulf:
They have different types... We can create a method, but we don't cut down so much code.

PR #133, opened 2017-02-01
Addon enabled when PB is by default

bakulf:


groovecoder:
* Add-on is working when I set Firefox to "Never remember history"
* Add-on is working when I set Firefox to "Remember history"
* Add-on is disabled when I set Firefox to "Remember history" **IN A PRIVATE BROWSING WINDOW**

Interestingly, when I set Firefox to "Never remember history" and open a "New Private Window" - it looks exactly the same? Is that expected? All (i.e., normal and PBM) windows look the same when "Never remember history" is set?

If so, r+

webextension/background.js
```diff
@@ -1,14 +1,23 @@
...
-browser.tabs.query({}).then(tabs => {
-  for (let tab of tabs) { // eslint-disable-line prefer-const
-    if (tab.incognito) {
-      disableAddon(tab.id);
-    }
+    browser.tabs.query({}).then(tabs => {
+      for (let tab of tabs) { // eslint-disable-line prefer-const
+        if (tab.incognito) {
+          disableAddon(tab.id);
+        }
```
groovecoder: Are these the only 2 places we're doing this `if (tab.incognito) { disableAddon(tab.id); }` logic? If we're doing it in more places, maybe we should refactor into a helper function?

Issue #130, opened 2017-01-31
old/new container menu from "+" new tab button overlapping


kjozwiak:
Because we're setting ["privacy.usercontext.longPressBehavior;1"](https://github.com/mozilla/testpilot-containers/blob/86e3cb48869c8115918db7b5cd160ebee843faa5/index.js#L55), you'll run into situations where you can have two drop down menu's appear from the "+" new tab button. We should be disabling the long press and keep the "hover" functionality that's currently being used.

<img width="458" alt="twodropdowns" src="https://cloud.githubusercontent.com/assets/2602313/22481078/d3fa52d4-e7c1-11e6-85a8-5332fd5edf35.png">

STR:

* launch fx (assumes you have the test pilot installed)
* long press the "+" new tab button and the old container menu will appear
* scroll through the containers in the old menu, and the new container menu will appear as per the above screenshot

groovecoder:
Long-press doesn't exist in 51, right? So I assume this is a 52-only bug?

kjozwiak:
> Long-press doesn't exist in 51, right? So I assume this is a 52-only bug?

Yup, it's only affecting fx52 and not fx51.. However, it's still occurring in fx51, commented [here..](https://github.com/mozilla/testpilot-containers/pull/135#issuecomment-277355641)

PR #135, opened 2017-02-01
Long press

bakulf:


bakulf:
Issue #130

groovecoder:
Can't reproduce the original issue on 51; is this a 52-only bug?

bakulf:
Yes, it's only in nightly.

groovecoder:
The code here looks good. @pdehaan or @kjozwiak - can either of you spot-check on 52? I'm keeping my setup with 51 while I code to stick with the release experience.

kjozwiak:
> Long-press doesn't exist in 51, right? So I assume this is a 52-only bug?

This was only affecting fx52.. I couldn't reproduce it while using fx51..

@bakulf, it looks like this is still broken in fx52.. I double checked and made sure I was running the latest code that included your fix:

> kjozwiak:testpilot-containers kjozwiak$ git log
> commit 227dbb02f1f52f8476deb679a6410cecc780b6dd
> Merge: b406557 33b9881
> Author: Andrea Marchesini <baku@ippolita.net>
> Date:   Fri Feb 3 00:13:49 2017 +0100
> 
> Merge pull request #135 from bakulf/longPress

I checked my index.js file and it looks like the pref was removed..

code

I also checked about:config and couldn't find `privacy.usercontext.longPressBehavior` anywhere... So I'm not really sure why this is still reproducible under fx52 :/

config

Thoughts?

Issue #114, opened 2017-01-26
disabling/enabling containers via about:preferences#privacy when test pilot installed


kjozwiak:
Currently, when you disable containers via about:preferences#privacy, you'll still be able to create containers with the test pilot, however, they'll behave like regular tabs even though they're decorated as containers.

STR: (assumes you have the test pilot installed)

* disable containers via about:preferences#privacy
* open a new personal container via the "Containers" button in the toolbar
* visit any website that uses cookies (FB, Twitter etc..)
* under about:preferences#privacy, look at "remove individual cookies" and you'll notice they're loading within the default context even though the tabs are decorated as containers

We can either:

* remove the enable/disable feature under about:preferences#privac for the test pilot
* make disable/enable work with the test pilot

TanviHacks:
The addon should set privacy.userContext.ui.enabled to false.

PR #148, opened 2017-02-03
about:preferences#privacy should not show Containers UI - Issue #114

bakulf:


groovecoder:
Easy game - does what it says on the tin.

Issue #151, opened 2017-02-03
Better behavior when user hides or deletes all open container types 


johngruen:
## STR
* Open only container tabs open in one container
* Close all non-container tabs
* Hide the container

## Current
* On MacOS, Window closes
* On Win/Linux FF closes

## Expected
New default container tab opens to users's home page

PR #154, opened 2017-02-06
Tab created when all the others are closed - issue #151

bakulf:


index.js
```diff
@@ -229,6 +229,29 @@ const ContainerService = {
...
+    if (tabs.length !== tabsToClose.length) {
+      return Promise.resolve(null);
+    }
+
+    const browserWin = windowUtils.getMostRecentBrowserWindow();
+
+    // This should not really happen.
+    if (!browserWin || !browserWin.gBrowser) {
+      return Promise.resolve(null);
+    }
```
groovecoder: This check is duplicated in `openTab` - can it be turned into a helper function?

index.js
```diff
@@ -241,26 +264,26 @@ const ContainerService = {
...
+      // This tab is going to be closed. Let's mark this tabObject as
+      // non-active.
+      object.active = false;
 
-        this._identitiesState[args.userContextId].hiddenTabUrls.push(object);
-        tab.close();
+      getFavicon(object.url).then(url => {
+        object.favicon = url;
+      }).catch(() => {
+        object.favicon = "";
```
groovecoder: Not new behavior in this patch, but why are we assigning the `object.favicon` here before we hide/close the tabs? We already get the favicon in `getTabs` right?

bakulf: You are right, but in theory, from an index.js point of view, getTabs cannot be called yet.
Plus, it seems that getFavicon is almost noop if the icon is already known.

PR #161, opened 2017-02-08
Add some filenames to .jpmignore

bakulf:


.jpmignore
```diff
@@ -11,3 +11,7 @@ node_modules/
 .stylelintrc
 .travis.yml
 testpilot-containers.xpi
+.vimrc
+.DS_Store
+webextension/.DS_Store
```
groovecoder: nit: this is an extra line - the above `.DS_Store` line will ignore all `.DS_Store` files.

PR #166, opened 2017-02-10
Fixing issues for edit screen on Linux. Fixes #157

### LINKED ISSUE #157: Edit container looks broken
screenshot

This is what I see locally. Fresh profile, current master.

jonathanKingston:
This should be checked on other platforms other than Linux. I really don't like the solution used however I don't know of anything that would work as image replacement hacks don't work and hiding the inputs makes the accessibility benefits of using radio's go away.

### DIFF
--- a/webextension/css/popup.css
+++ b/webextension/css/popup.css
@@ -176,8 +176,7 @@ table.unstriped tbody tr {
 
 .edit-container-panel [type="radio"] {
   display: inline;
-  margin-inline-start: -17px;
-  text-indent: -999999px;
+  opacity: 0;
 }
 
 .edit-container-panel [type="radio"]:first-child {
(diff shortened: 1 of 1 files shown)


groovecoder:
Now it looks a little funky on Mac?

edit-containers-alignment

webextension/css/popup.css
```diff
@@ -181,7 +179,8 @@ table.unstriped tbody tr {
   margin-block-start: 0;
   margin-inline-end: 0;
   margin-inline-start: 0;
-  offset-inline-start: -18px;
+  offset-block-start: -5px;
```
groovecoder: This is the line that causes the jank on Mac OS. 😢

jonathanKingston: Is there a way to get z-index to work or something here to hide them?

jonathanKingston:
/CC @groovecoder this is something I would love your time on as you have more experience with popups

jonathanKingston:
@groovecoder this should solve the issues mentioned 👍

Issue #184, opened 2017-02-14
currently using about:blank rather than about:newtab when opening container tabs


kjozwiak:
When you open container tabs, you'll currently get a blank page. We should be loading about:newtab rather than about:blank when opening container tabs.

CC'ing @jonathanKingston as per IRC :)

<img width="1287" alt="screen shot 2017-02-14 at 1 58 28 pm" src="https://cloud.githubusercontent.com/assets/2602313/22944218/f5ba6542-f2bd-11e6-9385-79ac12d3976d.png">

PR #185, opened 2017-02-14
Setting default tab page to about:newTab. Fixes: #184

jonathanKingston:


index.js
```diff
@@ -504,9 +505,9 @@ const ContainerService = {
         userContextId = args.userContextId;
       }
 
-      const tab = browserWin.gBrowser.addTab(args.url || null, { userContextId });
+      const tab = browserWin.gBrowser.addTab(args.url || DEFAULT_TAB, { userContextId });
       browserWin.gBrowser.selectedTab = tab;
-      browserWin.focusAndSelectUrlBar();
+      browserWin.focusUrlBar();
```
groovecoder: We decided to stop selecting the url bar? 😢

bakulf: I prefer browserWin.focusAndSelectUrlBar(), plus it's not related with this PR. Separate commit? separate PR?

jonathanKingston: Wait why would you want that? Anyway I did this as with the `newTab` it showed and was selected.

jonathanKingston: Pushed fix anyway.

PR #188, opened 2017-02-15
Popup hover and clickable area changes. Fixes #120

### LINKED ISSUE #120: Change container UI when hovering over the '>' arrow
When I hover over the '>' arrow, remove the '+' icon on the row and highlight the '>' somehow, so I know that clicking on the '>' is a different action than '+'.

jonathanKingston:


groovecoder:
Note: this implements @youwenliang direction - i.e., have to click on the '+' to get a new tab, and the rest of the container row in the pop-up drills down into the container details panel.

In a meeting @TanviHacks said we should create a new tab when the user clicks anywhere on the row except the '>' icon. But now that I've used the add-on a while, I've discovered that I use the tab-bar "+" way more often than I want to use the pop-up to create tabs, and I use the pop-up to hide/show/move containers.

So, now I like this UX and implementation much more.

Issue #178, opened 2017-02-13
Containers created by addons when the addon is disabled?


bakulf:
I see 2 options:

1. We restore the previous containers
2. We just remove all the containers, and reintroduce the default 4 ones?

I vote for 2 (it's easier). Of course this is an issue just for nightly users.

ChenMorpheus:
I would prefer the first one since users might not think "disable" is equal to "erase all data." If we do go the second one, I would recommend prompting warning message when users turn off the add-on, showing something like "this will remove all relevant data, are you sure to disable the add-on?"

PR #189, opened 2017-02-15
fix #174: remove add-on containers on uninstall

### LINKED ISSUE #174: cookies not being removed from cookies.sqlite when removing container add-on
We're not removing container cookies when the feature is disabled or removed via:

* disabled using `privacy.userContext.enabled;true` under about:config
* removing the container XPI via about:addons (containers_experiment-0.0.2-fx.xpi)

When someone removes/disables the container feature, the cookies are still being stored within cookies.sqlite and can be re-used. If someone re-installs/re-enables the container feature, they could potentially use an already established session. I've reproduced a similar situation using facebook.com as described below:

* install fx51.0.1 or fx52.0b5 with the latest container XPI (used containers_experiment-0.0.2-fx.xpi in this instance)
* once installed, login into facebook.com using a personal container
* remove the container XPI via about:addons under "Extensions"
* disable the container feature via about:config (see #173)
* restart fx (containers should be completely removed/disabled)
* inspect cookies.sqlite and you'll notice that the cookies are all still present (used SQLite Manager)
* re-install the same container XPI
* open a personal container and visit facebook.com

You'll still be logged into the same account as before as the cookies haven't been removed from the sqlite database when the container XPI was initially removed.

CCing @TanviHacks & @bakulf. I've also created [bug#1338735](https://bugzilla.mozilla.org/show_bug.cgi?id=1338735) to get this fixed in the nightly implementation of containers.

groovecoder:
I'm sure this isn't the right/best way to do this(?) - using `Ci.nsICookieManager2` directly, but we need to get these MVP bugs fixed.

@bakulf @jonathanKingston - what do you think?

### DIFF
--- a/index.js
+++ b/index.js
@@ -66,10 +66,16 @@ const ContainerService = {
     // uninstallation. This object contains also a version number, in case we
     // need to implement a migration in the future.
     if (installation) {
+      let preInstalledIdentities = []; // eslint-disable-line prefer-const
+      ContextualIdentityService.getIdentities().forEach(identity => {
+        preInstalledIdentities.push(identity.userContextId);
+      });
+
       const object = {
         version: 1,
         prefs: {},
         metricsUUID: uuid.uuid().toString(),
+        preInstalledIdentities: preInstalledIdentities
       };
 
       PREFS.forEach(pref => {
@@ -850,9 +856,6 @@ const ContainerService = {
       CustomizableUI.createWidget(widget);
     }
 
-    // Let's delete the configuration.
-    delete ss.storage.savedConfiguration;
-
     for (let window of windows.browserWindows) { // eslint-disable-line prefer-const
       this._getOrCreateContainerWindow(viewFor(window)).shutdown();
     }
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -840,6 +842,15 @@ const ContainerService = {
     // all the configuration must go away now.
     this._windowMap = {};
 
+    // Remove all cookies with origin attributes
+    ContainerService.queryIdentities().then(identities => {
+      for (const identity of identities) {
+        CookieManager.removeCookiesWithOriginAttributes(JSON.stringify({userContextId: identity.userContextId}));
```
bakulf: The issue here is that we are not removing the identities at all. When an identity is deleted, the cookies are removed.
What we should do here is to remove all the identities or 0, or some of them based on the version of FF and the previous configuration of containers (before the installation of the addon). See #178 .

groovecoder:
Updated to completely remove containers that did not exist when the add-on was installed.

index.js
```diff
@@ -73,6 +73,12 @@ const ContainerService = {
       });
 
       ss.storage.savedConfiguration = object;
+
+      let preInstalledIdentities = []; // eslint-disable-line prefer-const
+      ContextualIdentityService.getIdentities().forEach(identity => {
+        preInstalledIdentities.push(identity.userContextId);
+      });
+      ss.storage.preInstalledIdentities = preInstalledIdentities;
```
bakulf: can you move this into savedConfiguration?

groovecoder:
Moved the `preInstalledIdentities` into `savedConfiguration` object.

index.js
```diff
@@ -66,10 +66,16 @@ const ContainerService = {
...
+      let preInstalledIdentities = []; // eslint-disable-line prefer-const
+      ContextualIdentityService.getIdentities().forEach(identity => {
+        preInstalledIdentities.push(identity.userContextId);
+      });
+
       const object = {
         version: 1,
         prefs: {},
         metricsUUID: uuid.uuid().toString(),
+        preInstalledIdentities: preInstalledIdentities
```
bakulf: preInstalledIdentities

PR #190, opened 2017-02-15
Fixing osx menus. Also refresh context and file menus when a custom c…

### LINKED ISSUE #127: New Container Tab broken under file menu under OSX
@bakulf, it looks like commit 5fbc42a2888bb04b22201d7d7390ab85ca85e6e8 has regressed the containers access point via the file menu. 

**STR under fx51:**

* close all the currently opened windows
* while FX is still running, select File -> New Container Tab -> and select any container
* once the new window appears with the container, select File -> New Container Tab

**STR under fx52:**

* once fx is running, click "File -> New Container Tab"

Bisect results:

commit 5fbc42a2888bb04b22201d7d7390ab85ca85e6e8
Author: baku <baku@mozilla.com>
Date:   Mon Jan 23 09:34:19 2017 +0100

    File Menu updated

:100644 100644 01a4afc41d7ebe68d387ee29f9d03de889230431 52685930fea666f619fa3a30512799ae9214cc7f M	index.js

jonathanKingston:
…ontainer is created. Fixes #127

### DIFF
--- a/index.js
+++ b/index.js
@@ -1027,62 +1027,63 @@ ContainerWindow.prototype = {
   // Generic menu configuration.
   _configureMenu(menuId, excludedContainerCb, clickCb, arrayName) {
     const menu = this._window.document.getElementById(menuId);
+    this._disableElement(menu, arrayName);
     if (!this._disableElement(menu, arrayName)) {
-      return Promise.resolve(null);
+      // Delete stale menu that isn't native elements
+      while (menu.firstChild) {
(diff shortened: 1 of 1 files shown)


jonathanKingston:
We didn't have a bug for the custom context menu issue. It was caused yesterday by the uninstall work @bakulf worked on.

The OSX issue which I completely forgot from dealing with the native menus is that they don't like to be updated async so updating from promises on the event just isn't possible from all the testing I have done.

PR #191, opened 2017-02-16
Uninstall events

jonathanKingston:


jonathanKingston:
@groovecoder also includes PR for https://github.com/mozilla/testpilot-containers/pull/166 by mistake, please review that first, if there are issues I can create a PR without that.

When I moved the event handlers to handle event I realised all the add button code was much more simplified so made the neatening commit to make it super consistent rather than rely on magic keys everywhere.

PR #193, opened 2017-02-16
Adding new tab button hover timeout. Fixes #111

### LINKED ISSUE #111: possibly adding a longer hover delay for the plus panel button
We should look into possibly adding a longer hover delay for the plus panel button, or look into other possible solutions if the delay isn't viable. When you're closing tabs, the mouse will quickly hover over the "+" button and will popup the submenu for a few seconds..

Example of the issue:
* https://youtu.be/PDfLZL52VAY

jonathanKingston:


groovecoder:
Looks good!

Issue #181, opened 2017-02-14
Open Container Tab & Arrow Down menu's using older UI


kjozwiak:
There are two container access points that are currently still following the old UI as pictured below. They're also not displaying the icons for custom containers.

We can either:
* remove these two access points when someone installs the container test pilot
* make the two menu access points use the same UI as other container access points 

**_Open Container Tab under Hamburger -> Customize:_**

<img width="183" alt="screen shot 2017-02-14 at 11 05 15 am" src="https://cloud.githubusercontent.com/assets/2602313/22937439/9ed05154-f2a6-11e6-94cd-53f9fc324a04.png">

**_Arrow down menu when there's enough tabs opened:_**

<img width="442" alt="screen shot 2017-02-14 at 11 04 48 am" src="https://cloud.githubusercontent.com/assets/2602313/22937505/c7fc4844-f2a6-11e6-831b-e9a11b728a2f.png">

bakulf:
This bug is fixed.

PR #200, opened 2017-02-17
Customizable Container panel #181

bakulf:


jonathanKingston:
@bakulf please fix the linting errors that travis has picked up but yeah looks fine to me 👍

Issue #179, opened 2017-02-14
listing tabs via ">" doesn't always work when moving container tabs around


kjozwiak:
When you move a container tab into it's own separate window, and than reattach it back into the original window, the ">" next to that particular container will be greyed out even though there's a tab opened that's using that particular container.

Reproduced using the following:
* latest commit 3a2e351661c4a4968fa142ee3afc4bea262f1a04
* containers_experiment-0.0.3-fx.xpi

STR: (assumes you have the latest test pilot installed)

> Example: https://youtu.be/fIDd4lfyumg

* open a personal container and load any website, I used reddit.com
* detach the personal container into it's own separate window
* reattach the personal container into the original window
* click on the "Container" button and attempt clicking in ">"

groovecoder:
@bakulf I was able to fix this by removing this check from `_remapTabsIfMissing`:
```
if (userContextId in this._identitiesState) {
      return;
}
```
But are we worried that iterating over tabs will cause performance issues?

bakulf:
If you remove those tabs:
1. hidden tabs are gone - this in theory should not be a problem per-se, because we should not have hidden tabs when we do this operation.
2. we are hiding the 'real' bug: why don't we have the right openedTabs value in this._identitysState[usercontextId] ? It seems that we receive a 'close' tab notification and not the 'open' tab. This could explain why openedTab is == 0. I would debug the correct sequence of close/open events.

groovecoder:
@bakulf: I think we're missing this event because it's at the platform level? I.e., Dragging the tab to create the window isn't picked up by our ContextualIdentity module?

bakulf:
I debugged this issue a bit and the issue is this: when a tab is moved to a new window, we correctly receive a 'open' tab event and then a 'close' tab event. But the opened tab, before the swapping of the docShell doesn't have userContextId set, so, for us, that is not a container tab.
I'm working on it.

PR #202, opened 2017-02-17
Remap tabs when open/close/activated #179

bakulf:


jonathanKingston:
Code looks fine, have not tested etc but logic seems sound. @bakulf is the issue that at the time of the tab events like on open the usercontext-id isn't assigned to the tab?

Is there any certainty `this._getUserContextIdFromTab` actually fixes this issue or does it just increase the time between a race condition happening?

jonathanKingston:
Yeah so this change actually just delays the call to `this._getUserContextIdFromTab`  but `restyleTab` in `tabs.on("open", tab => {` calls this also ` const userContextId = ContainerService._getUserContextIdFromTab(tab);`

Isn't this likely to cause the same issues with restyling? Assuming this was the actual bug you mentioned here: https://github.com/mozilla/testpilot-containers/issues/179#issuecomment-280594828

bakulf:
The issue is about timing: We receive:
- open(UCI=0) - when the new tab is created after the dragging of the UCI=1 tab
- close(UCI=1) - the tab in the original window is closed
---- here the docShell is swapped between ----
- active(UCI=1) - finally we have something useful to use

index.js
```diff
@@ -804,6 +811,8 @@ const ContainerService = {
       const indicator = viewFor(tab.window).document.getElementById("userContext-indicator");
       indicator.setAttribute("data-identity-icon", identity.image);
       indicator.style.listStyleImage = "";
+    }).then(() => {
```
bakulf: This is not part of the commit. sorry.

jonathanKingston: P sure this is in master from the previous merge no worries.

jonathanKingston:
Cool LGTM then thanks for explaining 👍

jonathanKingston:
@bakulf I'm afk from now though sorry didn't sign off irc

PR #212, opened 2017-02-18
Banking to Finance

### LINKED ISSUE #7: Default Containers: change "Banking" to "Finance"
The ones we have are probably good, but I'd like to touch base our UR team to determine what these should be.

bakulf:
Fixes #7

### DIFF
--- a/index.js
+++ b/index.js
@@ -77,6 +77,17 @@ const ContainerService = {
       });
 
       ss.storage.savedConfiguration = object;
+
+      // Maybe rename the Banking container.
+      if (prefService.get("privacy.userContext.enabled") !== true) {
+        const identity = ContextualIdentityService.getIdentityFromId(3);
+        if (identity && identity.l10nID === "userContextBanking.label") {
+          ContextualIdentityService.update(identity.userContextId,
+                                           "Finance",
+                                           identity.icon,
+                                           identity.color);
+        }
+      }
     }
 
     // Enabling preferences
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -77,6 +77,17 @@ const ContainerService = {
       });
 
       ss.storage.savedConfiguration = object;
+
+      // Maybe rename the Banking container.
+      if (prefService.get("privacy.userContext.enabled") !== true) {
+        let identity = ContextualIdentityService.getIdentityFromId(3);
```
pdehaan:
```
/home/travis/build/mozilla/testpilot-containers/index.js
   89:13  error    'identity' is never reassigned. Use 'const' instead  prefer-const
```

index.js
```diff
@@ -77,6 +77,17 @@ const ContainerService = {
...
+        if (identity && identity.l10nID == "userContextBanking.label") {
```
pdehaan:
```
/home/travis/build/mozilla/testpilot-containers/index.js
   90:41  error    Expected '===' and instead saw '=='                  eqeqeq
```

index.js
pdehaan: How permanent is this renaming code? Is this only needed for Dev/QA because we were all testing beta builds, and new Test Pilot users would never have this issue because they would never have had a "Finance" container, or is this because users may have already been using containers in Nightly?

bakulf: The idea here is to rename Banking to Finance in any release where containers are not already enabled.

Issue #183, opened 2017-02-14
restoring container tabs after the add-on has been removed


kjozwiak:
Once you've removed the containers add-on from `about:addons`, you'll still be able to restore the container tabs that have been closed when the add-on was removed via `History -> Recently Closed Tabs -> Restore all tabs`. Even though the container feature has been disabled, the tabs that were restored will still work and be decorated as containers.

> Example: https://youtu.be/48tkp-eS_w4

STR: (assumes you have the add-on installed in the latest fx release)

* open three different container tabs and load random website in each container
* visit about:addons and select "Extensions"
* remove the "Containers Experiment" add-on
* visit about:config and ensure `privacy.userContext.enabled;false` & `privacy.usercontext.about_newtab_segregation.enabled;false`
* click on History -> Recently Closed Tabs -> Restore all tabs

The three container tabs that were previously closed when the add-on was removed, will be restored and will use the container UI. If you load a website within one of those restored containers, you'll notice that the cookies are still using OA's, even though containers has been removed.

jonathanKingston:
This is reproducible with normal platform containers also in release.

STR:
1. Enable containers in about:config
2. Create a tab, go to a page other than new tab
3. Close tab
4. Disable containers
5. Use history menu to reopen tab

Being that we would have to work round this platform bug does this need to be MVP @kjozwiak?

groovecoder:
Agreed - I don't know how many platform-level bugs we can fix over in the add-on. :/ It sucks that we may bring more attention to these bugs, but maybe we should wait to see if users report the same bug to know how many users are hitting it?

bakulf:
https://bugzilla.mozilla.org/show_bug.cgi?id=1340450
But we still need to fix it in the addon.

kjozwiak:
Thanks for filing the bug @bakulf... Agreed, I think we should fix this within the add-on as well. We might run into more serious issues (crashes) if someone ends up restoring a container tab while the feature has been disabled.

PR #215, opened 2017-02-19
Fix the restoring container tabs after the add-on has been removed #183

bakulf:


bakulf:
Luke, here a new patch. Thanks for testing this issue! and sorry for pushing this patch. I didn't want to have code landed too close to the release. Better to land everything today/tomorrow max.

bakulf:
Note that SessionStore can fail if the tab doesn't have any content loaded. This is why I use try/catch.

groovecoder:
Yup, this does it.

PR #216, opened 2017-02-19
tidy up tab dropdown menupanels

johngruen:
Fixes #180

PR #217, opened 2017-02-20
Custom tab rendering fix

### LINKED ISSUE #150: visual issues with custom container tabs
There's some visual issues when creating customer containers. When I initially create a customer container, the initial tab will have the top portion of the tab missing from the tab bar. It will also overlap with the "+" new tab button.

This is the issue that @groovecoder also ran into while demoing containers. I can reproduce this pretty easily by simply opening a brand new custom container.

* the top of the tab is missing from the initial custom container tab
* the tab and the "+" new tab button are being overlapped

customcontainer

Platform being used: macOS 10.12.3 x64

**STR under fx51:**

* create a new custom container via the test pilot "Containers" button
* close all currently opened windows and ensure that FX is still running
* select "File -> New Container Tab -> [select the custom container created previously)

**STR under fx52:**

* create a new custom container via the test pilot "Containers" button
* simply open the new customer container and you'll see the issues in the tab bar.

johngruen:
A bit of a hack but, 
Fixes #150 

NB, this was done on top of #216 b/c it touches the same file

### DIFF
--- a/data/usercontext.css
+++ b/data/usercontext.css
@@ -1,59 +1,75 @@
-[data-identity-color="blue"] {
+/* HACK: Custom Container vars do not propigate correctly
+until the container tab is blurred and refocused, 
+adding the data-identity-color with the default hex
+value, or chrome url path as an alternate selector mitiages this bug.*/
+[data-identity-color="blue"],
+[data-identity-color="#00a7e0"] {
   --identity-tab-color: #37adff;
   --identity-icon-color: #37adff;
 }
 
-[data-identity-color="turquoise"] {
+[data-identity-color="turquoise"],
+[data-identity-color="#01bdad"] {
   --identity-tab-color: #00c79a;
   --identity-icon-color: #00c79a;
 }
 
-[data-identity-color="green"] {
+[data-identity-color="green"],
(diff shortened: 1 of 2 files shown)


johngruen:
I just noticed that this fix does not work for the custom circle icon in the URL bar. We might want to check if this works in beta/release

groovecoder:
This fixes, but #150 is also already fixed on master? So I don't know that we need this patch anymore?

johngruen:
@groovecoder wasn't working for me on master before i added this code, but maybe i was doing something wrong...lemmie confirm that i'm not hallucinating

johngruen:
@groovecoder yeah, i still have the issue on master in 53. This fixes it, but could you test in beta/release?

groovecoder:
So, neither 51 nor 52 had the issue in master.

But, 53 had the issue in master, and this patch fixes it in 53.

johngruen:
well, kinda...the custom circle icon still seems to be a problem. Thoughts on how to mitigate that?

groovecoder:
Not atm. Block merge on it?

johngruen:
nah

Issue #172, opened 2017-02-10
TypeError: ContextualIdentityService.getIdentityFromId(...) is undefined cookies.js:88:15


kjozwiak:
We're breaking the cookie manager when removing containers with associated tabs. I can reproduce this pretty reliably with custom containers. Once this happens, the user won't be able to view/manage their cookies via about:preferences#privacy..

See also [bug#1319929](https://bugzilla.mozilla.org/show_bug.cgi?id=1319929)

```
TypeError: ContextualIdentityService.getIdentityFromId(...) is undefined cookies.js:88:15
	gCookiesWindow._isPrivateCookie chrome://browser/content/preferences/cookies.js:88:15
	gCookiesWindow._loadCookies chrome://browser/content/preferences/cookies.js:503:13
	gCookiesWindow._populateList chrome://browser/content/preferences/cookies.js:52:5
	gCookiesWindow.init chrome://browser/content/preferences/cookies.js:35:5
	onload chrome://browser/content/preferences/cookies.xul:1:1
```

<img width="572" alt="screen shot 2017-02-10 at 4 19 05 pm" src="https://cloud.githubusercontent.com/assets/2602313/22844516/aa98ef2a-efac-11e6-97af-c1866bff16bd.png">

STR:

* install fx51.0.1, buildId: 20170125094131, changeset: 327e081221b0
* install the latest XPI (0.0.2) in this instance
* create a new customer container and open reddit.com
* open several of the top links in the new custom container via the right click context menu
* hide all the tabs that are opened in the custom container via "Hide these container tabs"
* once all the tabs are hidden, delete the entire container
* access the cookie manager via about:preferences#privacy

TanviHacks:
Marking this as MVP.  Let me know if you disagree.

groovecoder:
I was able to reproduce this in fx 51.0b14 on a blank profile without installing any add-ons at all. Is this a lower-level bug? Not related to this TP add-on/experiment?

kjozwiak:
> I was able to reproduce this in fx 51.0b14 on a blank profile without installing any add-ons at all. Is this a lower-level bug? Not related to this TP add-on/experiment?

[...]

PR #219, opened 2017-02-20
Fix the CookieManager issue #172

bakulf:


johngruen:
@bakulf you'll have to rebase against this: https://github.com/mozilla/testpilot-containers/pull/222

johngruen:
@jonathanKingston looks like this was failing travis for other reasons as well.

PR #232, opened 2017-02-21
Putting back fingerprint icon back in place.

jonathanKingston:


johngruen:
@jonathanKingston can you slap a screenshot in here and also rebase i guess?

PR #236, opened 2017-02-22
Use proxy

jonathanKingston:


jonathanKingston:
@johngruen can I get a +1 for this please :)

PR #249, opened 2017-02-22
Add ability to check for theme dark on changing themes. Fixes #207

### LINKED ISSUE #207: browserAction icon should change color on dark themes

jonathanKingston:


index.js
```diff
@@ -42,6 +42,7 @@ const PREFS = [
 
 const { attachTo, detachFrom } = require("sdk/content/mod");
 const { Cu } = require("chrome");
+const { Services }  = require("resource://gre/modules/Services.jsm");
```
bakulf: alphabetic order.

index.js
```diff
@@ -57,6 +58,7 @@ const { viewFor } = require("sdk/view/core");
 const webExtension = require("sdk/webextension");
 const windows = require("sdk/windows");
 const windowUtils = require("sdk/window/utils");
+const { LightweightThemeManager } = Cu.import("resource://gre/modules/LightweightThemeManager.jsm", {});
```
bakulf: same here.

webextension/background.js
```diff
@@ -1,3 +1,47 @@
...
+      16: "img/container-site-w-24.png",
+      32: "img/container-site-w-48.png"
+    };
+  }
+  browser.browserAction.setIcon({
+    path: icons
+  });
+}
+
+let existingTheme = null;
```
bakulf: I don't like to have global variables like this. Can you please create an object and move this code into it?
This object can do: 1 checkTheme, updateTheme and so on. Basically all the theme management.

index.js
```diff
@@ -170,6 +173,8 @@ const ContainerService = {
       "removeIdentity",
       "updateIdentity",
       "getPreference",
+      "getTheme",
+      "onThemeChanged",
```
bakulf: Why do  you expose it? It seems that it's not used.

index.js
```diff
@@ -227,6 +232,16 @@ const ContainerService = {
           sendReply(this[message.method](message));
         }
       });
+
+      // This is only used for theme notifications
+      api.browser.runtime.onConnect.addListener((port) => {
```
bakulf: The logic of the theme seems big enough to be in a ContainerTheme object by itself.

groovecoder: If not a new object, at least group these methods together so all the theme-managing logic is together.

groovecoder:
ping @jonathanKingston - if you can rebase and maybe group the methods together, I'll approve & merge here.

jonathanKingston:
@groovecoder I consolidated all the code to be in the same place instead of initting it's own state and so on created a fair few lines in the parent object and more issues to test etc.

index.js
```diff
@@ -292,6 +298,51 @@ const ContainerService = {
...
+    });
+  },
+
+  triggerThemeChanged(theme, topic) {
+    this._onThemeChangedCallbacks.forEach((callback) => {
+      callback(theme, topic);
+    });
+  },
+
+  observe(subject, topic) {
```
groovecoder: I don't know much about the observers service. I assume that the `observe` name is somewhat magical? I.e., the `Services.obs.addObserver(this, ...)` call calls `this.observe(subject, topic)` which enters here?

index.js
```diff
@@ -115,6 +117,7 @@ const ContainerService = {
   _identitiesState: {},
   _windowMap: new Map(),
   _containerWasEnabled: false,
+  _onThemeChangedCallbacks: [],
```
groovecoder: Does this need to be an array of callbacks? I only see a single callback that we're adding here. I.e., `port.postMessage` to send the `theme, topic` to the background `themeManager`?

Issue #246, opened 2017-02-22
Containers tabs are not displaying the Activity Stream "New Tab" page content if both are installed


SoftVision-PaulOiegas:
[Affected versions]:
- Firefox 51.0.1 and above
- Activity Stream 1.5.0
- Containers 0.9.3

[Affected platforms]:
- All Windows
- All Mac OS
- All Linux

[Prerequisites]:
- Have Activity Stream and Containers experiments installed on a new Firefox profile.

[Steps to reproduce]:
1. Open Firefox browser with the profile from prerequisites.
2. Click on the "New Tab" button and observe page content.
3. Hover over "New Tab" button and choose to open a new container type tab (any of the 4).
4. Observe the container new tab page content.

[Expected results]:
- Both normal new tab and containers new tab pages are displaying the Activity Stream page content.

[Actual results]: 
- Step 2: Default "New Tab" page displays the Activity Stream page content.
- Step 4: Containers new tab pages are displaying default Firefox "about:newtab" page content.

[Notes]:
- Here is a screen recording of the issue: https://goo.gl/QKAgdM.

johngruen:
https://github.com/mozilla/testpilot/pull/2227

johngruen:
there's an enhancement where we actually message something in the new tab, but that might be a different ticket.

bakulf:
my patch uses the default tab URL when a new tab is opened. This is nice to have and we should merge this code. About activity stream not able to deal with an empty container, what about if we create a separate landing page for containers?
I mean, when a new container tab is opened, instead showing about:something, we show a page scribing what containers are (exactly how we do for about:privatebrowsing).

groovecoder:
Activity Stream and Containers are both graduated out of Test Pilot now. Container tabs now show the same about:newtab as regular tabs.

PR #254, opened 2017-02-23
default tab page: issue #246

bakulf:


groovecoder:
This shows the activity stream home in a new container tab, but since most users' containers' history will start empty, it's a broken activity stream page?

txp-container-activity-stream

bakulf:
Well, definitely it makes sense to open the default homepage instead forcing an about:newtab.
This seems a bug in activity stream: as you can see all the strings are not localized.

groovecoder:
Closing this in favor of https://github.com/mozilla/testpilot-containers/issues/246 - we will warn users that activity stream doesn't work in container tabs. Better for users to get an un-broken about:newtab page than a broken activity stream page. I also filed https://github.com/mozilla/activity-stream/issues/2193 "upstream" to activity stream project.

Issue #257, opened 2017-02-23
"Move tabs to a new window" button is  selectable when the container is hidden


Softvision-VladBacia:
**[Affected versions]:**
-Firefox 51.0.1 and up
-Containers Experiment 0.9.3

**[Affected Platforms]:**
-All Windows
-All Mac
-All Linux

**[Prerequisites]:**
-Containers Experiment (v0.9.3) is installed on a Firefox new profile.
-Have a few websites in a container.

**[Steps to reproduce]:**
1. Click on "Containers" button from Toolbar.
2. Click on ">" More button for he container with the websites.
3. Click on "Hide this container" button.
4. Click on "Move tabs to a new window" button.

**[Expected result]:**
- "Move tabs to a new window" button is grayed out. It cannot be clicked.

**[Actual result]:**
- "Move tabs to a new window" button is clickable, after clicking on it, the drop-down menu is dismissed and it leaves the impression that containers will open in a new window, but that doesn't happen.

![move hidden tabs](https://cloud.githubusercontent.com/assets/9248513/23258153/91421790-f9cf-11e6-857f-bde3dea0a6ae.gif)

groovecoder:
Fixed in https://github.com/mozilla/testpilot-containers/pull/261/commits/4ec371df6c9d740980c682ae25eafbec4b047112

Softvision-VladBacia:
Great fix :thumbsup: . The issue is no longer reproducible with Containers Experiment 1.0.1 .

PR #261, opened 2017-02-24
Move to new window should show hidden tabs as well - issue #257

bakulf:


groovecoder:
This is a fix, but doesn't fix to the "Expected result" in https://github.com/mozilla/testpilot-containers/issues/257. @johngruen or @TanviHacks - what behavior do we want here?

groovecoder:
In addition, if we disable the "Move to new window" feature completely, we may not want to merge this.

bakulf:
I think we want to merge this doesn't matter what. It's nice to have. In the end, if I want to move tabs in a separate window, why should I do it with 2 clicks?

bakulf:
after the meeting, it seems we are happy with this patch, right? :)

TanviHacks:
When tabs are hidden and we select move tabs, the tabs should be both moved
and shown

Issue #242, opened 2017-02-22
Moving Container tabs to a new window breaks Pageshot, Pulse add-ons and the Test Pilot website


SoftVision-CiprianMuresan:
[Notes]:
- If the Test Pilot website is also moved this way, refreshing the page with F5 and address bar button no longer work. And if you refresh the page using the context menu, the all installed experiments no longer appear enabled.
- It seems that moving the tabs in a new container window causes Pageshot to hit [issue #1655.](https://github.com/mozilla-services/pageshot/issues/1655)
- And the same thing happens in Pulse's case, where it hits [issue #120](https://github.com/mozilla/pulse/issues/120) . However, the Pulse icon does not appear after a refresh. 

[Affected versions]:
- Firefox 51.0.1
- Containers 0.9.3

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- The "xpinstall.signatures.required" is set to "false" and latest Containers add-on is installed on a new clean profile.
- Pulse and Pageshot are also installed on the same profile.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites.
2. Navigate to https://en.wikipedia.org/wiki/Main_Page and open a few links in container tabs. (eg. Personal)
3. Click the "Containers" button from the browser toolbar.
4. Click the arrow next to the previously used container and choose to "Move tabs to a new window".
5. Observe the URL bar.
6. Click the "Pageshot" button from the browser toolbar and observe the behavior.

[Expected results]:
- Step 5. The Pulse icon is displayed in the URL bar.
- Step 6. The Pageshot overlay covers the entire screen.

[Actual results]:
- Step 5. The Pulse icon is no longer displayed in the URL bar.
- Step 6. The Pageshot overlay does not appear on the screen and after a while it shows an error.

[Additional Notes]:
- Here is a screen recording with the issue https://goo.gl/E9rgJv.

6a68:
This bug sounds really bad, but I actually don't think it's anything to do with the containers addon.

The pulse and pageshot bugs are both symptoms of the same bug with the sdk tabs library: sdk/tabs loses track of tabs if the parent window changes.

Moving tabs to a new window without the containers addon would have the same effect.

I'd actually suggest we close this as invalid, since it's just uncovering known bugs that affect the other two addons.

groovecoder:
[...]

PR #276, opened 2017-02-27
for #242: disable move when incompatible add-ons detected

groovecoder:
I dislike almost everything about this patch/PR, but we want to disable the move feature when we detect incompatible add-ons. There's a hover tooltip, but [bug 1269421](https://bugzilla.mozilla.org/show_bug.cgi?id=1269421) means I had to also add a message in the pop-up itself ...

txp-containers-move-incompat

To verify/test ...

1. Install Pulse (or PageShot) addons (via Test Pilot add-on or manually)
2. Install Containers addon
3. Create a container tab
4. Open the "container tabs" panel
5. [ ] Verify "Move tabs" is disabled with an appropriate message
6. Un-install or disable pulse (or PageShot) addon(s)
7. Open the "container tabs" panel again
8. [ ] Verify "Move tabs" is re-enabled

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,6 +8,12 @@ const DEFAULT_TAB = "about:newtab";
 const SHOW_MENU_TIMEOUT = 100;
 const HIDE_MENU_TIMEOUT = 300;
 
+const INCOMPATIBLE_ADDON_IDS = [
+  "pulse@mozilla.com",
+  "snoozetabs@mozilla.com",
+  "jid1-NeEaf3sAHdKHPA@jetpack" // PageShot
+];
+
 const IDENTITY_COLORS = [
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -488,6 +496,23 @@ const ContainerService = {
     this._sendEvent(payload);
   },
 
+  checkIncompatibleAddons() {
+    return new Promise(resolve => {
+      AddonManager.getAddonsByIDs(INCOMPATIBLE_ADDON_IDS, (addons) => {
+        if (addons.length === 0) {
```
bakulf: Just do:

addons = addons.filter((a) => a && a.isActive);
resolve(addons.length !== 0);

groovecoder: Ah, nice!

PR #278, opened 2017-02-27
fix #272: add event telemetry for incompat addons

### LINKED ISSUE #272: Record when user encounters incompatible add-on message
When a user encounters the disabled "move" feature because of incompatible add-ons, record that into Telemetry, so we learn how many users it's affecting.

groovecoder:
This adds a Telemetry ping when users encounter the "Incompatible addons" message in the containers pop-up. The ping is sent with `uuid` so we will be able to group the pings to know *how many users* are seeing the message - not just the raw counts of how many times it's shown.

To verify:

1. Install Test Pilot add-on (required to create Telemetry channel)
2. Install Pulse (or PageShot) experiment(s)
3. Install Containers experiment
4. Open a container tab
5. Open the pageAction popup
6. Click the container
7. [ ] Verify you see the "incompatible with experiments" message on the disabled move action
8. Go to `about:telemetry`, "Archived ping data" ...

txp-about-telemetry

* [ ] Verify the latest "Raw payload" of the ping data is the "incompatible-addons-detected" ping.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -169,6 +169,15 @@ of a `testpilottest` telemetry ping for each scenario.
   }
 ```
 
+* When a user encounters the disabled "move" feature because of incompatible add-ons
+
+```js
+  {
+    "uuid": <uuid>,
+    "event": "incompatible-addons-detected"
+  }
+```
+
 ### A Redshift schema for the payload:
 
 ```lua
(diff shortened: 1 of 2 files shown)


6a68:
@groovecoder The Test Pilot addon automatically adds the `ClientID` when experiment pings are forwarded to the Telemetry service, so there's no need to send a uuid like you do here. I'll open a Test Pilot bug to make that clearer in the docs.

6a68:
response from IRC:

> •groovecoder> _6a68: during our metrics review, I was told querying our own data + the Firefox ClientID metric would be ... complicated.

PR #285, opened 2017-02-28
Hide non container icon in URL bar. Fixes #281

### LINKED ISSUE #281: container icons under the awesome bar persisting into non-container tabs
When creating a new container tab with the latest containers_experiment-1.0.0-fx.xpi, the awesome bar container icon will persist into non-container tabs. We should fix this before launching as it might confuse a lot of users as this is a pretty obvious bug... Users might think:

* why they can't open a regular tab via "+" once they notice the icons persisting across non-container tabs
* think they're using a container tab due to the UI but in reality are using a non-container tab

> Example of issue: https://youtu.be/42fOmlJGV_4

STR:

* install containers_experiment-1.0.0-fx.xpi (the latest XPI)
* open a personal container via the new tab "+" button
* open a regular tab via the new tab "+" button

You'll notice that the personal container icon will appear under the awesome bar for a non-container tab.

jonathanKingston:


jonathanKingston:
@johngruen or @groovecoder this is to fix the #281 issue. I merged it into master so it is ready to go into prod with this PR.

PR #292, opened 2017-02-28
Many things

bakulf:


index.js
```diff
@@ -956,12 +956,16 @@ const ContainerService = {
 
     const userContextId = ContainerService._getUserContextIdFromTab(tab);
     return ContainerService.getIdentity({userContextId}).then(identity => {
+      const hbox = viewFor(tab.window).document.getElementById("userContext-icons");
+
       if (!identity) {
+        hbox.setAttribute("data-identity-color", "");
+        hbox.hidden = true;
```
jonathanKingston: This shouldn't be needed as the CSS takes care of this.

index.js
```diff
@@ -956,12 +956,16 @@ const ContainerService = {
...
         return;
       }
 
-      const hbox = viewFor(tab.window).document.getElementById("userContext-icons");
       hbox.setAttribute("data-identity-color", identity.color);
+      hbox.hidden = false;
```
jonathanKingston: Same here

index.js
jonathanKingston: Happy for you to remove the CSS that does this instead though :)

index.js
```diff
@@ -784,7 +784,7 @@ const ContainerService = {
       } else {
         this._recentBrowserWindow().then(browserWin => {
           resolve(browserWin);
-        });
+        }).catch(() => {});
```
jonathanKingston: Probably should throw something like:
```
new Error("This shouldn't happen when checking for a window");
```

Or similar.... and any other catch's we have without errors

index.js
```diff
@@ -1253,9 +1263,14 @@ ContainerWindow.prototype = {
   _configureAllTabsMenu() {
     return this._configureMenu("alltabs_containersTab", null, e => {
       const userContextId = parseInt(e.target.getAttribute("data-usercontextid"), 10);
-      ContainerService.openTab({
-        userContextId: userContextId,
-        source: "alltabs-menu"
+      ContainerService.showTabs({
```
jonathanKingston: Would it be easier to show all tabs in the openTab function so this would never happen again?

index.js
```diff
@@ -228,15 +228,16 @@ const ContainerService = {
...
     });
 
     // Modify CSS and other stuff for each window.
 
-    this.configureWindows().catch(() => {});
+    this._configureWindows().catch(() => {});
 
     windows.browserWindows.on("open", window => {
-      this.configureWindow(viewFor(window)).catch(() => {});
+      this._configureWindow(viewFor(window)).catch(() => {});
```
jonathanKingston: Please add Errors here too.

bakulf: Yeah, we should do it for each catch(). Separate commit for it, separate PR.

PR #376, opened 2017-03-13
Removal of upgrade process. Bump version to 1.1.2 Fixes 373

jonathanKingston:


groovecoder:
Skipping `uninstall` completely causes other bugs where the pageAction pop-up won't open. I've got a different approach/branch that preserves the custom containers AND keeps everything working.

Issue #373, opened 2017-03-13
updating to 1.1.1 removes custom containers including all associated cookies


kjozwiak:
Updating to 1.1.1 removes all the custom containers and their associated cookies. The default containers that come with the experiment (Personal, Banking, Shopping, Work) don't seem to be affected. Users who are heavy container users and have several custom containers will lose all their data and will have to re-create their custom containers once they're updated..

@groovecoder already found and instance where a user ran into this issues:
* https://discourse.mozilla-community.org/t/containers-update-closed-all-container-tabs/14367

We need to either revert back to 1.0.4 which I believe was the last version or quickly get a fix updated into production to try mitigating the damage.

STR:

* install test pilot via https://testpilot.firefox.com/
* install containers_experi...-1.0.4-fx.xpi manually
* open a personal, work and banking container and visit several websites within each of those containers
* create two custom containers and visit several websites
* wait around 3-5 minutes till FX pings the test pilot and Containers is updated to 1.1.1 (check via about:support)
* once the Container experiment is updated to 1.1.1, all the custom containers will be removed
* if you take a look at the cookies.sqlite database, all the associated cookies with those containers have also been removed

groovecoder:
2.0.0 released and published today. Users should start seeing it update now. It will destroy the custom 1.x.x containers, but we'll follow up quickly with a 2.0.1 release that should verify we preserve custom containers in future updates.

PR #378, opened 2017-03-14
Fix upgrading issue whilst also ensuring dead objects won't happen. #373

jonathanKingston:


groovecoder:
The [additional check for `ss.storage.savedConfiguration`](https://github.com/mozilla/testpilot-containers/pull/378/commits/4273ca67333d0a6952c3383dff521bfda20fe8c5#diff-168726dbe96b3ce427e7fedce31bb0bcR127) fixes the issue I saw.

I was able to update 1.0.4 -> 2.0.0 -> 2.1.0 and 1.1.1 -> 2.0.0 -> 2.1.0 and the add-on kept working the whole time.

As expected, the 1.x -> 2.0.0 upgrade destroyed custom containers (hence the 2.x "backwards-incompatible" version), but the 2.0.0 -> 2.1.0 upgrade preserves them.

So, this looks good to me. We should probably merge and release ASAP before more users create more custom containers that will be destroyed when they get updated to 2.0.0.

groovecoder:
In addition, the 1.x -> 2.x upgrade creates new UUIDs for users, so we should keep track of versions in our dashboard queries.

PR #379, opened 2017-03-15
Better sanity

jonathanKingston:
- Renamed the linter such that it can be more flexible in checking other code (we were the only user of it so it's fine and important to do it earlier)
- Added / to escaping, old fxos added that so I would rather keep that
- Moved as many things to using attrs and textContent as possible to limit further issues from identity.name or tab.name other vars used in innerHTML use the escaped fn


groovecoder:
`npm test` passes.

I also did a spot-check where I made this branch into a 2.0.1 release and upgraded from a 2.0.0 release made from master branch. Custom containers were preserved, and this branch correctly escapes container names everywhere they are shown.

Issue #311, opened 2017-03-02
Consider a tab context menu option to convert a normal tab to a container


pmac:
I often click a link from an external app and a normal tab opens, but I'm logged into the web app in a container. I'd like to be able to just convert the tab rather than copy the URL, close the original tab, open a new container tab, and paste the URL.

hasanen:
Perhaps it isn't that big of job add a feature to convert tab from container to another within this?

attis:
This might be a good enough solution for #325 I think.

maverick74:
This is a very important feature that will ease job a LOT, from my point of view :).

Thumbs up for this :)

smichel17:
Privacy-concerned user here. I would prefer that external links not all be opened in the same context. So, I would not consider this a good enough solution for #325. Would you consider allowing users to specify which container they would like to open a new tab in each time? Here's how this could work:

---
First, add a setting akin to the one for downloads, `Open external links in` with the options of selecting a default container or `Always ask me where to open links`. Screenshot for reference:

![image](https://cloud.githubusercontent.com/assets/6563664/23560453/82cebf94-0008-11e7-9181-a85a514048c0.png)

If `Always ask me where to open links` is selected, link get pre-filled into the URL bar with the cursor placed at the end, with drop-down options to open it in either the global namespace or a specific container. Quick pixel art mockup:

![firefox-containers-link](https://cloud.githubusercontent.com/assets/6563664/23560203/9e039a7e-0007-11e7-8b5d-794b36cb61f6.png)

aleth:
This would be very useful, possibly together with "Move all tabs to container..." (similar to what exists for bookmarks).

While containers were primarily designed to keep identities compartmentalized, one should not underestimate how many people will use them simply for tab management (e.g. as a replacement for tab groups).

smichel17:
I think there are two related but definitely different questions here: one is about *privacy* and the other is about *convenience.*

### Privacy
- **The problem:** It is difficult to keep profiles separate because it is hard to control what container an external link is opened in.
- **The solution:** Add the ability to choose which container a tab is opened into.
- This is a duplicate of #325.
- Moving a tab to a new container after it has been opened is not a solution; by that point, the separation of containers has been compromised.

A great example of the privacy problem just happened to me: I was in my development container and went to create an account on http://discourse.mozilla-community.org/. I put in my email address and sent a confirmation email, opened my email client, clicked the confirmation link... and confirmed my account outside of a container. Ooops.

[...]

Issue #306, opened 2017-03-02
Allow classifying of sites or urls as private,work, ...


crabel:
Some urls/sites are always "private", some are always "work". At least for me. So, it would be nice, if those urls would be opened automatically as "private" or "work" when I type them or select them from bookmarks.

Currently I have to choose the type of a tab, before I open it. Otherwise it is opened in a "standard" tab.

But I really would like to be able to simply type the url or choose it from my bookmarks and the correct (previously "mapped") tag is used automatically.

Note:  This feature shouldn't extend to links I click, only to bookmarks and links I type.
Note2: If I select the type "private" manually, it should also not override my manual setting.

groovecoder:
We anticipate this enhancement may get a lot of up-votes. :)

chris-morgan:
> Note: This feature shouldn't extend to links I click, only to bookmarks and links I type.

By “links I click” I presume you mean from inside the browser. Links from *outside* the browser should open in the classified container as though I had typed them.

cpeterso:
Once the extension has the capability to remember containers for sites, it would be interesting to include a default classification list for popular sites. So out of the box, facebook.com might automatically be assigned to the Personal container and amazon.com to the Shopping container without any user configuration.

crabel:
@chris-morgan Yes, I meant: "Links I click in a tab". Links from outside should be treaded the same as bookmarks or links I enter in a new tab. I also don't think a container should change. When I open a "private" container and type a "work" url, it should stay "private". (Note: I am not a 100% attached to this behavior, but I think it is the most clear/logic/intuitive)

The main question for me is, will people understand that. I mean, you are in the shopping container and click some link "Login with Facebook" or something. And then they complain, because they are already authenticated (of course, in the other container). But that's a question that affects the whole idea of containers. I think, the idea of separation needs to be communicated very clearly, e.g. with a nice image showing that each container is "its own world".

jonathanKingston:
If we implement this only for URL bar and bookmarks I think we could drop the default on link clicks within containers perhaps.

We could also consider cookies in other containers to provide the user with options on link press too. So if the user has clicked a link within work mail to github we could look at cookies and offer a modal to allow them to approve jumping into a work or personal container. That way we keep the security/privacy benefits of containers whilst offering the most contextual UX.

kyleloyka:
Could this idea be expanded so that by default, there could be a separate container for every (visited) root domain?

groovecoder:
[...]

PR #386, opened 2017-03-16
WIP of auto-container-origin

jonathanKingston:
Super WIP PR just to demo the functionality for @johngruen and @groovecoder ...

STR (Assumes default search is google):

1. create Google container
2. add a new google container tab
3. go to google.com
4. right click, "Always open in this container"
5. open a new non container tab
6. Search string "test"

Tab created in step 6. Will be in "Google" container.

Loads of polish needed like:
- Change context menu when already auto linking.
- Show warning for paranoid users like myself on autolinking
- Allow container based auto dismissal of paranoid warning
- Potentially show URL autolinking in other menus

### DIFF
--- a/webextension/background.js
+++ b/webextension/background.js
@@ -1,14 +1,266 @@
+const assignManager = {
+  CLOSEABLE_WINDOWS: new Set([
+    "about:startpage",
+    "about:newtab",
+    "about:home",
+    "about:blank"
+  ]),
+  MENU_ASSIGN_ID: "open-in-this-container",
+  MENU_REMOVE_ID: "remove-open-in-this-container",
+  storageArea: {
+    area: browser.storage.local,
+
+    getSiteStoreKey(pageUrl) {
+      const url = new window.URL(pageUrl);
+      const storagePrefix = "siteContainerMap@@_";
+      return `${storagePrefix}${url.hostname}`;
+    },
+
+    get(pageUrl) {
+      const siteStoreKey = this.getSiteStoreKey(pageUrl);
+      return new Promise((resolve, reject) => {
+        this.area.get([siteStoreKey]).then((storageResponse) => {
+          if (storageResponse && siteStoreKey in storageResponse) {
+            resolve(storageResponse[siteStoreKey]);
+          }
+          resolve(null);
+        }).catch((e) => {
+          reject(e);
+        });
+      });
+    },
+
+    set(pageUrl, data) {
+      const siteStoreKey = this.getSiteStoreKey(pageUrl);
+      return this.area.set({
+        [siteStoreKey]: data
+      });
+    },
+
+    remove(pageUrl) {
+      const siteStoreKey = this.getSiteStoreKey(pageUrl);
+      return this.area.remove([siteStoreKey]);
+    },
+
+    deleteContainer(userContextId) {
+      const removeKeys = [];
+      this.area.get().then((siteConfigs) => {
+        Object.keys(siteConfigs).forEach((key) => {
+          // For some reason this is stored as string... lets check them both as that
+          if (String(siteConfigs[key].userContextId) === String(userContextId)) {
+            removeKeys.push(key);
+          }
+        });
+        this.area.remove(removeKeys);
+      }).catch((e) => {
+        throw e;
+      });
+    }
+  },
+
+  init() {
+    browser.tabs.onActivated.addListener((info) => {
+      browser.tabs.get(info.tabId).then((tab) => {
+        this.calculateContextMenu(tab);
+      }).catch((e) => {
+        throw e;
+      });
+    });
+
+    browser.windows.onFocusChanged.addListener((windowId) => {
+      browser.tabs.query({active: true, windowId}).then((tabs) => {
+        if (tabs && tabs[0]) {
+          this.calculateContextMenu(tabs[0]);
+        }
+      }).catch((e) => {
+        throw e;
+      });
+    });
+
+    browser.runtime.onMessage.addListener((neverAskMessage) => {
+      const pageUrl = neverAskMessage.pageUrl;
+      if (neverAskMessage.neverAsk === true) {
+        // If we have existing data and for some reason it hasn't been deleted etc lets update it
+        this.storageArea.get(pageUrl).then((siteSettings) => {
+          if (siteSettings) {
+            siteSettings.neverAsk = true;
+            this.storageArea.set(pageUrl, siteSettings);
+          }
+        }).catch((e) => {
+          throw e;
+        });
+      }
+    });
+
+    browser.contextMenus.onClicked.addListener((info, tab) => {
+      const userContextId = this.getUserContextIdFromCookieStore(tab);
+      // Mapping ${URL(info.pageUrl).hostname} to ${userContextId}
+      if (userContextId) {
+        let actionName;
+        let storageAction;
+        if (info.menuItemId === this.MENU_ASSIGN_ID) {
+          actionName = "added";
+          storageAction = this.storageArea.set(info.pageUrl, {
+            userContextId,
+            neverAsk: false
+          });
+        } else {
+          actionName = "removed";
+          storageAction = this.storageArea.remove(info.pageUrl);
+        }
+        storageAction.then(() => {
+          browser.notifications.create({
+            type: "basic",
+            title: "Containers",
+            message: `Successfully ${actionName} site to always open in this container`,
+            iconUrl: browser.extension.getURL("/img/onboarding-1.png")
+          });
+          this.calculateContextMenu(tab);
+        }).catch((e) => {
+          throw e;
+        });
+      }
+    });
+
+    browser.webRequest.onBeforeRequest.addListener((options) => {
+      if (options.frameId !== 0 || options.tabId === -1) {
+        return {};
+      }
+      return Promise.all([
+        browser.tabs.get(options.tabId),
+        this.storageArea.get(options.url)
+      ]).then(([tab, siteSettings]) => {
+        const userContextId = this.getUserContextIdFromCookieStore(tab);
+        if (!siteSettings
+            || userContextId === siteSettings.userContextId
+            || tab.incognito) {
+          return {};
+        }
+
+        this.reloadPageInContainer(options.url, siteSettings.userContextId, tab.index, siteSettings.neverAsk);
+        this.calculateContextMenu(tab);
+        // If the user just opened the tab, we can auto close it
+        if (this.CLOSEABLE_WINDOWS.has(tab.url)) {
+          browser.tabs.remove(tab.id);
+        }
+        return {
+          cancel: true,
+        };
+      }).catch((e) => {
+        throw e;
+      });
+    },{urls: ["<all_urls>"], types: ["main_frame"]}, ["blocking"]);
+
+    browser.webRequest.onCompleted.addListener((options) => {
+      if (options.frameId !== 0 || options.tabId === -1) {
+        return {};
+      }
+      browser.tabs.get(options.tabId).then((tab) => {
+        this.calculateContextMenu(tab);
+      }).catch((e) => {
+        throw e;
+      });
+    },{urls: ["<all_urls>"], types: ["main_frame"]});
+  },
+
+
+  deleteContainer(userContextId) {
+    this.storageArea.deleteContainer(userContextId);
+  },
+
+  getUserContextIdFromCookieStore(tab) {
+    if (!("cookieStoreId" in tab)) {
+      return false;
+    }
+    const cookieStore = tab.cookieStoreId;
+    const container = cookieStore.replace("firefox-container-", "");
+    if (container !== cookieStore) {
+      return container;
+    }
+    return false;
+  },
+
+  isTabPermittedAssign(tab) {
+    // Ensure we are not an important about url
+    // Ensure we are not in incognito mode
+    if (this.CLOSEABLE_WINDOWS.has(tab.url)
+        || tab.incognito) {
+      return false;
+    }
+    return true;
+  },
+
+  calculateContextMenu(tab) {
+    // There is a focus issue in this menu where if you change window with a context menu click
+    // you get the wrong menu display because of async
+    // See: https://bugzilla.mozilla.org/show_bug.cgi?id=1215376#c16
+    // We also can't change for always private mode
+    // See: https://bugzilla.mozilla.org/show_bug.cgi?id=1352102
+    const cookieStore = this.getUserContextIdFromCookieStore(tab);
+    browser.contextMenus.remove(this.MENU_ASSIGN_ID);
+    browser.contextMenus.remove(this.MENU_REMOVE_ID);
+    // Ensure we have a cookieStore to assign to
+    if (cookieStore
+        && this.isTabPermittedAssign(tab)) {
+      this.storageArea.get(tab.url).then((siteSettings) => {
+        // ✓ This is to mitigate https://bugzilla.mozilla.org/show_bug.cgi?id=1351418
+        let prefix = "   "; // Alignment of non breaking space, unknown why this requires so many spaces to align with the tick
+        let menuId = this.MENU_ASSIGN_ID;
+        if (siteSettings) {
+          prefix = "✓";
+          menuId = this.MENU_REMOVE_ID;
+        }
+        browser.contextMenus.create({
+          id: menuId,
+          title: `${prefix} Always Open in This Container`,
+          checked: true,
+          contexts: ["all"],
+        });
+      }).catch((e) => {
+        throw e;
+      });
+    }
+  },
+
+  reloadPageInContainer(url, userContextId, index, neverAsk = false) {
+    const loadPage = browser.extension.getURL("confirm-page.html");
+    // If the user has explicitly checked "Never Ask Again" on the warning page we will send them straight there
+    if (neverAsk) {
+      browser.tabs.create({url, cookieStoreId: `firefox-container-${userContextId}`, index});
+    } else {
+      const confirmUrl = `${loadPage}?url=${url}`;
+      browser.tabs.create({url: confirmUrl, cookieStoreId: `firefox-container-${userContextId}`, index}).then(() => {
+        // We don't want to sync this URL ever nor clutter the users history
+        browser.history.deleteUrl({url: confirmUrl});
+      }).catch((e) => {
+        throw e;
+      });
+    }
+  }
+};
+
+const messageHandler = {
+  init() {
+    const port = browser.runtime.connect();
+    port.onMessage.addListener(m => {
+      switch (m.type) {
+      case "lightweight-theme-changed":
+        themeManager.update(m.message);
+        break;
+      case "delete-container":
+        assignManager.deleteContainer(m.message.userContextId);
+        break;
+      default:
+        throw new Error(`Unhandled message type: ${m.message}`);
+      }
+    });
+  }
+};
+
 const themeManager = {
   existingTheme: null,
   init() {
     this.check();
-
-    const port = browser.runtime.connect();
-    port.onMessage.addListener(m => {
-      if (m.type === "lightweight-theme-changed") {
-        this.update(m.theme);
-      }
-    });
   },
   setPopupIcon(theme) {
     let icons = {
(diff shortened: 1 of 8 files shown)


groovecoder:
This is for #306? Anything for #311 in here?

webextension/background.js
```diff
@@ -1,3 +1,63 @@
...
+  }
+  return false;
+}
+
+const storagePrefix = "originMap@@_";
+const storageArea = browser.storage.local;
+
+browser.contextMenus.onClicked.addListener((info, tab) => {
+  const pageUrl = new window.URL(info.pageUrl);
+  const userContextId = getUserContextIdFromCookieStore(tab.cookieStoreId);
```
groovecoder: 🎉 `tab.cookieStoreId` Yay Firefox 52 WebExtensions!

jonathanKingston: Does this need to work for 51? Totally forgot about that :/

jonathanKingston:
Sorry yea this is for #306.

Nothing for #311 here. There are other issues with doing that I need to explain there. I think we should keep them separate.

webextension/background.js
```diff
@@ -1,3 +1,63 @@
...
+  // Mapping ${pageUrl.host} to ${userContextId}
+  if (userContextId) {
+    const originStoreKey = `${storagePrefix}${pageUrl.host}`;
+    storageArea.set({[originStoreKey]: userContextId});
+  }
+});
+
+browser.webRequest.onBeforeRequest.addListener((options) => {
```
groovecoder: Could we also add this check?
```
if (options.frameId !== 0) {
  return {}
}
```
So we only do this logic for the requests that happen in the top frame?

I.e., we just want to check the first/top request in the tab - not all the requests?

webextension/background.js
```diff
@@ -1,3 +1,63 @@
...
+
+function getUserContextIdFromCookieStore(cookieStore) {
+  const container = cookieStore.replace("firefox-container-", "");
+  if (container !== cookieStore) {
+    return container;
+  }
+  return false;
+}
+
+const storagePrefix = "originMap@@_";
```
groovecoder: nit: more descriptive name? `originContainerMap@@_` ?

webextension/background.js
```diff
@@ -1,3 +1,63 @@
...
+     }
+     browser.tabs.create({url: options.url, cookieStoreId: `firefox-container-${mapUserContextId}`});
+     // If the user just opened the tab, we can auto close it
+     if (tab.url == "about:newtab" || tab.url === "about:home") {
+       browser.tabs.remove(tab.id);
+     }
+     return {
+      cancel: true,
+     };
+     return {};
```
groovecoder: unreachable?

webextension/background.js
```diff
@@ -1,3 +1,63 @@
...
+  }).catch((e) => {
+    console.log('err', e);
+    return {};
+  });
+},{urls: ["<all_urls>"], types: ["main_frame"]}, ["blocking"]);
```
groovecoder: Ah, the `main_frame` `RequestFilter` here takes care of my concern about processing every request. 👍

groovecoder:
>     Show warning for paranoid users like myself on autolinking
Absolutely 👍 

>     Allow container based auto dismissal of paranoid warning
I would leave this out at first, until users tell us they're sick of the warning? Based on the experience in Blok, I imagine most users will get their origins and containers linked in the first couple weeks, and it seems we should err on the side of reminding them about the privacy & security concerns of auto-linking rather than err on the side of users easily falling back into insecure browsing.

>     Potentially show URL autolinking in other menus
👍 for the tab-bar context menu (is this possible in WebExtension?), the container icon in the Awesomebar (see https://github.com/mozilla/testpilot-containers/issues/387 - possible in WebExtension?), and/or a new pageAction.

webextension/background.js
```diff
@@ -1,3 +1,63 @@
...
+  if (container !== cookieStore) {
+    return container;
+  }
+  return false;
+}
+
+const storagePrefix = "originMap@@_";
+const storageArea = browser.storage.local;
+
+browser.contextMenus.onClicked.addListener((info, tab) => {
```
groovecoder: Do we need to add:
```
if (info.menuItemId !== "open-in-this-container") {
  return;
}
```
To make sure that clicks on other context menus don't trigger the assignment?

jonathanKingston: Yeah if we add any certainly. There is a click handler per each menu item too, I just didn't test that out yet.

webextension/background.js
jonathanKingston: Added the check anyway for paranoia sake.

webextension/background.js
jonathanKingston: Yup deleted.

webextension/background.js
jonathanKingston: Done... the `@@_` is my paranoia about something merging with the key itself as it's just a boring key val.

jonathanKingston:
I put in a rough warning when users are directed to a container, currently this opens in the container and I use history replace to move them into the same container. It might actually make more sense to use message passing to the background page to open in a selected container (this would allow the user flow of people who want to auto assign the container but also want to change it *sometimes*).

To make the never ask checkbox work I had to use message passing anyway.

I would have to put it a fair bit more work in to get any of these features:
- container select box, picker whatever
- container name (The background script would need to maintain/fetch just for this request the list of containers which actually would be a pain)

I think we should show the full URL here however... I also think it looks ugly :(

This really needs some of @johngruen's touch probably either way it's lower on the priorities I have on making pretty.
selection_515

groovecoder:
👍 to some styling work on it.

Some clearer copy may also be:

"
Are you sure you want to open <url> in your pre-assigned container? The page will have access to this container's cookies and other data.

<Open in non-Container> <Open in pre-assigned container"> <Never ask me this again>
"

johngruen:
hey @jonathanKingston @groovecoder work week this week has put me behind. Should be able to clean up by friday.

webextension/background.js
bakulf: Yeah. I introduced cookieStoreId, but it's not in FF51.... we can still retrieve the userContextId from the tab as we do in index.js.

webextension/background.js
```diff
@@ -1,3 +1,82 @@
...
+},{urls: ["<all_urls>"], types: ["main_frame"]}, ["blocking"]);
+
+
+function loadOtherPage(url, userContextId, neverAsk = false) {
+  const loadPage = browser.extension.getURL("confirm-page.html");
+  // If the user has explicitly checked "Never Ask Again" on the warning page we will send them straight there
+  if (neverAsk) {
+    browser.tabs.create({url, cookieStoreId: `firefox-container-${userContextId}`});
+  } else {
+    browser.tabs.create({url: `${loadPage}?url=${url}`, cookieStoreId: `firefox-container-${userContextId}`});
```
bakulf: What about the position? We should open the new tab in the same position where we are.
Plus, doing this we are breaks the Back-Forward feature.

bakulf: Another point would be: what about if the URL is loaded from a window.open() ? This breaks also 'the web' because you will open a tab, instead of a window, and the parent window will not have a reference to the new one.

groovecoder:
The Firefox sec warning page is a good style to copy:

sec-warning

webextension/background.js
jonathanKingston: I would somewhat argue breaking window.opener is a feature. I'm not sure we can fix that through a web extension either right?

jonathanKingston:
@groovecoder @bakulf @kjozwiak and @johngruen I pushed an update for the branch auto-container-origin

This has the index mentioned by @bakulf and the new error layout as discussed with @groovecoder and demoed in the meeting.

It is still missing the ability to remove the unassign though.

webextension/background.js
```diff
@@ -1,3 +1,174 @@
+const assignManager = {
+  closeableWindows: new Set([
```
groovecoder: nit: `CLOSEABLE_WINDOWS` to denote it is a constant/hard-coded set.

webextension/background.js
```diff
@@ -1,3 +1,174 @@
...
+    "about:startpage",
+    "about:newtab",
+    "about:home",
+    "about:blank"
+  ]),
+  menuAssignId: "open-in-this-container",
```
groovecoder: `MENU_ASSIGNED_ID`

webextension/background.js
```diff
@@ -1,3 +1,174 @@
...
+  menuRemoveId: "remove-open-in-this-container",
```
groovecoder: `MENU_REMOVE_ID`

webextension/background.js
```diff
@@ -1,3 +1,174 @@
...
+      browser.tabs.query({active: true}).then((tabs) => {
+        if (tabs && tabs[0]) {
+          this.calculateContextMenu(tabs[0]);
+        }
+      }).catch((e) => {
+        throw e;
+      });
+    });
+
+    browser.runtime.onMessage.addListener((request) => {
```
groovecoder: Can we change `request` to something more accurate like `neverAskMessage`? When I first read this, I thought `request` was an HTTP request of some kind.

webextension/background.js
```diff
@@ -1,3 +1,174 @@
...
+      browser.tabs.get(options.tabId).then((tab) => {
+        this.calculateContextMenu(tab);
+      }).catch((e) => {
+        throw e;
+      });
+    },{urls: ["<all_urls>"], types: ["main_frame"]});
+  },
+
+  getOriginStoreKey(url) {
+    const storagePrefix = "originContainerMap@@_";
```
groovecoder: nit: `STORAGE_PREFIX` to denote hard-coded constant string.

webextension/background.js
```diff
@@ -1,3 +1,174 @@
...
+    browser.contextMenus.onClicked.addListener((info, tab) => {
+      const pageUrl = new window.URL(info.pageUrl);
+      const userContextId = this.getUserContextIdFromCookieStore(tab);
+      // Mapping ${pageUrl.origin} to ${userContextId}
+      if (userContextId) {
+        const originStoreKey = this.getOriginStoreKey(pageUrl);
+        let actionName;
+        let storageAction;
+        if (info.menuItemId === this.menuAssignId) {
+          actionName = "added";
```
groovecoder: "added" should be "assigned"?

webextension/background.js
```diff
@@ -1,3 +1,174 @@
...
+          storageAction = this.storageArea.set({[originStoreKey]: userContextId});
+        } else {
+          actionName = "removed";
```
groovecoder: "removed" should be "unassigned"?

webextension/background.js
```diff
@@ -1,3 +1,174 @@
...
+          storageAction = this.storageArea.remove([originStoreKey]);
+        }
+        storageAction.then(() => {
+          browser.notifications.create({
+            type: "basic",
+            title: "Containers",
+            message: `Successfully ${actionName} ${pageUrl.origin} to always open in this container`,
```
groovecoder: `pageUrl.hostname` may look nicer/cleaner.

webextension/background.js
```diff
@@ -1,3 +1,174 @@
...
+          this.calculateContextMenu(tabs[0]);
+        }
+      }).catch((e) => {
+        throw e;
+      });
+    });
+
+    browser.runtime.onMessage.addListener((request) => {
+      if (request.neverAsk === true) {
+        this.storageArea.set({neverAsk: true});
```
groovecoder: Ah, so this sets a global "never ask" preference. I thought we should start with a "Never ask again **for this site**".

webextension/background.js
```diff
@@ -1,3 +1,174 @@
...
+      const pageUrl = new window.URL(options.url);
+      const originStoreKey = this.getOriginStoreKey(pageUrl);
+      return Promise.all([
+        browser.tabs.get(options.tabId),
+        this.storageArea.get([originStoreKey, "neverAsk"])
+      ]).then(([tab, storage]) => {
+        const userContextId = this.getUserContextIdFromCookieStore(tab);
+        let neverAsk = false;
+        if (!(originStoreKey in storage)) {
+          return;
```
groovecoder: `return {}` to be more explicit?

webextension/background.js
```diff
@@ -1,3 +1,174 @@
...
+          browser.contextMenus.create({
+            id: this.menuRemoveId,
+            title: "Remove Open in This Container",
+            contexts: ["all"],
+          });
+          return;
+        }
+        browser.contextMenus.create({
+          id: this.menuAssignId,
+          title: "Always Open in This Container",
```
groovecoder: Maybe "Assign this site to this Container" ?

webextension/background.js
```diff
@@ -1,3 +1,174 @@
...
+    browser.contextMenus.remove(this.menuRemoveId);
+    // Ensure we have a cookieStore to assign to
+    // Ensure we are not an important about url
+    if (cookieStore && !this.closeableWindows.has(tab.url)) {
+      const originStoreKey = this.getOriginStoreKey(new URL(tab.url));
+      this.storageArea.get([originStoreKey]).then((store) => {
+        if (store[originStoreKey]) {
+          browser.contextMenus.create({
+            id: this.menuRemoveId,
+            title: "Remove Open in This Container",
```
groovecoder: Maybe "Un-assign this site from this Container"

webextension/js/confirm-page.js
```diff
@@ -0,0 +1,25 @@
+const redirectUrl = new URL(window.location).searchParams.get("url");
+document.getElementById("redirect-url").textContent = redirectUrl;
+const redirectOrigin = new URL(redirectUrl).origin;
```
groovecoder: `const redirectHostname = new URL(redirectUrl).hostname` and `textContent = redirectHostname` will probably look cleaner?

webextension/confirm-page.html
```diff
@@ -0,0 +1,33 @@
...
+    <div class="title">
+      <h1 class="title-text">Hey, we heard you liked containers!</h1>
+    </div>
+    <form id="redirect-form">
+      <p>
+        A page told us to open:
+      </p>
+      <div id="redirect-url"></div>
+      <p>
+        You asked for <span id="redirect-origin"></span> to always open in <span id="container-name">this</span> container, are you sure you want to confirm?<br />
```
groovecoder: Maybe `You assigned <span id="redirect-hostname"></span> to <span id="container-name">this</span> container, ...`

webextension/css/popup.css
```diff
@@ -383,7 +383,7 @@ span ~ .panel-header-text {
 .panel-footer {
   align-items: center;
   background: #efefef;
-  block-size: 55px;
+  block-size: 54px;
```
groovecoder: Screw you 55th pixel!

webextension/manifest.json
```diff
@@ -19,8 +19,15 @@
...
+    "<all_urls>",
+    "activeTab",
     "cookies",
-    "tabs"
+    "contextMenus",
+    "notifications",
+    "storage",
+    "tabs",
+    "webRequestBlocking",
+    "webRequest"
```
groovecoder: Wow, we really **are** creating a tab-manager add-on.

webextension/css/popup.css
jonathanKingston: This was unrelated but part of the menu being unaligned.

webextension/manifest.json
jonathanKingston: We were just faking it till we made it before, now we are the real deal 🦊

webextension/confirm-page.html
jonathanKingston: Happy to change "asked for" to "added" as above however. I personally think that is less code'y.

webextension/background.js
jonathanKingston: It would however be inaccurate though as I'm currently not uplifting http web to behave the same as https on purpose. (port number I care less about)

webextension/background.js
jonathanKingston: Using uppercase here makes me cry a little (especially when it's not a const too), much like Hungarian notation does but I will not argue ;)

webextension/background.js
groovecoder: Ah, right ... http://gravatar.com and https://gravatar.com are treated differently. Alright then.

webextension/background.js
groovecoder: 200w_d

groovecoder:
Also, this needs a version bump to 2.1.0

webextension/background.js
```diff
@@ -1,3 +1,206 @@
...
+        } else {
+          actionName = "removed";
+          storageAction = this.storageArea.remove(info.pageUrl);
+        }
+        storageAction.then(() => {
+          const pageUrl = new window.URL(info.pageUrl);
+          browser.notifications.create({
+            type: "basic",
+            title: "Containers",
+            message: `Successfully ${actionName} ${pageUrl.origin} to always open in this container`,
```
groovecoder: What if we just remove the `${pageUrl.origin}` completely? A significant origin string overflows the notification interface, and the important feedback here seems to be the action. Can we assume knows the origin they just assigned, since they just assigned it?

kjozwiak:
Issues found while testing the `auto-container-origin` branch on `jonathanKingston/testpilot-containers`:

****
**_Bug 1:_**

"Always Open in This Contianer" is being displayed under the context menu under `about:privatebrowsing` when opening a PB window while the currently focussed tab is a container.

STR:

* open the personal container via the "+" new tab button and load any website
* while the personal container is the currently focussed tab, open a new PB window
* once the PB window opens, right click on the `about:privatebrowsing` page

screen shot 2017-03-29 at 3 53 46 pm

****

**_Bug 2_**

Once you've assigned an origin to a specific container, opening that particular origin within PB won't load the website and will produce the following error in the browser console:

`Error: Illegal to set non-private cookieStorageId in a private window`

Even though an origin has been assigned to a specific container, users should still be able to open that website within PB as those are two separate features. For example, lets say I assigned google.ca to my "Google" container but I want to search for something that won't appear in my history and I don't feel like creating a "throw away" container. I won't be able to use google.ca in this situation.

STR:

* open the personal container via the "+" new tab button and load any website
* open a PB window and load the same website within PB

****

jonathanKingston:
@kjozwiak Issues 1,2,6 are solved.

Issues 3,4,5 are not. I'm not actually sure 5 is solvable with the extension. 3,4 are not solvable in web extensions see: https://bugzilla.mozilla.org/show_bug.cgi?id=1352189 I will look into message passing from SDK when I get home (I was hoping to completely dog food web extensions here :( )

jonathanKingston:
```
TypeError: types.taggedTemplates.includes is not a function
    at isSafeNode (/home/jonathan/projects/testpilot-containers/node_modules/eslint-plugin-no-unescaped/lib/rules/enforce.js:91:37)
    at EventEmitter.AssignmentExpression (/home/jonathan/projects/testpilot-containers/node_modules/eslint-plugin-no-unescaped/lib/rules/enforce.js:152:18)
    at emitOne (events.js:82:20)
    at EventEmitter.emit (events.js:169:7)
    at NodeEventGenerator.enterNode (/home/jonathan/projects/testpilot-containers/node_modules/eslint/lib/util/node-event-generator.js:39:22)
    at CodePathAnalyzer.enterNode (/home/jonathan/projects/testpilot-containers/node_modules/eslint/lib/code-path-analysis/code-path-analyzer.js:607:23)
    at CommentEventGenerator.enterNode (/home/jonathan/projects/testpilot-containers/node_modules/eslint/lib/util/comment-event-generator.js:98:23)
    at Controller.enter (/home/jonathan/projects/testpilot-containers/node_modules/eslint/lib/eslint.js:928:36)
    at Controller.__execute (/home/jonathan/projects/testpilot-containers/node_modules/estraverse/estraverse.js:397:31)
    at Controller.traverse (/home/jonathan/projects/testpilot-containers/node_modules/estraverse/estraverse.js:501:28)
```

Linting rule I wrote is borked... should fix which is annoying as I want this rule to be replaced.

jonathanKingston:
Also... this latest patch fixes 3 and 4 and rebases.

groovecoder:
I would still be good to merge this, but I notice assigning a site to auto-open in a container now triggers a `page-requests-completed-per-tab` event with `pageRequestCount` of 0 for the confirm page being loaded in `firefox-default` container. 😐

kjozwiak:
Verified that bug#1, 2, 3, 4 and 6 have been fixed.

PR #389, opened 2017-03-16
WIP/DNM: Add page requests metric 209

groovecoder:


docs/metrics.md
```diff
@@ -49,7 +49,7 @@ whether they are actively interacting with it.
   * Average number of container tabs when new window was clicked
 * How many containers do users have hidden at the same time? (when should we measure this? each time a container is hidden?)
 * Do users pin container tabs? (do we have existing Telemetry for pinning?)
-* Do users change URLs in a container tab? (sounds like it could be a flood unless we only record the first URL change?)
+* Do users visit more pages in container tabs than non-container tabs?
```
groovecoder: To query for this, we will need to do something like:
```
SELECT AVG(pagerequestcount)
FROM containers_testpilottest
WHERE event = 'page-requests-completed-per-tab'
AND   totalcontainerscount = 0
GROUP BY uuid
```
compared to
```
SELECT AVG(pagerequestcount)
FROM containers_testpilottest
WHERE event = 'page-requests-completed-per-tab'
AND   totalcontainerscount > 0
GROUP BY uuid
```

groovecoder:
@bakulf or @jonathanKingston - can I get an r? on the JS code here?

webextension/manifest.json
```diff
@@ -19,8 +19,11 @@
   "homepage_url": "https://testpilot.firefox.com/",
 
   "permissions": [
+    "contextualIdentitities",
```
jonathanKingston: Is this even required? This is only available in 53 which is an issue.

webextension/manifest.json
```diff
@@ -19,8 +19,11 @@
...
     "cookies",
-    "tabs"
+    "tabs",
+    "webRequest",
```
jonathanKingston: Merge issues ahead... :feelsgood:

webextension/background.js
```diff
@@ -72,3 +73,28 @@ function disableAddon(tabId) {
   browser.browserAction.disable(tabId);
   browser.browserAction.setTitle({ tabId, title: "Containers disabled in Private Browsing Mode" });
 }
+
+browser.tabs.onCreated.addListener(tab => {
```
jonathanKingston: Is it worth having these methods self contained. I did the same for the assign code as it ended up getting mental. Seems like this is all related to tracking tabPageCounters etc.

jonathanKingston: Either that or for efficiency maybe we have an that does all the browser.apiName obj call methods to other objects for both the assignment and counting... however I dunno if we have a perf issue.

Like:
```
const apiObj = {
  init() {
    browser.webRequest.onCompleted.addListener((details) => {
      tabPageCounters.request.onComplete(....)
      assignManager.calculateContextMenu(...)
    });
  }
}
```

webextension/manifest.json
groovecoder: This is how I get to `tab.cookieStoreId` right?

webextension/background.js
groovecoder: Makes more sense to contain `browser.*` API calls when they may share some code and/or scope. Like the `themeManager`. So something more like ...
```
const TabPageCountManager = {
  init() {
    this.counter = {};
    browser.tabs.onCreated.addListener(this.initTabPageloadCounter);
    browser.tabs.onRemoved.addListener(this.sendTabPageloadCountAndDelete);
    browser.webRequest.onCompleted.addListener(this.incrementTabPageloadCount, {urls: ["<all_urls>"], types: ["main_frame"]});
  },
  initTabPageloadCounter() { ... },
  sendTabPageloadCountAndDelete() { ... },
  incrementTabPageloadCount() { ... }
}
```

groovecoder:
Refactored a new `tabPageCounter`.

webextension/manifest.json
groovecoder: Sure enough, didn't need this. Just needed the `cookies` permission.

webextension/background.js
```diff
@@ -43,7 +44,41 @@ const themeManager = {
   }
 };
 
+const tabPageCounter = {
+  init() {
+    browser.tabs.onCreated.addListener(this.initTabCounter);
+    browser.tabs.onRemoved.addListener(this.sendTabCountAndDelete);
```
jonathanKingston: Does this capture window closures?

webextension/background.js
```diff
@@ -43,7 +44,41 @@ const themeManager = {
...
+    browser.webRequest.onCompleted.addListener(this.incrementTabCount, {urls: ["<all_urls>"], types: ["main_frame"]});
```
jonathanKingston: Do you care about tab.tabId might be -1 for non rendered tabs?

webextension/background.js
```diff
@@ -1,3 +1,4 @@
+const tabPageRequestCounter = {};
```
jonathanKingston: This may as well be internal to tabPageCounter no?

webextension/background.js
```diff
@@ -43,7 +44,41 @@ const themeManager = {
...
+      method: "sendTelemetryPayload",
+      event: "page-requests-completed-per-tab",
+      userContextId: tabPageRequestCounter[tab].cookieStoreId,
+      pageRequestCount: tabPageRequestCounter[tab].pageRequests
+    });
+    delete tabPageRequestCounter[tab.id];
+  },
+
+  incrementTabCount(details) {
+    browser.tabs.get(details.tabId).then(tab => {
```
jonathanKingston: Are we (singleArg) => {} or singleArg => {}. One of us will have to change. Seems we are inconsistent anyway :(

webextension/background.js
groovecoder: Just checked in about:telemetry and yes - it does.

webextension/background.js
groovecoder: What's an example of a non-rendered tab I can spot-check?

webextension/background.js
groovecoder: I mildly prefer NOT to have the extra `(` and `)` characters in the code just to reduce character noise. But I don't care too much.

webextension/background.js
groovecoder: Moved.

PR #414, opened 2017-04-04
Fixing about: scheme assignment. Fixes #413

### LINKED ISSUE #413: Some about: protocol urls can assign to a container
It doesn't actually assign etc however it's confusing
selection_555

jonathanKingston:


webextension/background.js
```diff
@@ -160,7 +160,8 @@ const assignManager = {
   isTabPermittedAssign(tab) {
     // Ensure we are not an important about url
     // Ensure we are not in incognito mode
-    if (this.CLOSEABLE_WINDOWS.has(tab.url)
+    const url = URL(tab.url);
```
groovecoder: `TypeError: Constructor URL requires 'new'`

Issue #321, opened 2017-03-03
combine addon telemetry data with containers telemetry data


bakulf:
I just realized that the use I do of Containers, do not require the use of the addon. Mainly I have 2 containers tabs, both pinned, and session restore to load gmail in those tabs.
From a addon point of view, I don't use containers, but in practice, I am an active user.
Probably we should introduce a ping with containers tabs opened each X minutes. How does it sound?

groovecoder:
FWIW, if you have the add-on installed and you run Firefox, Test Pilot itself will ping that you have it installed and are actively using Firefox with it.

Periodic telemetry pings are problematic for at least a couple reasons:

1. Need to update the privacy notices of the addon
2. Periodic pings make back-end analysis harder - it's harder to measure active usage because we have to filter out the passive periodic pings from the active user-driven pings

Not impossible, but if the main concern is to make sure you're still counted as an "active user" - you should be, even with the usage pattern you describe here.

bakulf:
As discussed on IRC, we want to send a telemetry ping when the tab is activated.
@groovecoder do you want to take this issue?

groovecoder:
I've got https://github.com/mozilla/testpilot-containers/pull/389 for #209. But one of the issues there is that the page count pings are sent on `browser.tabs.onRemoved` events, which do not seem to fire when Firefox quits.

We may be able to build on top of #389 for this, but it will take a bit more planning.

groovecoder:
@bakulf - there's an [`idle` WebExtension API](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/idle) such that we could start a counter `idle.onStateChanged === "active"` and stop and send the ping `idle.onStateChanged in ["idle", "locked"]`. Would that cover your usage?

groovecoder:
@bakulf and comment here? would the `idle` API/event capture your usage?

bakulf:
Yes, that would work. And maybe we can send a telemetry ping when FF starts if there are containers pinned tabs on startup.

dannycolin:
Closing in favor of [bug 1276002](https://bugzilla.mozilla.org/show_bug.cgi?id=1276002). We definitely to readd some telemetry probes! :)

PR #435, opened 2017-04-06
for #321: page count telemetry per activity

groovecoder:
Adds a new `page-requests-completed-per-activity` event ping with `pageRequestCount` metric. Note: these pings are in addition to the `-per-tab` pings, so we will analyze them separately.

Steps to verify:
* Create some (non-container and/or container) tabs
* Make some page loads in some of the tabs
* Leave your computer idle for ~60s (don't close any of the tabs)
* Go to about:telemetry
  * Check "archived ping data"
  * The payload drop down should contain "*-testpilottest" options - pick some
  * Verify they have the expected `page-requests-completed-per-activity` event string, and the proper `userContextId` and `pageRequestCount` values
* Make some more page loads in the still-open tabs
* Go to about:telemetry
  * Check "archived ping data"
  * The payload drop down should contain "*-testpilottest" options - pick some
  * Verify they have the expected `page-requests-completed-per-tab` event string, and the proper `userContextId` and `pageRequestCount` values

### DIFF
--- a/webextension/background.js
+++ b/webextension/background.js
@@ -283,10 +283,10 @@ const messageHandler = {
     browser.idle.onStateChanged.addListener((newState) => {
       browser.tabs.query({}).then(tabs => {
         for (let tab of tabs) { // eslint-disable-line prefer-const
-          if (newState === "idle" || newState === "locked") {
+          if (newState === "idle") {
             tabPageCounter.sendTabCountAndDelete(tab.id, "user-went-idle");
-          } else if (newState === "active") {
-            tabPageCounter.initTabCounter(tab, "user-became-active");
+          } else if (newState === "active" && tab.active) {
+            tabPageCounter.initTabCounter(tab);
           }
         }
       }).catch(e => {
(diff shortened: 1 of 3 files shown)


groovecoder:
@jonathanKingston - can you help me figure out why (it seems) like the page requests are double-counted in this code? I always see multiples of 2 in the `page-requests-completed-*` pings, but I can't tell why each page request would be double-counted? Maybe I'm just cross-eyed from looking at the code too long on my own.

groovecoder:
I figured out the double-count issue. I left 2 calls to `incrementTabCount` even though I changed the function to increment both counts.

webextension/background.js
```diff
@@ -280,6 +280,20 @@ const messageHandler = {
       });
     });
 
+    browser.idle.onStateChanged.addListener((newState) => {
+      browser.tabs.query({}).then(tabs => {
+        for (let tab of tabs) { // eslint-disable-line prefer-const
+          if (newState === "idle" || newState === "locked") {
```
groovecoder: Yes, `browser.idle.onStateChanged` seems to be a browser event - not a tab event. When the user goes idle, I send and reset all the tabs' "`per-activity`" page counters. The scenario here is:

1. User loads a few pages in a container tab
2. User switches to some other tab(s)
3. User goes idle

I'll see if I can move the `initTabCounter(tab, "user-became-active")` call into `browser.tabs.onActivated` so we only initialize an activity counter for the tab when the user activates it.

I'd be fine to drop `locked` state.

webextension/background.js
```diff
@@ -333,30 +347,62 @@ const themeManager = {
...
-      pageRequestCount: this.counter[tabId].pageRequests
-    });
-    delete this.counter[tabId];
+  sendTabCountAndDelete(tabId, why = "user-closed-tab") {
+    if (why === "user-closed-tab") {
+      browser.runtime.sendMessage({
+        method: "sendTelemetryPayload",
+        event: "page-requests-completed-per-tab",
+        userContextId: this.counters[tabId].tab.cookieStoreId,
+        pageRequestCount: this.counters[tabId].tab.pageRequests
```
groovecoder: Good idea; I'll try to add that.

webextension/background.js
```diff
@@ -333,30 +347,62 @@ const themeManager = {
...
+      this.counters[tabId].activity = {
+        "cookieStoreId": this.counters[tabId].tab.cookieStoreId,
+        "pageRequests": 0
+      };
+    }
   },
 
   incrementTabCount(tab) {
-    this.counter[tab.id].pageRequests++;
+    this.counters[tab.id].tab.pageRequests++;
```
groovecoder: I don't **think** I'm counting the delta between tab lifetime and activity states.

The 2 page counters here are:

* "per-tab" counter
* "per-activity" counter

### per-tab
1. User opens container tab
2. User visits 3 different pages
3. User closes container tab
   * We send a `page-requests-completed-per-tab` ping with the value 3

### per-activity
1. User opens a container tab
2. User visits 1 page in the container tab
3. User never closes tab
4. User goes "idle"
   * We send a `page-requests-completed-per-activity` ping with the value 1

webextension/background.js
```diff
@@ -333,30 +347,62 @@ const themeManager = {
...
+    } else if (why === "user-went-idle") {
+      browser.runtime.sendMessage({
+        method: "sendTelemetryPayload",
+        event: "page-requests-completed-per-activity",
+        userContextId: this.counters[tabId].activity.cookieStoreId,
+        pageRequestCount: this.counters[tabId].activity.pageRequests
+      });
+      // When we send the ping because the user went idle,
+      // only reset the 'activity' counter
+      this.counters[tabId].activity = {
```
groovecoder: Well-spotted.

You're right there shouldn't be any counts in between. But, I'll call it defensive coding for a future when someone might call `initTabCounter(tab, "user-became-active")` without a previous call to `sendTabCountAndDelete(tab, "user-went-idle").

I just caught a bug in our Telemetry data because [we call `showTabs` from within `openTab`](https://github.com/mozilla/testpilot-containers/blob/3e657a2e8d952da30c963ae929bc8a774db9a425/index.js#L861), resulting in duplicate `show-tabs` event pings that I had to filter out in the SQL query. :cry:

webextension/background.js
groovecoder: It's getting complicated, so I filed https://github.com/mozilla/testpilot-containers/issues/446 for this.

groovecoder:
Okay, I dropped the `locked` idle state change event, and I simplified the tab counter code so that it won't send a 0-count activity ping for *every* tab. I.e., it only sends activity page-count pings for tabs that have actually been activated.

r? @jonathanKingston

PR #436, opened 2017-04-07
Fixing links that redirect creating another tab when creating an assi…

### LINKED ISSUE #421: opening origins that belong to specific containers via right click on certain engines opens an extra tab
When opening an origin from google.com that's already assigned to a specific container in a different container using  "Open Link in New Container Tab" via the right click context menu, another tab is opened in the second container which is blank and never closed.

Reproduced using: Google, Yahoo
Couldn't reproduce using: DDG, Bing

STR:

* install the latest containers experiment (2.1.1 used in this case)
* open a new personal container and visit reddit.com
* right click on reddit and select "Always Open in This Container"
* open a default contianer tab and visit google.com
* search for reddit
* on the main search result, right click on reddit -> Open Link in New Container Tab -> Work

You'll get a new tab that's opened in the Work container that's completely blank with the following URL which doesn't get removed:

https://www.google.ca/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&cad=rja&uact=8&ved=0ahUKEwiYo6jar4vTAhXs6oMKHTt3CMcQFggaMAA&url=https%3A%2F%2Fwww.reddit.com%2F&usg=AFQjCNFTmzNOTpLfW33wey8AIySJgUmv_g&bvm=bv.151426398,d.amc

screen shot 2017-04-04 at 1 56 12 pm

screen shot 2017-04-04 at 1 56 25 pm

jonathanKingston:
…gned tab. Fixes #421 

Using a setTimeout to fix this is likely not the best at all for this however:

A better solution would be using this to track when new tabs get created from links, then tracking through the redirect:
https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webNavigation/onCreatedNavigationTarget

Another would be using history of the dangling tab to see how long the user has had that page open rather than a setTimeout, however injecting a content script into every page to capture this wouldn't be much better either. Perhaps we should request for an API for tab.history() or something?

This might be able to be cleaned up in 53 when the tab created could be in the same container as we can query for container names and display icons there.

### DIFF
--- a/webextension/background.js
+++ b/webextension/background.js
@@ -127,8 +127,21 @@ const assignManager = {
 
         this.reloadPageInContainer(options.url, siteSettings.userContextId, tab.index + 1, siteSettings.neverAsk);
         this.calculateContextMenu(tab);
-        // If the user just opened the tab, we can auto close it
-        if (this.CLOSEABLE_WINDOWS.has(tab.url)) {
+
+        /* Removal of existing tabs:
+           We aim to open the new assigned container tab / warning prompt in it's own tab:
+             - As the history won't span from one container to another it seems most sane to not try and reopen a tab on history.back()
+             - When users open a new tab themselves we want to make sure we don't end up with three tabs as per: https://github.com/mozilla/testpilot-containers/issues/421
+           If we are coming from an internal url that are used for the new tab page (CLOSEABLE_WINDOWS), we can safely close as user is unlikely losing history
+           Detecting redirects on "new tab" opening actions is pretty hard as we don't get tab history:
+           - Redirects happen from Short URLs and tracking links that act as a gateway
+           - Extensions don't provide a way to history crawl for tabs, we could inject content scripts to do this
+               however they don't run on about:blank so this would likely be just as hacky.
+           We capture the time the tab was created and close if it was within the timeout to try to capture pages which haven't had user interaction or history.
+        */
+        if (this.CLOSEABLE_WINDOWS.has(tab.url)
+            || (messageHandler.lastCreatedTab
+            && messageHandler.lastCreatedTab.id === tab.id)) {
           browser.tabs.remove(tab.id);
         }
         return {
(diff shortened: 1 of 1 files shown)


groovecoder:
Spot-check looks good following the steps in #421 and I spot-checked it thru a t.co redirect too.

Instead of starting a timer when the tab is created, could we use an [`onHeadersReceived`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webRequest/onHeadersReceived) handler to detect the redirect requests *BEFORE* the assigned container requests?

Something like this?

```
browser.webRequest.onHeadersReceived.addListener(details => {
  if (details.statusCode in ["301", "302"]) {
    for (let header of details.responseHeaders) {
      if (header.name === "Location" && assignManager.get(header.value)) {
        reloadPageInContainer(header.value, siteSettings.userContextId, details.tab.index+1, siteSettings.neverAsk);
        details.tab.remove();
      }
    }
  }
}, {urls: ["<all_urls>"], types: ["main_frame"]}, ["responseHeaders"]);
```

groovecoder:
Per @jonathanKingston - the main offenders (e.g., Google) don't actually use 30x Location: redirects, so using `onHeadersReceived` wouldn't work here.

Will have to live with the `setTimeout` approach in Test Pilot.

jonathanKingston:
I knew I had to check something before leaving. Will double check before we
release even though it's merged as mentioned the solution is neater and
similar to the other direction I was taking. Sorry for the delay
@groovecoder!

jonathanKingston:
Sorry @groovecoder  for the delay here, google uses this example code:
```
<script>window.googleJavaScriptRedirect=1</script><META name="referrer" content="origin"><script>var n={navigateTo:function(b,a,d){if(b!=a&&b.google){if(b.google.r){b.google.r=0;b.location.href=d;a.location.replace("about:blank");}}else{a.location.replace(d);}}};n.navigateTo(window.parent,window,"https://www.mozilla.org/nl/firefox/new/");
</script><noscript><META http-equiv="refresh" content="0;URL='https://www.mozilla.org/nl/firefox/new/'"></noscript>
```

No 300ish status code is given.

groovecoder:
Thanks for closing the loop on this @jonathanKingston.

PR #447, opened 2017-04-12
Adding in context menu to tab context menu. Fixes #424.

### LINKED ISSUE #424: add assignment menu on the tab contextual menu
to ease discover of the menu (and ease its use), I would suggest to add the `always open in this container` menu in the tab contextual menu additionally to the page menu

jonathanKingston:


webextension/background.js
```diff
@@ -205,7 +205,7 @@ const assignManager = {
           id: menuId,
           title: `${prefix} Always Open in This Container`,
           checked: true,
-          contexts: ["all"],
+          contexts: ["all", "tab"],
```
groovecoder: This only works in 53 right? So we should merge + release next week April 19?

jonathanKingston: Doesn't cause issues in stable from what I can see though. But yeah 53 only.

Issue #463, opened 2017-04-25
container functionality lost when FX crashes in the same session that the experiment has been installed in


kjozwiak:
Whenever FX crashes and a user has the Containers experiment installed, they'll lose all the functionality that test pilot has installed. Examples:

* Containers button under the toolbar completely disappears
* Always Open in This Container disappears from the context menu
* The containers drop down from the new tab "+" button disappears

You'll also get the following error message in the browser console:

> TypeError: ss.storage.savedConfiguration is undefined - runner.js:1663
> Stack trace:
> init@resource://gre/modules/commonjs/toolkit/loader.js -> resource://testpilot-containers/index.js:171:1
> exports.main@resource://gre/modules/commonjs/toolkit/loader.js -> resource://testpilot-containers/index.js:1590:3
> run@resource://gre/modules/commonjs/toolkit/loader.js -> resource://gre/modules/commonjs/sdk/addon/runner.js:153:7
> startup/</<@resource://gre/modules/commonjs/toolkit/loader.js -> resource://gre/modules/commonjs/sdk/addon/runner.js:87:9
> process@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:917:23
> walkerLoop@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:801:7
> scheduleWalkerLoop/<@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:737:11
> 
> TypeError: ss.storage.savedConfiguration is undefined - Promise-backend.js:920
> Stack trace:
> init@resource://gre/modules/commonjs/toolkit/loader.js -> resource://testpilot-containers/index.js:171:1
> exports.main@resource://gre/modules/commonjs/toolkit/loader.js -> resource://testpilot-containers/index.js:1590:3
> run@resource://gre/modules/commonjs/toolkit/loader.js -> resource://gre/modules/commonjs/sdk/addon/runner.js:153:7
> startup/</<@resource://gre/modules/commonjs/toolkit/loader.js -> resource://gre/modules/commonjs/sdk/addon/runner.js:87:9
> process@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:917:23
> walkerLoop@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:801:7
[...]

PR #470, opened 2017-04-28
Recovering after a crash - issue #463

bakulf:


jonathanKingston:
This looks ok, however I would like to double check an upgrade process just to be sure.

PR #471, opened 2017-04-28
Move SDK code into web extension code

jonathanKingston:
As part of the gradual move away from SDK code, I have moved remove, create and update to the popup code.

Everything seems to be working just fine still, I wasn't able to test telemetry as my setup is borked for that at the moment.

I will keep updating this with other methods as I have them unless people are ok for this to merge etc.

### DIFF
--- a/webextension/js/popup.js
+++ b/webextension/js/popup.js
@@ -152,7 +152,7 @@ const Logic = {
       tabs.forEach((tab) => {
         cb(tab);
       });
-    });
+    }).catch((e) => {throw e;});
   },
 
   _containers(userContextId) {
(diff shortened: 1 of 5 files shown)


webextension/js/popup.js
```diff
@@ -141,6 +141,65 @@ const Logic = {
...
+    browser.tabs.query({
+      cookieStoreId: Logic.cookieStoreId(userContextId)
+    }).then((tabs) => {
+      tabs.forEach((tab) => {
+        cb(tab);
+      });
+    }).catch((e) => {throw e;});
+  },
+
+  _containers(userContextId) {
```
groovecoder: Rename to `_containerTabs` to be more clear what's returned.

webextension/js/popup.js
```diff
@@ -141,6 +141,65 @@ const Logic = {
...
+    });
+
+    const removeTabsPromise = Logic._containers(userContextId).then((tabs) => {
+      const tabIds = tabs.map((tab) => tab.id);
+      return browser.tabs.remove(tabIds);
+    });
+
+    return removeTabsPromise.then(() => {
+      const removed = browser.contextualIdentities.remove(Logic.cookieStoreId(userContextId));
+      // Send delete event to webextension/background.js
```
groovecoder: This seems to be double-sending the `delete-container` Telemetry message? (See line 178 above)

groovecoder:
Fixed my own nit-pick and submitted https://github.com/mozilla/testpilot-containers/pull/479.

webextension/js/popup.js
jonathanKingston: Nah just looked into this... I know this is an old PR however the eventName was reused but the key was different. In the latest PR. I have moved more code to use the "method" key which should make it more consistent rather than type. I also made the methodNames all camelcase rather than the telemetry which is dasherized.

PR #478, opened 2017-05-02
Add lib/shield to enable shield study

groovecoder:
I'd like to get some more eyes on this. In summary this ...

1. Brings [`shield-studies-addon-utils`](https://www.npmjs.com/package/shield-studies-addon-utils) into `lib/shield` directory
2. Brings [`experiment.js`](https://github.com/mozilla/testpilot/blob/master/addon/src/lib/metrics/experiment.js) from Test Pilot into `lib/testpilot` directory to give us the same `TelemetryController` wrapper.
3. Configures a skeleton proof-of-concept study in `study.js`
4. Starts the study after the `webExtension.startup()`
5. Includes `shield run` instructions in the `README.md`
6. Adds a `npm run build-shield` to create a `shield-study-containers.xpi` version of the add-on

Right now the study just opens an orientation tab with a message that corresponds to the variation into which shield puts the user. (See the new [Shield Containers Product Hypothesis](https://docs.google.com/document/d/1vMD-fH_5hGDDqNvpRZk12_RhCN2WAe4_yaBamaNdtik/edit#)) for variations.

Need to test especially:

1. That the code still runs with/under Test Pilot
2. That the code also runs with shield


study.js
```diff
@@ -0,0 +1,42 @@
+const self = require("sdk/self");
```
bakulf: No header here?

groovecoder: Added

study.js
```diff
@@ -0,0 +1,46 @@
...
+const tabs = require("sdk/tabs");
+
+const shield = require("./lib/shield/index");
+
+const studyConfig = {
+  name: self.addonId,
+  days: 28,
+  surveyUrls: {
+  },
+  variations: {
```
gregglind: need filling with the 'feature init' code, whatever that is.

study.js
```diff
@@ -0,0 +1,46 @@
...
+const self = require("sdk/self");
+const { when: unload } = require("sdk/system/unload");
+const tabs = require("sdk/tabs");
+
+const shield = require("./lib/shield/index");
+
+const studyConfig = {
+  name: self.addonId,
+  days: 28,
+  surveyUrls: {
```
gregglind: this will do no surveys anywhere, as written.  If you want them, this needs fill.

study.js
```diff
@@ -0,0 +1,46 @@
...
+    // If the user already has testpilot-containers extension, they are in the
+    // Test Pilot experiment, so exclude them.
+    return super.isEligible();
+  }
+
+  whenEligible () {
+  }
+
+  whenInstalled () {
+    tabs.open(`data:text/html, Thank you for helping us study Containers in Firefox. You are in the ${this.variation} variation.`);
```
gregglind: you will want to fill this, if you need it at all :)

study.js
groovecoder: Yup, still planning to fill these in.

groovecoder:
I'll check this again tomorrow.

There's a known bug where the telemetry pings don't work in shield, because `testpilot-metrics.js` needs the `experiment.js` module provided in the Test Pilot add-on. But if this causes no issues for Test Pilot, I'd like to merge this ASAP before bit-rot.

jonathanKingston:
> There's a known bug where the telemetry pings don't work in shield, because testpilot-metrics.js needs the experiment.js module provided in the Test Pilot add-on. But if this causes no issues for Test Pilot, I'd like to merge this ASAP before bit-rot

Ah ok so that other file is a shim to do nothing when in shield?
Nothing looks bad at all, I thought we were going to share more of the wrapping telemetry code.

Maybe a quick chat over video before it rots and I lose understanding?

groovecoder:
So ...

1. `index.js` always uses `testpilot-metrics.js`
2. `testpilot-metrics.js` will check if the Test Pilot add-on is present
  * If it is, we don't need to include/start the `experiment.js` code
  * If it is NOT, we need to include/start the `experiment.js` code

jonathanKingston:
Thanks for the explanation.... Little tired lately so I needed to be on the
same page. R+ then.

groovecoder:
Okay, this is completely ready. `README.md` should have the right instructions.

The existing build & deploy setup for Test Pilot is still the default. I added a new command: `npm run build-shield` that changes `package.json` `name` to `shield-study-containers`, builds the `.xpi`, and then changes the `package.json` `name` back to `testpilot-containers`.

This means the test pilot add-on will send pings with `@testpilot-containers` and the shield add-on will send pings with `@shield-study-containers`.

PR #479, opened 2017-05-04
move SDK code into webextension

groovecoder:
supersedes https://github.com/mozilla/testpilot-containers/pull/471 with my nit-pick fixed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -196,11 +196,10 @@ const ContainerService = {
       "moveTabsToWindow",
       "queryIdentities",
       "getIdentity",
-      "createIdentity",
-      "updateIdentity",
       "getPreference",
       "sendTelemetryPayload",
       "getTheme",
+      "refreshNeeded",
       "forgetIdentityAndRefresh",
       "checkIncompatibleAddons"
     ];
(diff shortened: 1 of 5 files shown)


jonathanKingston:
Don't merge this. I need to look into the issues mentioned in #471 on PTO at the moment but will fix when I get back. I have more conversion to merge too.

jonathanKingston:
Also thanks @groovecoder!

index.js
```diff
@@ -911,62 +910,6 @@ const ContainerService = {
...
-    });
-
-    for (let arg of [ "name", "color", "icon"]) { // eslint-disable-line prefer-const
-      if (!(arg in args)) {
-        return Promise.reject("createIdentity must be called with " + arg + " argument.");
-      }
-    }
-
-    const color = this._fromNameToColor(args.color);
-    const icon = this._fromNameToIcon(args.icon);
```
bakulf: Do we still need these 2 methods?

webextension/js/popup.js
```diff
@@ -141,6 +141,58 @@ const Logic = {
...
+    browser.runtime.sendMessage({
+      method: "sendTelemetryPayload",
+      event: eventName,
+      userContextId
+    });
+
+    const removeTabsPromise = Logic._containers(userContextId).then((tabs) => {
+      const tabIds = tabs.map((tab) => tab.id);
+      return browser.tabs.remove(tabIds);
+    });
```
bakulf: can you just do:

return Logic._containers(userContextId).then((tabs) => { .. })
           .then(() => { ... the rest? })

index.js
jonathanKingston: Convert still uses these. I think what I have on my laptop removes this through. Will submit Monday

webextension/js/popup.js
jonathanKingston: Sure just not a fan of multiple promise chains is all. The tabs remove is promise ready and we are waiting on that.

jonathanKingston:
I'm picking this up, hope that is ok.

jonathanKingston:
Quoting @groovecoder:

Looks really good. Couple comments. Also, I found a bug:

1. Create a custom container
2. Open a tab in the container
3. Close all other tabs except that container tab
4. Delete the custom container

Expected/existing results:
The container is deleted, and the tab is destroyed, effectively closing the window.

Actual/new results:
The tab is destroyed and the window is closed, but when I open a new window, the custom container is still there - it wasn't deleted.


@andymckay is gonna love this, I can code around it but... basically the cause of this is because the popup closes a tab which in turn kills the window, which in turn kills the code it is running itself mid resolving the promise for a tab removal. It is probably an edge case for web extensions however certainly a pain in the arse.


I recoded this to be more reliable but still the window closing seems to trump everything.
```
  removeIdentity(userContextId) {
    const eventName = "delete-container";
    if (!userContextId) {
      return Promise.reject("removeIdentity must be called with userContextId argument.");
    }

    this.sendTelemetryPayload({
      event: eventName,
      userContextId
    });

    const removeTabsPromise = Logic._containerTabs(userContextId).then((tabs) => {
      const tabIds = tabs.map((tab) => tab.id);
      return browser.tabs.remove(tabIds);
    });

    return removeTabsPromise.then(() => { // << code never gets past this....
      const removed = browser.contextualIdentities.remove(Logic.cookieStoreId(userContextId));
      // Send delete event to webextension/background.js
      browser.runtime.sendMessage({
        type: eventName,
        message: {userContextId}
      });
      browser.runtime.sendMessage({
        method: "forgetIdentityAndRefresh"
      }).then(() => {
        return removed;
      }).catch((e) => {throw e;});
    });
  },
```

@andymckay would you like me to make a simpler test case and verify if chrome has an issue when not using promises etc?

PR #485, opened 2017-05-08
Remove sdk

jonathanKingston:
supersedes #479 which supersedes #471 
... superception

The logic has been moved into the background script which should last even if the window is closed.

Notice the comment on 628 of webextension/js/popup.js: https://github.com/mozilla/testpilot-containers/compare/master...jonathanKingston:remove-sdk#diff-0effdbf34829f2ab5dac6ed4d8906066R628 basically the response from the background is ignored as the window that has the popup in it gets removed.

### DIFF
--- a/webextension/background.js
+++ b/webextension/background.js
@@ -74,17 +74,6 @@ const assignManager = {
   },
 
   init() {
-    browser.runtime.onMessage.addListener((m) => {
-      switch (m.type) {
-      case "delete-container":
-        assignManager.deleteContainer(m.message.userContextId);
-        break;
-      case "never-ask":
-        this._neverAsk(m);
-        break;
-      }
-    });
-
     browser.contextMenus.onClicked.addListener((info, tab) => {
       const userContextId = this.getUserContextIdFromCookieStore(tab);
       // Mapping ${URL(info.pageUrl).hostname} to ${userContextId}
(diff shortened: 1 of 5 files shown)


webextension/background.js
```diff
@@ -251,13 +237,101 @@ const assignManager = {
...
   // We use this to catch redirected tabs that have just opened
   // If this were in platform we would change how the tab opens based on "new tab" link navigations such as ctrl+click
   LAST_CREATED_TAB_TIMER: 2000,
 
   init() {
+    browser.runtime.onMessage.addListener((m) => {
+      let response;
+
+      switch (m.method) {
+      case "deleteContainer":
```
groovecoder: nit: indentation

webextension/background.js
```diff
@@ -251,13 +237,101 @@ const assignManager = {
...
+        const removed = browser.contextualIdentities.remove(this.cookieStoreId(userContextId));
+        removed.then(() => {
+          assignManager.deleteContainer(userContextId);
+          browser.runtime.sendMessage({
+            method: "forgetIdentityAndRefresh"
+          }).then(() => {
+            resolve({done: true, userContextId});
+          }).catch((e) => {throw e;});
+        }).catch((e) => {throw e;});
+      }).catch((e) => {throw e;});
```
groovecoder: 3oedus8leayggiwtco

jonathanKingston: Baahaha

jonathanKingston: I wonder if there is a neater way to do this?

webextension/background.js
jonathanKingston: This is the consistent (with the other two switches) way that we have setup case statements... it's ugly as crap I know. I can edit the rule as a follow up perhaps?

groovecoder: Meh ... can do that later.

PR #487, opened 2017-05-08
Interim WIP patch of more removal

jonathanKingston:
Backing this up... super buggy at the moment. Adding PR for visibility only.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,6 +4,7 @@
 
 const XUL_NS = "http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul";
 const DEFAULT_TAB = "about:newtab";
+const LOOKUP_KEY = "$ref";
 
 const SHOW_MENU_TIMEOUT = 100;
 const HIDE_MENU_TIMEOUT = 300;
@@ -73,41 +74,43 @@ Cu.import("resource:///modules/CustomizableWidgets.jsm");
 Cu.import("resource:///modules/sessionstore/SessionStore.jsm");
 Cu.import("resource://gre/modules/Services.jsm");
 
-// ----------------------------------------------------------------------------
 // ContextualIdentityProxy
 
 const ContextualIdentityProxy = {
   getIdentities() {
+    let response;
     if ("getPublicIdentities" in ContextualIdentityService) {
-      return ContextualIdentityService.getPublicIdentities();
+      response = ContextualIdentityService.getPublicIdentities();
+    } else {
+      response = ContextualIdentityService.getIdentities();
     }
 
-    return ContextualIdentityService.getIdentities();
-  },
-
-  getUserContextLabel(userContextId) {
-    return ContextualIdentityService.getUserContextLabel(userContextId);
+    return response.map((identity) => {
+      return this._convert(identity);
+    });
   },
 
   getIdentityFromId(userContextId) {
+    let response;
     if ("getPublicIdentityFromId" in ContextualIdentityService) {
-      return ContextualIdentityService.getPublicIdentityFromId(userContextId);
+      response = ContextualIdentityService.getPublicIdentityFromId(userContextId);
+    } else {
+      response = ContextualIdentityService.getIdentityFromId(userContextId);
     }
-
-    return ContextualIdentityService.getIdentityFromId(userContextId);
+    if (response) {
+      return this._convert(response);
+    }
+    return response;
   },
 
-  create(name, icon, color) {
-    return ContextualIdentityService.create(name, icon, color);
+  _convert(identity) {
+    return {
+      name: ContextualIdentityService.getUserContextLabel(identity.userContextId),
+      icon: identity.icon,
+      color: identity.color,
+      userContextId: identity.userContextId,
+    };
   },
-
-  update(userContextId, name, icon, color) {
-    return ContextualIdentityService.update(userContextId, name, icon, color);
-  },
-
-  remove(userContextId) {
-    return ContextualIdentityService.remove(userContextId);
-  }
 };
 
 // ----------------------------------------------------------------------------
@@ -149,7 +152,7 @@ const ContainerService = {
         // Maybe rename the Banking container.
         const identity = ContextualIdentityProxy.getIdentityFromId(3);
         if (identity && identity.l10nID === "userContextBanking.label") {
-          ContextualIdentityProxy.update(identity.userContextId,
+          ContextualIdentityService.update(identity.userContextId,
                                          "Finance",
                                          identity.icon,
                                          identity.color);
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -74,43 +74,6 @@ Cu.import("resource:///modules/sessionstore/SessionStore.jsm");
 Cu.import("resource://gre/modules/Services.jsm");
 
 // ----------------------------------------------------------------------------
-// ContextualIdentityProxy
-
-const ContextualIdentityProxy = {
```
jonathanKingston: @bakulf I think we are safe to remove all this now right?

index.js
```diff
@@ -653,7 +624,7 @@ const ContainerService = {
         userContextId: args.userContextId,
         url: object.url,
         nofocus: args.nofocus || false,
-        window: args.window || null,
+   //TOCHECK     window: args.window || null,
```
jonathanKingston: @groovecoder we need to check the default behaviour of tab.create works as if this is passed. I suspect there are some edge cases I have not tested yet but it seems to work.

index.js
```diff
@@ -840,58 +811,33 @@ const ContainerService = {
...
+  },
 
-      let promise;
-      if (userContextId) {
-        promise = this.showTabs(args);
-      } else {
-        promise = Promise.resolve(null);
-      }
+  // Identities management
+  queryIdentitiesState() {
```
jonathanKingston: Minified the dependency on native code here, only index.js calls queryIdentities and external calls queryIdentitesState...

jonathanKingston: The follow up to this can move out the state tracking into the background page after hide and sort is moved there too.

webextension/background.js
```diff
@@ -289,6 +289,44 @@ const backgroundLogic = {
...
 
+  openTab(options) {
+    const url = options.url || undefined;
+    const userContextId = ("userContextId" in options) ? options.userContextId : 0;
+    const active = ("nofocus" in options) ? options.nofocus : true;
+/*
+TODO check if we are opening in the correct window
+    return new Promise(resolve => {
+      if ("window" in args && args.window) {
+        resolve(args.window);
```
jonathanKingston: Need to check if the window finding code behaves the same, currently tab.create seems to behave the same.

webextension/background.js
```diff
@@ -289,6 +289,44 @@ const backgroundLogic = {
...
+    // Only send telemetry for tabs opened by UI - i.e., not via showTabs
+    if (source && userContextId) {
+      this.sendTelemetryPayload({
+        "event": "open-tab",
+        "eventSource": source,
+        "userContextId": userContextId,
+//IORE FOR NOW          "clickedContainerTabCount": this._containerTabCount(userContextId)
+      });
+    }
+    // Autofocus url bar will happen in 54: https://bugzilla.mozilla.org/show_bug.cgi?id=1295072
```
jonathanKingston: @groovecoder this might be a reason to delay the release of this code though.

webextension/background.js
```diff
@@ -289,6 +289,44 @@ const backgroundLogic = {
...
+*/
+    const source = ("source" in options) ? options.source : null;
+
+    // Only send telemetry for tabs opened by UI - i.e., not via showTabs
+    if (source && userContextId) {
+      this.sendTelemetryPayload({
+        "event": "open-tab",
+        "eventSource": source,
+        "userContextId": userContextId,
+//IORE FOR NOW          "clickedContainerTabCount": this._containerTabCount(userContextId)
```
jonathanKingston: Ah crap, I need to fix this.

webextension/background.js
```diff
@@ -289,6 +289,44 @@ const backgroundLogic = {
...
+      });
+    }
+    // Autofocus url bar will happen in 54: https://bugzilla.mozilla.org/show_bug.cgi?id=1295072
+
+    // What is this doing? promise = this.showTabs(args);
```
jonathanKingston: I couldn't really work out what this was supposed to be doing, however there is telemetry attached to this method when it was index.js

webextension/js/popup.js
```diff
@@ -85,13 +85,27 @@ const Logic = {
...
   refreshIdentities() {
-    return browser.runtime.sendMessage({
-      method: "queryIdentities"
-    })
-    .then(identities => {
-      this._identities = identities;
-    });
+    return Promise.all([
+      browser.contextualIdentities.query({}),
+      browser.runtime.sendMessage({
```
jonathanKingston: This code merges the two together for now, just so we can reduce the code later easier whilst safely removing our dependency on it here. As mentioned earlier this can be cleared up in a follow up when we have hide and sort here.

webextension/js/popup.js
```diff
@@ -552,7 +568,7 @@ Logic.registerPanel(P_CONTAINER_EDIT, {
     return browser.runtime.sendMessage({
       method: "createOrUpdateContainer",
       message: {
-        userContextId: identity.userContextId || false,
+        userContextId: Logic.userContextId(identity.cookieStoreId) || false,
```
jonathanKingston: Some of these can probably be cleared up later when we have all the code ported over. I just want to reduce change as much as possible. Neatening etc can happen later.

index.js
jonathanKingston: This needs Firefox 54 for removal... we also need to change it to the "Public" method instead.

The upgrade code in the install process doesn't seem to work on >54 so I need to make it support both.

jonathanKingston:
@groovecoder backing this up again, this is much closer but has the mentioned upgrade issues still where some old custom containers end with black or missing icons. This also seems to happen in 53 too. The update code might not be firing perhaps... will check.

The only other things are:
- window args removed from openTab, but still seems to work fine
- removal of showTabs from openTab code... is this needed besides the telemetry?

webextension/background.js
```diff
@@ -339,6 +386,9 @@ const messageHandler = {
       case "lightweight-theme-changed":
         themeManager.update(m.message);
         break;
+      case "open-tab":
+        backgroundLogic.openTab(m.message);
+        break;
```
groovecoder: `openTab` and `open-tab` are confusing here. Can we comment the difference?

index.js
```diff
@@ -167,6 +150,18 @@ const ContainerService = {
...
+      const newIcon = this._fromIconToName(identity.icon);
+      const newColor = this._fromColorToName(identity.color);
+      if (newIcon !== identity.icon || newColor !== identity.color) {
+        ContextualIdentityService.update(identity.userContextId,
+                                       ContextualIdentityService.getUserContextLabel(identity.userContextId),
+                                       newIcon,
+                                       newColor);
+      }
+    });
+
```
groovecoder: I'm not sure why we need this block?

jonathanKingston:
The code block is to fix broken icons in the database. If you check my
addon sea containers it is receiving raw icon paths as the icon name and
hex values for colours. This is the most important part of this change I
want to land.

index.js
jonathanKingston: See comment below... for fixing storage

webextension/background.js
jonathanKingston: Done

jonathanKingston:
@groovecoder I think this is finally ready :)

PR #506, opened 2017-05-11
Assignment onboarding panel and icon badge

groovecoder:
This adds a new onboarding-panel-4 for site assignments. ( @ChenMorpheus may still update the text )

It also creates a way for us to add a "NEW" badge on the browser icon for major versions that introduce new or improved features. (In this case, I put `"2.2.1"` in the `MAJOR_VERSION` array, assuming we will release this as 2.2.1) It should draw more users' attention to the icon to get them to click and see the new onboarding panel.

image


webextension/js/popup.js
```diff
@@ -63,14 +64,20 @@ const Logic = {
...
     this.refreshIdentities()
 
     // Routing to the correct panel.
     .then(() => {
       // If localStorage is disabled, we don't show the onboarding.
-      if (!localStorage || localStorage.getItem("onboarded3")) {
+      if (!localStorage || localStorage.getItem("onboarded4")) {
         this.showPanel(P_CONTAINERS_LIST);
+      } else if (localStorage.getItem("onboarded3")) {
+        this.showPanel(P_ONBOARDING_4);
```
jonathanKingston: Are these in the right order? (I didn't check a fresh profile, the order looks odd though)

index.js
```diff
@@ -256,6 +258,9 @@ const ContainerService = {
 
     webExtension.startup().then(api => {
       api.browser.runtime.onMessage.addListener((message, sender, sendReply) => {
+        if (message.method === "checkForMajorUpgrade") {
+          sendReply(reason === "upgrade" && MAJOR_VERSIONS.indexOf(self.version) > -1);
```
jonathanKingston: The following will work in the background code:
```
const MAJOR_VERSIONS = ["2.2.0"];

async function checkExtensionInfo() {
  const manifestPath = browser.extension.getURL("manifest.json");
  const response = await fetch(manifestPath);
  const extensionInfo = response.json();
  if (MAJOR_VERSIONS.indexOf(extensionInfo.version) > -1)) {
    browser.browserAction.setBadgeBackgroundColor({color: "rgba(0,217,0,255)"});
    browser.browserAction.setBadgeText({text: "NEW"});
  }
}
checkExtensionInfo()
```

This would need some idea of updating though... perhaps a similar local storage to clear when the user has opened the browser action.

webextension/js/popup.js
```diff
@@ -63,14 +64,20 @@ const Logic = {
   _panels: {},
 
   init() {
+    // Remove browserAction "upgraded" badge when opening panel
+    browser.browserAction.setBadgeBackgroundColor({color: ""});
```
jonathanKingston: Can this be in it's own method like `clearVersionUpdate` and use state to store?

This would solve the other comment of only showing on update etc.

webextension/js/popup.js
groovecoder: Yup. I double-checked by doing `localStorage.clear()` in the browser action console. It's a bit odd because we check the user's onboarding state in reverse order.

index.js
groovecoder: With an embedded WebExtension, the `browser` object available to the legacy add-on code during startup only has the `runtime` property to set up communication between the legacy code and the WebExtension code. (https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Embedded_WebExtensions#Starting_the_WebExtension) And, only the WebExtension code can initiate the communication. 😢 Which is why I had to create another message listener to send a reply to the WebExtension. Only the WebExtension code can mess with the `browserAction` or `extension` APIs.

Or did you mean to move this entire function somewhere else outside of the `startup()` handler?

groovecoder:
Added a `Logic.clearBrowserActionBadge` method, but I don't think we can use `browser.extension` or `browser.browserAction` directly from the `index.js` code in an Embedded WebExtension.

jonathanKingston:
@groovecoder in the primitive testing I did, it looked like the pasted code worked in the background page sorry if that wasn't clear.

index.js
jonathanKingston: I meant move this code into WebExtension land. As mentioned the `runtime.onInstalled` may not fire but I don't think we need that to know if the user has seen it. We can store `latestSeenNewIndicator` or all of them etc. When the user pops open the doorhanger we can dismiss the badge indicator and store that they saw it.

groovecoder:
Okay, I updated this to move all the `browserAction` badge logic into the WebExtension code.

webextension/js/popup.js
```diff
@@ -85,6 +98,17 @@ const Logic = {
...
 
+  clearBrowserActionBadge() {
+    getExtensionInfo().then(extensionInfo=>{
+      browser.storage.local.get({browserActionBadgesClicked: []}).then(storage=>{
+        browser.browserAction.setBadgeBackgroundColor({color: ""});
+        browser.browserAction.setBadgeText({text: ""});
+        storage.browserActionBadgesClicked.push(extensionInfo.version);
+        browser.storage.local.set({browserActionBadgesClicked: storage.browserActionBadgesClicked});
+      }).catch(e=>{throw e;});
+    }).catch(e=>{throw e;});
```
jonathanKingston: nit:
```
async clearBrowserActionBadge() {
  try {
    const extensionInfo = await getExtensionInfo();
    const storage = await = browser.storage.local.get({browserActionBadgesClicked: []});
    browser.browserAction.setBadgeBackgroundColor({color: ""});
    browser.browserAction.setBadgeText({text: ""});
    storage.browserActionBadgesClicked.push(extensionInfo.version);
    browser.storage.local.set({browserActionBadgesClicked: storage.browserActionBadgesClicked});
  } catch (e) {
    // Do we even need to catch here if we are not handling it. I think async functions just throw as you would expect
  }
}
```

We can fix in a follow up though :)

groovecoder: Huh, I though I tried that and it failed to parse the `async`, but it works now. Cleaned up.

PR #509, opened 2017-05-16
Adding in shortcut and keyboard controls

jonathanKingston:
I think the command shortcut only works in nightly currently. It also doesn't seem to make the document active either which means it requires two tabs to get into the doorhanger which isn't the best. I tried focusing an element when the page became visible but nothing seems to work in that department.

### DIFF
--- a/webextension/css/popup.css
+++ b/webextension/css/popup.css
@@ -130,11 +130,13 @@ table {
   color: white;
 }
 
-.button.primary:hover {
+.button.primary:hover,
+.button.primary:focus {
   background-color: #0675d3;
 }
 
-.button.secondary:hover {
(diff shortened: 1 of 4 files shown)


jonathanKingston:
See the following issues related to this which I think make this mostly unusable currently:

- https://bugzilla.mozilla.org/show_bug.cgi?id=1365220
- https://bugzilla.mozilla.org/show_bug.cgi?id=1365222

Assuming these get fixed, I think this is actually nicer than popping the + button as we can later fix other bits (My rudimentary testing let me create a container etc however it wasn't the primary work for this patch)

groovecoder:
Needs conflict resolution, then it should be good-to-go.

PR #515, opened 2017-05-17
Remove underline from buttons. Fixes: #514

jonathanKingston:


jonathanKingston:
Forgot to pester you for this @groovecoder is this ok to go in?

PR #521, opened 2017-05-19
Adding container assignment exemption on confirm prompt. Fixes #500

### LINKED ISSUE #500: Assignment confirmation page changes
See https://docs.google.com/document/d/1zrk_M8Adqc05jMlgPFc0fyUdEysApCY3SO1QPpgT3gM/edit ...

The UI redesign tries to allow users open the site in the current container. (https://github.com/mozilla/testpilot-containers/issues/422)

image

jonathanKingston:


jonathanKingston:
@groovecoder I can add an error handler for the icon too like I do for Sea containers, even the extension tab apis return URLs that are not going to work. Showing a broken image will be nasty and somewhat common.

Would like a quick review of the assignment exemption logic.

The other thing I need to highlight is the telemetry will be wrong, what is the right path for this?I added: `click-to-reload-page-in-same-container` is that ok? Should we change the keys totally for both options... etc.

jonathanKingston:
@groovecoder I'm rebasing another patch on top of this as it will need new messages to the background.js for knowing if we are assigned or not so likely we can close this out as I might end up with a big commit. I just wanted to highlight the comment above^.

webextension/background.js
```diff
@@ -70,6 +70,18 @@ const assignManager = {
...
   },
 
+  // We return here so the confirm page can load the tab when exempted
+  async _exemptTab(m) {
+    const pageUrl = m.pageUrl;
+    // If we have existing data and for some reason it hasn't been deleted etc lets update it
+    const siteSettings = await this.storageArea.get(pageUrl);
+    if (siteSettings) {
+      siteSettings.exempted.push(m.tabId);
+      await this.storageArea.set(pageUrl, siteSettings);
```
groovecoder: This permanently stores the exemption for the tab ID, which means after a Firefox restart, a tab with the same ID going to the site will automatically skip the confirm page and open in the current container.

webextension/js/confirm-page.js
```diff
@@ -1,31 +1,85 @@
...
   document.location.replace(redirectUrl);
 }
+
+load();
+
+async function openInContainer(redirectUrl, cookieStoreId) {
+  const tabs = await browser.tabs.query({active: true});
+  browser.runtime.sendMessage({
+    method: "sendTelemetryPayload",
+    event: "click-to-reload-page-in-same-container",
```
groovecoder: Nice!

groovecoder: Need to add this to the `metrics.md` doc.

webextension/js/confirm-page.js
```diff
@@ -1,31 +1,85 @@
...
+    }
+  });
+}
+
+function createFavicon(pageUrl, redirectUrlElement) {
+  const origin = new URL(pageUrl).origin;
+  const imageElement = document.createElement("img");
+  imageElement.src = `${origin}/favicon.ico`;
+
+  redirectUrlElement.prepend(imageElement);
```
jonathanKingston: Not really... however we could just say "Test pilot is an experiment... yada" which we have for other issues like this and fix when stable. I already got involved in discussion about getting a way to query the favicon store in extensions instead.
We could load it in a throw away container that is never visible etc but I'm not sure it's actually solving the privacy issue. (I think I mentioned it in the issue were @ChenMorpheus wanted the title too)

webextension/background.js
jonathanKingston: As discussed in the meeting, I knew about this forgot to mention I would file a follow up for it etc. Thanks for the review!

jonathanKingston:
Superseded by #535

PR #530, opened 2017-05-24
Removal of tab state context menu as unable to detect tab assignment …

### LINKED ISSUE #520: "Always Open in This Container" not always displaying the correct state
When assigning a particular website to a container using "Always Open in This Container", right clicking on other container tabs that are currently opened will display the incorrect state for "Always Open in This Container".

STR: (example: [https://youtu.be/8RBx9-xta4w](https://youtu.be/8RBx9-xta4w))

* install the latest [unbranded version of fx53.0.2](https://wiki.mozilla.org/Add-ons/Extension_Signing#Unbranded_Builds)
* install Containers 2.3.0 via the `bump-version-2.3.0` branch
* open a personal container and load any website (used reddit.com)
* open a work container and load a website (used linkedin.com)
* open a banking container and load a website (used paypal.com)
* once you have all the tabs opened, switch back into the personal container
* right click on the page and select "Always Open in This Container"

Once you've assigned that website to the personal container, right clicking on either the work or banking container will show `Always Open in This Container` as enabled even though those websites are not associated with a particular container. It seems like the states in the context menu are incorrectly being displayed.

I created a video that I've added above that shows how `Always Open in This Container` jumps around into different states that don't necessarily match the actual state of each tab.

jonathanKingston:
…state. Fixes #520

### DIFF
--- a/webextension/background.js
+++ b/webextension/background.js
@@ -201,7 +201,7 @@ const assignManager = {
           id: menuId,
           title: `${prefix} Always Open in This Container`,
           checked: true,
-          contexts: ["all", "tab"],
+          contexts: ["all"],
         });
       }).catch((e) => {
         throw e;


jonathanKingston:
As mentioned in: https://bugzilla.mozilla.org/show_bug.cgi?id=1215376

There isn't the ability to detect the context menu before it is clicked. Due to our menu being stateful based on the tab we can't show it for tab item.

PR #540, opened 2017-05-26
Cleaning up layout issues for current tab panel.

jonathanKingston:


jonathanKingston:
@johngruen you don't need to approve this or anything however. This is for the shield study and I remember you saying you loved CSS and wanted to polish this further. (I have the current assignments management list to do still but everything else is done functionality wise)

PR #542, opened 2017-05-26
Assign menu for wrong tab

jonathanKingston:


jonathanKingston:
Whoops pr for other is in this, sorry.

webextension/css/popup.css
```diff
@@ -133,6 +133,11 @@ table {
   --identity-icon: url("/img/usercontext.svg#chill");
 }
 
+#current-tab [data-identity-icon="default-tab"] {
+  fill: currentColor;
+  background: center center no-repeat url("/img/blank-tab.svg");
```
groovecoder:
```
webextension/css/popup.css
 138:3  ✖  Expected       order/declaration-block-properties-alphabetical-order
           background to                                                       
           come before                                                         
           fill
```

groovecoder:
Also, I had to restore the `contexts: ["all", "tab"],` in `background.js` to get the context menu to show back up in the tab bar.

webextension/css/popup.css
jonathanKingston: This issue was with the other commit, Pushed there sorry

jonathanKingston:
@groovecoder this isn't the issue. The bug is fixing that when you change the tab you can get the wrong state... in content clicks whilst it is still loading.

jonathanKingston:
Woah how did you get multiple items here in always? Is this normal on mac? @groovecoder

groovecoder:
First time I've ever seen multiple items in that menu.

jonathanKingston:
My mac isn't working to test this, could you run off some console output for me?
I suspect this is removing the addon a few times maybe? I tried replicating for a long time and I am super not convinced this patch caused the double menu.

groovecoder:
I still don't see the original bug fixed on this branch? ... i.e., the "Always open in this Container" context menu option *is checked* until I change tabs or change windows and go back.

>     Open a new container tab (i.e. personal).
>     Go to first page (i.e. twitter.com) and mark it to always open in this container.
>     Go to a second page (i.e. tweetdeck.twitter.com).
>     Right-click to open the context menu and see, the always open in this container menu item is "checked", despite tweetdeck won't open in this container by default, only twitter.
> 

containers-always-open-bug

jonathanKingston:
Hmm this looks like it could be worse on mac perhaps? Again I still don't have a working mac that is easy to test on. Will look into further, it certainly stopped the quick tab switching str to me, perhaps not the typing a new URL I don't think I checked that.

groovecoder:
Are you testing with twitter.com vs. tweetdeck.twitter.com?

jonathanKingston:
@groovecoder no I'm checking for moving quickly between any assigned site and not assigned whilst it is still loading. There are other cases it catches. The twitter/tweetdeck one seems the same though.

I was able to replicate the multiple items issue... again unrelated. It appears reloading the extension in the debugger doesn't delete all context items.

jonathanKingston:
As agreed with @kjozwiak this is easier to manage as one commit as they are all related. #576 is all shield work merged into one

Issue #546, opened 2017-05-28
Change keyboard shortcut to Ctrl+Comma


jonathanKingston:
Due to a bug which impacts shortcuts when the location bar is focused https://bugzilla.mozilla.org/show_bug.cgi?id=1325697 Ctrl+Y is only usable when not focusing the URL bar and other text fields.

I'm going to propose we use Ctrl+Comma as it appears to be mostly unused and also Sea Containers use Ctrl+Period which makes it a nice relationship there.

PR #547, opened 2017-05-28
Exchanging Ctrl+Y for Ctrl+Comma due to text field collisions. Fixes …

jonathanKingston:
…#546

### DIFF
--- a/webextension/manifest.json
+++ b/webextension/manifest.json
@@ -36,7 +36,7 @@
   "commands": {
     "_execute_browser_action": {
       "suggested_key": {
-        "default": "Ctrl+Y"
+        "default": "Ctrl+Period"
       },
       "description": "Open containers panel"
     }


jonathanKingston:
Got any better suggestions? We could make it Ctrl+Comma on mac too if you like?

groovecoder:
On this branch with `Ctrl+Comma` specified, the keyboard shortcut didn't work. Do we have to specify different shortcuts for the different platforms? Cmd+Y makes sense to me because it's right next to Cmd+T for new-tab.

Maybe Cmd+. ?

groovecoder:
Cmd+Shift+O is right next to Cmd+Shift+P for Private Browsing window ...

jonathanKingston:
Ctrl+Shift+O is open library.
Ctrl/Cmd+Y is a text operation which is why we suffer these issues.

We can set specific ones for different platforms yeah, by default Ctrl is Cmd on mac, but we can specify MacCtrl on macs.
Cmd/Ctrl+. is what I used for Sea Containers which was why I picked comma. I'm happy to yield that and not really care about mac or use Ctrl+Shift+. perhaps? I would prefer to pick something else though.

They also should override native platform ones too... I had no idea that it was preferences.

jonathanKingston:
@groovecoder you can see the current permitted shortcuts here:
http://searchfox.org/mozilla-central/source/browser/components/extensions/schemas/commands.json#14

They **Should** override defaults however appear not to always.

jonathanKingston:
@groovecoder any more thoughts here?

groovecoder:
Afraid not ... maybe Cmd+Shift+C for 'Containers' ?

jonathanKingston:
@groovecoder nah that opens console for me.
Perhaps Ctrl+. and change Sea Containers to Ctrl+, / Ctrl+Shift+.

That sound ok?

groovecoder:
Yeah `Ctrl` + `.` works for me.

groovecoder:
Though `Ctrl` + `.` is already "Stop" ... but so is `Esc` ...

jonathanKingston:
Stop for what?

groovecoder:
Stop loading the page.

jonathanKingston:
@groovecoder do you think that is an issue? I have never used it, I thought it was escape. Perhaps it is mac specific?

I think the only decent options are:

- Mac: Cmd+.
- Linux/Win Ctrl+.

Or

- Linux/Win/Mac Ctrl+.

Sea containers can use Shift for the above.

groovecoder:
Let's do `Cmd`+`.` on Mac, `Ctrl`+`.` on Linux/Win.

jonathanKingston:
I have merged this into Sea containers to be Shift for the same. However it hasn't released yet (assuming we will have a few users overlap here). I think this is safe to merge. I will try and get a release of Sea Containers out this week with more features as releasing just a regression of a keyboard shortcut is likely not cool.

jonathanKingston:
As agreed with @kjozwiak this is easier to manage as one commit as they are all related. #576 is all shield work merged into one

Issue #398, opened 2017-03-20
Customized containers lose their color and icon when extension disabled


nkestrel:
STR:
1. Enable Containers Experiment.
2. Edit container colors and icons to something different.
3. Create container tabs (must be done after editing containers).
4. Disable Containers Experiment.
5. Note color of previously created container tabs.
6. Go to `about:preferences#containers`

![Containers in about:preferences](https://i.imgur.com/T47nsH5.png)

This can be very confusing because it makes container tabs look like non-container tabs, although #userContext-label is still visible in the location bar. Re-customizing the containers brings back their color and icon.

groovecoder:
To be clear - these are custom colors & icons you assigned to the built-in containers? (I.e., Personal, Work, Banking, Shopping)

kjozwiak:
> To be clear - these are custom colors & icons you assigned to the built-in containers? (I.e., Personal, Work, Banking, Shopping)

@groovecoder correct, this happens when you either:

* change the colour/icon of a built-in container via the Test Pilot Containers icon
* create a new custom container via the Test Pilot Containers icon

I actually noticed this happening a few weeks ago but didn't bother creating a bug due to the fact that in #114, we decided to disable access to `about:perferences#containers` by changing `privacy.userContext.ui.enabled;false` which removes the "Containers Tab" section under `about:preferences#privacy`. However, users can still access that portion of the preferences via manually typing `about:perferences#containers` into the awesome bar.

As @nkestrel mentioned above (I didn't think about this case earlier), this is going to negatively affect users who install the test pilot, change the icons and colours of the built-in containers, and than revert back to the platform implementation of the feature. Once the user visits `about:perferences#containers`, they'll run into the same problem. 

@groovecoder, we should probably fix this in the next sprint. Thoughts?

I created [bug#1345643](https://bugzilla.mozilla.org/show_bug.cgi?id=1345643) weeks back which is similar to this issue and will hopefully address this on the platform side of things. If someone as containers disabled, we shouldn't allow access to `about:perferences#containers`.

nkestrel:
This doesn't just affect `about:preferences#containers`, it affects all references to container color and icon, including the appearance of container tabs themselves.

![Customized container](https://i.imgur.com/q2NMjsd.png)

kjozwiak:
Thanks @nkestrel, looks like this bug actually affects all container access points as well. We should fix this in the next sprint as this is going to affect users who edit the built-in containers and eventually return to using the platform implementation.

[...]

PR #556, opened 2017-05-30
Reset color and icon when disabled - issue #398

bakulf:


groovecoder:
From the Travis job:
```
/home/travis/build/mozilla/testpilot-containers/index.js

  1487:58  error  Expected '!==' and instead saw '!='  eqeqeq

  1488:60  error  Expected '!==' and instead saw '!='  eqeqeq

  1493:76  error  Missing semicolon                    semi

  1499:78  error  Missing semicolon                    semi
```

jonathanKingston:
The question to me is why are we storing preInstalledIdentities then breaking them? Is something not working here? I'm not sure we need more code to make this work do we?

bakulf:
preInstalledIdentities contains just IDs. We cannot change it for compatibility reasons.

index.js
```diff
@@ -1587,6 +1597,32 @@ ContainerWindow.prototype = {
 
     return true;
   },
+
+  _shutdownContainers() {
+    ContextualIdentityProxy.getIdentities().forEach(identity => {
```
jonathanKingston: Can we add a comment explaining this whole function.

`// Changes icons back to defaults when they are not in native firefox`

or something similar

index.js
```diff
@@ -1587,6 +1597,32 @@ ContainerWindow.prototype = {
...
+      if (IDENTITY_ICONS_STANDARD.indexOf(identity.icon) !== -1 &&
+          IDENTITY_COLORS_STANDARD.indexOf(identity.color) !== -1) {
+        return;
+      }
+
+      if (identity.userContextId <= IDENTITY_ICONS_STANDARD.length) {
+        identity.icon = IDENTITY_ICONS_STANDARD[identity.userContextId - 1];
+      } else {
+        identity.icon = IDENTITY_ICONS_STANDARD[0];
```
jonathanKingston: We could actually check if the current icon doesn't exist in `IDENTITY_ICONS_STANDARD` which would keep icons how the user modified them.

Same for colors too.

jonathanKingston:
> preInstalledIdentities contains just IDs. We cannot change it for compatibility reasons.

Bum! It's something we will have to remove in the SDK removal, is there any way it can be put in `background.js` instead? Even if we have to message that?

jonathanKingston:
Rebased on the command line and merged in: https://github.com/mozilla/testpilot-containers/commit/3b9da05e673a531af964be4f332769d64ba0a134

PR #557, opened 2017-05-30
only start study for @shield-study-privacy

groovecoder:


jonathanKingston:
Maybe make a const for that ID but just a nit.

Issue #518, opened 2017-05-18
Consider support for "link click"+keyboard shortcut within content


jonathanKingston:
To improve the use of containers should we consider a shortcut like Alt+click for opening a container in "Default"?

We could also make Alt+click and drag down/long press work like the back/forward buttons in Firefox to open a container list.

/cc @groovecoder for discussion

jonathanKingston:
Alt+Ctrl/Alt+Shift might be needed at least on Ubuntu as they use Alt+Click for window dragging.
I can't even seem to override this within Firefox itself.

jonathanKingston:
The ugly looking PR I added is a proof of concept that looks a little like this:

![selection_672](https://cloud.githubusercontent.com/assets/338988/26728204/58f9847a-47a1-11e7-909e-3696c50d40f7.png)

jonathanKingston:
This could also be expanded to using focused links for Alt+tab too. I select the first item by default so keyboard users can tab here.

bakulf:
This is nice. But it's similar to right click -> open link in container X.

jonathanKingston:
@bakulf yeah its very similar. I find that menu too hard to use though. Allowing it for keyboard navigation would be a significant advantage too.
Users are familiar with Ctrl+Clicking a lot right click, move down two, across requires a lot more overhead for a common action.

jonathanKingston:
![selection_677](https://user-images.githubusercontent.com/338988/26857785-1b019d06-4b25-11e7-96ae-c9b2df5a6715.png)
![selection_678](https://user-images.githubusercontent.com/338988/26857787-1e433c0e-4b25-11e7-8b49-7458f8a84a73.png)

Added drop shadow and made it much like #433. If you get time @groovecoder this is now a working demo would be cool to get your thoughts. It's still not super bug free etc however I think it demos how usable this can be over the right click context menu.

jonathanKingston:
Outstanding issues:

- [ ] Some events don't seem to be cancelled, on mozilla.org the side menu still opens
- [ ] Some urls break and instead open a blank tab
- [ ] Keyboard navigation doesn't seem to work
  - [ ] Tabbing out of the menu what should happen
  - [ ] Arrow keys etc
- [ ] Probably needs a quick subtle animation
- [ ] Prevent text highlight
- [ ] Font sometimes is overridden

dannycolin:
That option would be great but It would be nice to consider that some people use the caret mode (F7). Does the list would be accessible to them.

PR #568, opened 2017-06-02
In content tab

jonathanKingston:
- [x] Messaging to get active container list
- [x] Correct icons
- [x] Messaging to create the container
- [x] Layout to copy https://github.com/mozilla/testpilot-containers/issues/433#issuecomment-305455327

This is a supserset of other UI fixes, I only care about your thoughts on this commit:

https://github.com/jonathanKingston/testpilot-containers/commit/8d763a645a7f07f3bcedf0f5734290f921503022

Essentially this gives an ugly in content context menu for Shift+Alt+Click a link to select in a container. This is less distance to travel than the sub context menu on right click and I think this will be much easier to use. (Not sure if Alt+Click is possible on mac, it's used by Ubuntu I might be able to drop the Shift for Mac if that is the case).

This is for: #518

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,6 +9,7 @@ module.exports = {
     "webextensions": true
   },
   "globals": {
+    "Utils": true,
     "CustomizableUI": true,
     "CustomizableWidgets": true,
     "SessionStore": true,
--- a/webextension/background.js
+++ b/webextension/background.js
@@ -408,6 +408,15 @@ const messageHandler = {
       let response;
 
       switch (m.method) {
+      case "getContainers":
+        response = browser.contextualIdentities.query({});
+        break;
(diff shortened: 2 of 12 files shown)


groovecoder:
Do we want to keep this PR and branch alive? Or is it Overcome By Events?

jonathanKingston:
I was just going to make a follow up addon instead it's an in content shortcut for faster container opening however doesn't belong in us. Will fix up this week (I would like to keep it open as a reminder).

groovecoder:
Weekly check-in - do we still need this PR open? Can we convert it into an issue?

jonathanKingston:
This has moved to: https://github.com/jonathanKingston/alt-context-containers-menu the code can live on there, I neatened up the code etc ready for publishing.

Issue #511, opened 2017-05-16
File shield study bugs


groovecoder:
Follow the process here:

https://mozilla.github.io/shield-studies-docs/study-process/

jonathanKingston:
@johngruen I'm working on #501 which is nearly complete, every other part is in a PR so might have follow ups but we are about ready for this. There are a fair few bugs for shield we are blocked on you for. Do you know what is happening with those?

Thanks

johngruen:
Meta Bug Here: https://bugzilla.mozilla.org/show_bug.cgi?id=1371384

Other tree bugs to be filed pending confirmation of documentation in today's check in

johngruen:
@groovecoder can you confirm that the metrics.md doc is up to date for the shield study?

groovecoder:
The `metrics.md` file is accurate with 1 exception:

> The user clicks "Open in Current container" to reload a site into a container after the user picked "Always Open in this Container".

This is new functionality in the shield study (or next version of Test Pilot add-on - however you want to look at it). So we need to add this probe. I'll do that now.

groovecoder:
Filed https://github.com/mozilla/testpilot-containers/pull/577

groovecoder:
https://github.com/mozilla/testpilot-containers/pull/582 merged, so I'm removing myself from this task now. `metrics.md` is up-to-date for the pings sent by both the test pilot add-on and the shield study add-on.

groovecoder:
Closing issue, please follow https://bugzilla.mozilla.org/show_bug.cgi?id=1371384

PR #577, opened 2017-06-13
for #511: add "same-container" telemetry ping

groovecoder:
May not need to merge ... ?

When I stepped thru the "Open in Current Container" flow, I didn't see a `click-to-reload-page-in-same-container` telemetry ping sent.

So, I added it to the `denySubmit` function.

However, there's code to send the ping in `openInContainer`, but that never seems to run?

I'm confused. 😢

### DIFF
--- a/webextension/js/confirm-page.js
+++ b/webextension/js/confirm-page.js
@@ -63,6 +63,10 @@ async function denySubmit(redirectUrl) {
     tabId: tab[0].id,
     pageUrl: redirectUrl
   });
+  browser.runtime.sendMessage({
+    method: "sendTelemetryPayload",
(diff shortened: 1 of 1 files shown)


jonathanKingston:
Ah good catch.
Sorry I think it got moved to the wrong function (may have been in the merges etc, I remember you reviewed it and it not being like this... dunno). Anyway will move this code and do a different PR.

I am still pretty worried about location.replace / closing a tab straight after sending telemetry like this. I think we do this in a few places. Perhaps we should wait for the background script to at least acknowledge the message and wait for that before redirecting or closing? (Will raise a separate bug to discuss)

PR #581, opened 2017-06-14
Fixing keyboard focus issues to new layout

jonathanKingston:


groovecoder:
What was the original bug here?

jonathanKingston:
@groovecoder wasn't one, it was a commit I found that I meant to add to the shield study changes. When I moved over the styles to look like the original mockups with the hover + icon in the browser action looking different. I just noticed the focus didn't match.

Issue #498, opened 2017-05-10
Create on-boarding variations


groovecoder:
From [the PHD](https://docs.google.com/document/d/1vMD-fH_5hGDDqNvpRZk12_RhCN2WAe4_yaBamaNdtik/edit)

* Existing on-boarding in Test Pilot
* [Privacy-emphasized on-boarding](https://drive.google.com/open?id=0B_NalwdgcXpqTTlnWXZmSXJzVUU)
* Control

@ChenMorpheus - how is the existing on-boarding different from the Control?

johngruen:
I'll chase down @ChenMorpheus for details

johngruen:
@ChenMorpheus has these ready; getting string review now

ChenMorpheus:
I just copy and paste my comment for Luke's question from the PHD to here for record.

Control group here is to collect data "without" Containers which are then compared to determine the changes that may occur due to the experimental treatment.

johngruen:
Strings being reviewed here: https://docs.google.com/document/d/1PgTjCvagjy3un64GfbIShvJBgi3FCe-YBd0XMdnm-Dc/edit?userstoinvite=mheubusch@mozilla.com&ts=593acfb8&actionButton=1

hobarrera:
Looks like the linked docs are private I need permissions to see it.

johngruen:
@hobarrera sorry about that; should be open now

hobarrera:
Yup thanks (though only the first link is public still).

ChenMorpheus:
@groovecoder 
Here is the asset for the last onboarding. Sorry that I thought we had provided it already.

https://www.dropbox.com/sh/bfyks23ncp46yuu/AAC4HCMRYwvqA1IA1xFHwq-7a?dl=0

jonathanKingston:
Hey @groovecoder could you make the command the following:

```
    "package-shield": "./node_modules/json/lib/json.js -I -f package.json -e 'this.name=\"shield-study-containers\"' && jpm xpi && ./node_modules/json/lib/json.js -I -f package.json -e 'this.name=\"testpilot-containers\"'",
```

That way the dependency comes from the installed version rather than global. I didn't have the global version so it silently failed etc.

Thanks

PR #591, opened 2017-06-14
for #498: security onboarding panels and logic

groovecoder:
Not quite ready for merge yet. I need the image for the 3rd security onboarding panel.

To test:
* `shield run . control -- --binary Nightly` should show the existing onboarding panels
* `shield run . securityOnboarding -- --binary Nightly` should show the new security onboarding panels


groovecoder:
Adding @johngruen b/c the copy needs some love too.

webextension/js/popup.js
```diff
@@ -69,15 +69,19 @@ const Logic = {
...
     this.clearBrowserActionBadge();
 
     // Retrieve the list of identities.
-    this.refreshIdentities()
+    const identitiesPromise = this.refreshIdentities();
+    // Get the onboarding variation
+    const variationPromise = this.getShieldStudyVariation();
 
     // Routing to the correct panel.
+    Promise.all([identitiesPromise, variationPromise])
```
jonathanKingston: `await Promise.all()` maybe?

webextension/js/popup.js
```diff
@@ -257,6 +270,14 @@ const Logic = {
     });
   },
 
+  getShieldStudyVariation() {
+    return browser.runtime.sendMessage({
```
jonathanKingston: await?

webextension/js/popup.js
```diff
@@ -159,6 +163,15 @@ const Logic = {
     }).catch((e) => {throw e;});
   },
 
+  getPanelSelector(panel) {
```
jonathanKingston: This is fine, I perhaps would have used a translation based approach where the strings are embedded in some JSON and some method looks up the variation translation, this would potentially be a little less coupled to having this method.

groovecoder:
I cleaned up the first instance of `.then()` with `await Promise.all()`, but I stopped short of going into the rabbit-hole of `async`+`await` thru all of `refreshIdentities` and `getShieldStudyVariation`.

I think I started on a JSON/translation approach, but it was easier and simpler to check the variation in the single `showPanel` call?

jonathanKingston:
@groovecoder cool I'm happy for all of this just wanted to note that it won't really scale as a solution if we decided to spin of 5 variations etc.

jonathanKingston:
I'm getting `getShieldStudyVariation` returning undefined meaning both variations for me are the same.

Ah... `json` whatever command that is doesn't exist on Ubuntu.

jonathanKingston:
`jq .name=\"shield-study-containers\" package.json` does the same for `jq` which I would expect is more universal.

I could only get it working with this though (seems jq doesn't like writing to the same file):
```
jq .name=\"shield-study-containers\" package.json > shield.json && mv shield.json package.json
```

jonathanKingston:
Hey @groovecoder could you make the command the following:

```
    "package-shield": "./node_modules/json/lib/json.js -I -f package.json -e 'this.name=\"shield-study-containers\"' && jpm xpi && ./node_modules/json/lib/json.js -I -f package.json -e 'this.name=\"testpilot-containers\"'",
```

That way the dependency comes from the installed version rather than global. I didn't have the global version so it silently failed etc.

Also how would you generate the variations via `npm run`?

Thanks

groovecoder:
Added the 3rd onboarding panel image for the securityOnboarding variation.

Also, updated the command to use the local `json` binary.

@jonathanKingston - we don't generate different `.xpi` files to match the different variations. The shield library code sees the variations defined in the single `study.js` module and selects a variation for the user for the rest of the study's duration.

groovecoder:
Removing the DO NOT MERGE label, as I think we can merge this. But would still like to get the latest & greatest copy from @johngruen.

johngruen:
@groovecoder this has been formally reviewed: 

https://docs.google.com/document/d/1PgTjCvagjy3un64GfbIShvJBgi3FCe-YBd0XMdnm-Dc/edit?usp=sharing

groovecoder:
Updated to final copy. Ready for final r+ and merge @jonathanKingston

jonathanKingston:
@groovecoder I wasn't able to get this to work with test pilot for this branch in a brand new profile. It might just be me though. r+ and merge if the code works for you in this setup though and I can review again tonight.

groovecoder:
Ah, the security on-boarding panel only shows up when running the `@shield-study-containers` add-on.

jonathanKingston:
@groovecoder I mean I was expecting the normal onboarding. I didn't get this.

jonathanKingston:
Sorry I lied, it just doesn't even load the extension for me. No browser action or anything.

jonathanKingston:
I merged as I was able to reproduce in master which was more worrying,

jonathanKingston:
I was being a dork, new profiles don't allow unsigned xpi's. Sorry.

Issue #598, opened 2017-06-16
Tab closing became very slow in today's nightly


past:
I've been using containers for quite some time, but I noticed after today's update that my nightly was very slow to close a tab when containers was enabled. This happened on my main profile in both windows and macOS. 

Here is a profile: https://perfht.ml/2rzd7Pf

mconley and florian suggested that sending telemetry is what is causing this. sendTelemetryPayload from resource://testpilot-containers/index.js may be triggering [bug 1369594](https://bugzil.la/1369594).

Could be related to https://github.com/mozilla/testpilot/issues/2422.

mikeconley:
Filed by a separate user here: https://bugzilla.mozilla.org/show_bug.cgi?id=1373822

jonathanKingston:
Hey @groovecoder the `sendTelemetryPayload` method goes straight to sending the call right? I wonder if using a message queue style setup to send the events on shutdown and when the browser is idle?

Should shared workers be investigated for preventing telemetry blocking the main code?

kjozwiak:
Here's another [profile](http://bit.ly/2rFp6pH) that was added in [bug#1373822](https://bugzilla.mozilla.org/show_bug.cgi?id=1373822) which demonstrates the performance regression.

grahamperrin:
Please: with release 56.0 of Firefox, is this still an issue? 

If it helps: Mozilla bug [1373822 - Lag when closing tabs](https://bugzilla.mozilla.org/show_bug.cgi?id=1373822) is RESOLVED INVALID.

past:
Well that bug was closed as invalid, because this issue was the place where we would track the problem. That being said, I haven't experienced this issue again in the last few months.

groovecoder:
Closing unless/until we can reproduce.

PR #599, opened 2017-06-16
for #598: stop calling getCurrentPingData

groovecoder:
See https://github.com/mozilla/testpilot/commit/fa0f73fd6a0911f54366e3c8bef836995d4491fe

### DIFF
--- a/lib/testpilot/experiment.js
+++ b/lib/testpilot/experiment.js
@@ -6,6 +6,9 @@ const { storage } = require('sdk/simple-storage');
 const {
   TelemetryController
 } = require('resource://gre/modules/TelemetryController.jsm');
+const {
+  TelemetryEnvironment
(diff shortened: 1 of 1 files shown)


jonathanKingston:
This LGTM but not sure if there are other side effects? You mentioned you wanted to delay merging this, is there any reason?

groovecoder:
I just prefer to avoid any code-churn while we're in the final stretch of code + QA for the shield study.

jonathanKingston:
Adding DNM based on recent meeting.

jonathanKingston:
@groovecoder we should probably merge this, the current code removes all telemetry anyway. (file a follow up if you want it back for test pilot)

jonathanKingston:
Closing as ripping it out instead, sorries @groovecoder

PR #602, opened 2017-06-19
Edit restyle

jonathanKingston:


jonathanKingston:
@groovecoder can you test this on mac for the radio styles please (especially on release)? I left it as several raw commits so I can cut out the second commit if it doesn't work.

Firefox 54 made it easier to style how we are with `appearance:none;` but I wanted to double check as it didn't work before.

groovecoder:
The radio options look correct. The icons and colors look kinda small, but I think that's intended?

image

webextension/js/popup.js
```diff
@@ -750,10 +751,30 @@ Logic.registerPanel(P_CONTAINER_EDIT, {
     this.initializeRadioButtons();
 
     Logic.addEnterHandler(document.querySelector("#edit-container-panel-back-arrow"), () => {
-      this._submitForm();
+      const formValues = new FormData(this._editForm);
+      if (formValues.get("container-id") !== NEW_CONTAINER_ID) {
+        this._submitForm();
+      } else {
+        Logic.showPreviousPanel();
```
groovecoder: This doesn't seem to work? When I edit a container and press "Enter" it doesn't do anything? I expected it to do the same as the back arrow: show the previous panel.

webextension/js/popup.js
```diff
@@ -857,8 +881,10 @@ Logic.registerPanel(P_CONTAINER_EDIT, {
     const userContextId = Logic.currentUserContextId();
     const assignments = await Logic.getAssignmentObjectByContainer(userContextId);
     this.showAssignedContainers(assignments);
+    document.querySelector("#edit-container-panel .panel-footer").hidden = !!userContextId;
```
groovecoder: This seems to set `hidden=""` on the edit panel. Which, combined with `[hidden] { display: none !important }` seems to always prevent the footer from showing up ...

image

I think this needs to be `if (!!userContextId) { document.querySelector("#edit-container-panel .panel-footer").removeAttribute('hidden'); }` ... because when I remove the hidden attribute completely, the footer shows up.

PR #605, opened 2017-06-20
Shirk my SDK code

jonathanKingston:


groovecoder:
I verified this mostly works, and the most noticeable bugs are already listed in https://github.com/mozilla/testpilot-containers/issues/486. Needs a merge from master to resolve conflicts.

webextension/js/background/tabPageCounter.js
```diff
@@ -0,0 +1,64 @@
+const tabPageCounter = {
```
groovecoder: We can remove this ... either now, or in a separate branch to prepare for AMO.

jonathanKingston: Will do later I wasn't sure if we wanted to account for opt in telemetry at any point. Was a little tricky to reengineer this and other telemetry bits (this was pre decision to drop it totally that I did the work) So I would rather leave it in as totally non functioning for now. We could *consider* supporting telemetry for people who want to help us also.

jonathanKingston:
@kjozwiak looks like you were already assigned anyway

groovecoder:
Things still look good. But when I installed this onto my existing 2.4.0 add-on, I lost all my container tabs? 😢

jonathanKingston:
@groovecoder just the tabs or the containers too?

groovecoder:
I lose the tabs *and* the containers. 😢 

(Note: in the screen-vid, the first orange tab is in a "Test AMO Update" container, which is then destroyed during the update)

container-tab-loss

jonathanKingston:
@groovecoder I couldn't replicate this when renumbering this to 3.0.0 after rebasing.

My STR was making the two XPIs from master and this branch:
- about:debugging add old
- Create some tabs and containers
- about:debugging add this

jonathanKingston:
@groovecoder I was able to replicate #703 following the same STR from about:addons though

groovecoder:
Okay, I re-tried again on 54, 55, 56, and 57 Nightly and I wasn't able to reproduce losing tabs & containers. So I think this is ready as soon as the travis-ci/`eslint` errors are fixed.

groovecoder:
Merged https://github.com/mozilla/testpilot-containers/pull/707 so this needs rebase & `eslint` fixes then it's ready to go.

jonathanKingston:
@groovecoder should be good to go merged and dead headed the code. I tested my TODO comments I removed, works with less code 👍

PR #637, opened 2017-06-23
Cleanup exemption code. On assignment set/remove clear exemptions and…

### LINKED ISSUE #635: Refreshing a website assigned to a different container wrongly opens the "Containers confirm navigation" page in a new tab
**[Affected versions]:**
- Firefox 54 and up
- Containers Experiment Build: 2.4.0

**[Affected Platforms]:**
- All Windows
- All Linux
- All Mac

**[Prerequisites]:**
- Containers Experiment is installed on a new browser profile.

**[Steps to reproduce]:**
1. Start the browser with the profile from prerequisites.
2. Open the same website in 2 different containers.
3. Assign the website to always open in the first container.
4. Refresh the page in the second container.
5. Observe the results.

**[Expected results]:**
- Container confirm navigation tab, replaces the page for which the refresh was made.

**[Actual results]:**
- After the page refresh is initialized, the "Container confirm navigation" page is opened in a new tab.

**[Notes]:**
- Attached a screen recording of the issue:
refresh in the same tab

jonathanKingston:
… set based on existing tabs. Fixes #635

### DIFF
--- a/webextension/background.js
+++ b/webextension/background.js
@@ -15,19 +15,24 @@ const assignManager = {
     },
 
     setExempted(pageUrl, tabId) {
-      if (!(tabId in this.exemptedTabs)) {
-        this.exemptedTabs[tabId] = [];
-      }
       const siteStoreKey = this.getSiteStoreKey(pageUrl);
-      this.exemptedTabs[tabId].push(siteStoreKey);
+      if (!(siteStoreKey in this.exemptedTabs)) {
+        this.exemptedTabs[siteStoreKey] = [];
+      }
+      this.exemptedTabs[siteStoreKey].push(tabId);
+    },
+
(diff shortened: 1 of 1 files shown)


jonathanKingston:
These arrays don't persist, they get blown away on browser restart. At
least that is the current case. They are in storage area because they did
persist and we might want to change that later perhaps too.

groovecoder:
Hmm ... I'm trying to think thru the case of a user who gets the update - will they get a bunch of funky behavior until they restart Firefox? I.e., if they have assigned sites to containers in the current browser session, will the tab in which they assigned be screwed up? Is that a big enough problem that we care?

groovecoder:
I tested the user scenario and their pending exempted tabs work fine.

PR #638, opened 2017-06-23
Add a code of conduct and contributing file.

jonathanKingston:


jonathanKingston:
I suggested the #containers room as we aren't just testpilot now (we can use the legacy room for planning and boring stuff). The other can be more for community.

PR #643, opened 2017-06-29
Pre shield

jonathanKingston:


index.js
```diff
@@ -1105,22 +1093,50 @@ ContainerWindow.prototype = {
...
+
+    // Only hack the normal plus button as the alltabs is done elsewhere
+    this.attachMenuEvent("plus-button", this._plusButton);
+
     attachTo(this._style, this._window);
   },
 
+  attachMenuEvent(source, button) {
+    const popup = button.querySelector(".new-tab-popup");
+    popup.addEventListener("popupshown", (event) => {
```
groovecoder: `  1113:43  error  'event' is defined but never used  no-unused-vars`

jonathanKingston: Arse I fixed that.... just never pushed

groovecoder:
Looks good and the 2 new pings show up in `about:telemetry` when I use the old long-press + UI.

Do we want to merge this now? If we merge it into master, it will go into Test Pilot with the next Test Pilot update. Or should we make the shield `.xpi` off of this branch?

index.js
```diff
@@ -1105,22 +1093,50 @@ ContainerWindow.prototype = {
...
+
+    // Only hack the normal plus button as the alltabs is done elsewhere
+    this.attachMenuEvent("plus-button", this._plusButton);
+
     attachTo(this._style, this._window);
   },
 
+  attachMenuEvent(source, button) {
+    const popup = button.querySelector(".new-tab-popup");
+    popup.addEventListener("popupshown", () => {
```
groovecoder: Crap. On Fx 54, this triggers `TypeError: popup is null`

groovecoder: Also affects 55. Looks like 56+ is fine?

groovecoder: Filed https://github.com/mozilla/testpilot-containers/issues/658

index.js
```diff
@@ -1437,11 +1341,6 @@ ContainerWindow.prototype = {
     }
   },
 
-  _shutdownPlusButtonMenu() {
```
groovecoder: This caused `TypeError: this._shutdownPlusButtonMenu is not a function` during a 2.3.0 -> 2.4.0 upgrade on Fx 56.

Issue #603, opened 2017-06-20
Implement new light/dark icons


jonathanKingston:
As per https://bugzilla.mozilla.org/show_bug.cgi?id=1329242#c17 soon Nightly will have the ability to specify light and dark icons which once in stable will help remove our icon code.

PR #654, opened 2017-07-03
Use new theme_icons manifest for specifying light and dark icons. Fix…

jonathanKingston:
…es #603


jonathanKingston:
Looks like this solution might not see the light of day. I will be changing it to the content fill option instead in my migration branch (which already has this).

PR #664, opened 2017-07-11
add surveyUrl for shield study participants

groovecoder:
* [x] Find out if removing the add-on via `about:addons` should trigger the survey opening.
  * [x] Debug why it's not

### DIFF
--- a/study.js
+++ b/study.js
@@ -7,10 +7,15 @@ const { when: unload } = require("sdk/system/unload");
 
 const shield = require("./lib/shield/index");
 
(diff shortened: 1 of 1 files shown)


groovecoder:
I had the wrong `surveyUrls` keys, so this is now working. @jonathanKingston - can you verify:

* `npm run build-shield`
* Install the `shield-study-containers.xpi` into a fresh Firefox profile (Nightly with `xpinstall.signatures.required` set to `false`)
* Remove the add-on
  * [ ] It should open a post-study survey

PR #668, opened 2017-07-13
Adding a keyboard shortcut that will work on MacOS. Ctrl+.

jonathanKingston:


webextension/manifest.json
```diff
@@ -35,7 +35,8 @@
   "commands": {
     "_execute_browser_action": {
       "suggested_key": {
-        "default": "Ctrl+Period"
+        "default": "Ctrl+Period",
+        "mac": "MacCtrl+Period"
```
groovecoder: Sure that's right? `control`+`.` still isn't working for me. 😢

groovecoder: Works in Fx55+ which is released Aug 8. So, are we okay with broken Mac keyboard shortcut until then? (My Cmd+Y is broken 50% of the time anyway, because a text editor in content steals that shortcut)

PR #681, opened 2017-07-17
Set default background to fix dark linux themes. Fixes #677

### LINKED ISSUE #677: Unreadable on dark GTK theme
Could you at least make the background of this extension popup white? I can't read it as is. It's grey on grey. Or, if, possible, just use the default colors provided by the OS so I can see light text on dark.

jonathanKingston:


jonathanKingston:
We could potentially cut this with the tab index fix for assign for the Shield study @groovecoder and a new release for test pilot perhaps?

PR #707, opened 2017-08-03
Fix on update icons being reset to central defaults. Fixes #703

### LINKED ISSUE #703: Custom contain icons and colors reset.
On a few devices and two different Firefox profiles I noticed that all of my custom containers (e.g. not shopping/personal/work etc.) have changed they are now using the fingerprint icon and some of them have a different color than before.

jonathanKingston:
@groovecoder let me know when this is merged and I can rebase it into my other patch (I would like to land that this week, I still can't replicate on Nightly at leasr the issue you had).

### DIFF
--- a/index.js
+++ b/index.js
@@ -1028,6 +1028,8 @@ const ContainerService = {
       // Let's forget all the previous closed tabs.
       this._forgetIdentity();
 
+      this._resetContainerToCentralIcons();
+
       const preInstalledIdentities = data.preInstalledIdentities;
       ContextualIdentityProxy.getIdentities().forEach(identity => {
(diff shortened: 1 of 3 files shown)


jonathanKingston:
@groovecoder I used 2.4.1 -> 2.5.0 -> 3.0.0 to verify that it's then fixed in the 2.5.0 upgrade path.
I used about:addons after manually turning off XPI signatures in Nightly.

groovecoder:
Ah, yeah, I forgot it was in the `shutdown` code, so it will still affect 2.4.1 -> 2.5.0. When I fixed a 2.5.0 container icon & color and updated to 3.0.0 it stayed fixed.

PR #713, opened 2017-08-05
Svg icon context change

jonathanKingston:


jonathanKingston:
This branch is now error free on the a build of Beta that I made.
I don't think the error is related.

The only issue is on beta dark theme the colour isn't supported.

selection_783

I could add back in all the theme manager code if you think this is important.

PR #715, opened 2017-08-06
Removal of legacy CSS for builds after 2015-08-06. Fixes #714

### LINKED ISSUE #714: Removal of legacy CSS
Now that the tab highlight fix has [landed](https://bugzilla.mozilla.org/show_bug.cgi?id=1387117), we can safely remove the code for at least Nightly users as of today. (We should check if the uplift worked later too as we should be able to remove from 56 too: https://bugzilla.mozilla.org/show_bug.cgi?id=1325902)

See: https://github.com/mozilla/testpilot-containers/issues/708#issuecomment-320504642

jonathanKingston:


jonathanKingston:
I think we need to check versions here in case we release a patch to 55.

groovecoder:
Styles look good in both a Nightly (57) build and Release (54) build.

PR #716, opened 2017-08-09
Production

tkntjsdw:


jonathanKingston:
I'm not sure who this account belongs to? I also don't think it makes sense to do this, it's just missing commits from the merge itself.

PR #726, opened 2017-08-13
Bootstrapify

jonathanKingston:
This is a WIP to fix the SDK removal issue. It removes EVERYTHING including restoring prefs, telemetry and other bits we were starting to remove.
This will have messed up styles in 55 because only 56 got uplifted for the tab changes.


jonathanKingston:
Even with legacy extensions enabled and signing off I am getting the following error in my embedded extension.
```
There was an error during installation: Add-on testpilot-containers@mozilla.com is not compatible with application version. add-on minVersion: 51.0a1. add-on maxVersion: *.
```

jonathanKingston:
@groovecoder this now works in Nightly too. If I get time later I will make it restore prefs on removal.

Is there anything else you can think of?

I could also restore containers etc however that is a lot of work in comparison I think.

jonathanKingston:
I have the code reading from the old storage however I am not certain where I should be writing to again which is a concern if web extensions decide to delete all legacy storage.

I'm going to go for writing to the same location for now as I can't find any advice on where I should be writing to.

install.rdf
```diff
@@ -0,0 +1,22 @@
...
+    <em:hasEmbeddedWebExtension>true</em:hasEmbeddedWebExtension>
+    <em:name>Testpilot containers</em:name>
+    <em:targetApplication>
+      <Description>
+        <em:id>{ec8030f7-c20a-464f-9b0e-13a3a9e97384}</em:id> <!--Firefox-->
+        <em:minVersion>51.0a1</em:minVersion>
+        <em:maxVersion>*</em:maxVersion>
+      </Description>
+    </em:targetApplication>
+    <em:version>3.0.0</em:version>
```
groovecoder: Master is already at 3.0.1 - should we bump this to 3.0.2?

webextension/js/background/index.html
```diff
@@ -22,7 +21,6 @@
     <script type="text/javascript" src="identityState.js"></script>
     <script type="text/javascript" src="messageHandler.js"></script>
     <script type="text/javascript" src="tabPageCounter.js"></script>
-    <script type="text/javascript" src="themeManager.js"></script>
```
groovecoder: Should also `git rm webextension/js/background/themeManager.js`?

groovecoder:
Looking good. Needs merge and Travis CI kick.

But I'm unclear how we'll address the `extensions.legacy.enabled = true` requirement?

jonathanKingston:
@groovecoder kicked travis hopefully should be ok as everything passes locally.

> But I'm unclear how we'll address the extensions.legacy.enabled = true requirement?

This should work for signed addons, could you test this please? (I don't have access to the super secret signing keys)

groovecoder:
Okay, I [signed the add-on with the stage signing service](https://mana.mozilla.org/wiki/display/SVCOPS/Sign+a+Mozilla+Internal+Extension) and it works. (Note: I had to set `xpinstall.signatures.dev-root` to `true` so Firefox would accept dev signatures)

So, this branch needs conflict resolution, merge, and we can deploy it.

**Note**: After merge, I'll tag a 3.1.0 release and attach a `stage-signed.xpi` for additional testing of the signature exception.

PR #738, opened 2017-08-17
make page-title respect container boundaries

drc0:
the current tab page-title no longer overflows the popup width


drc0:
closing and reopening to trigger a travis rebuild because of the forced update commit

jonathanKingston:
Sorry for the delay here.
I can't see an impact here, is there a certain setup which is causing this? I'm happy to merge if it's a certain OS or whatever but it seems to work as expected for my setup.

jonathanKingston:
Also if you could rebase this on that would be helpful, if not I can do that :)

groovecoder:
1 month with no activity. Closing.

Issue #665, opened 2017-07-11
Migrate existing Test Pilot users to AMO


groovecoder:
Congratulations on a successful Test Pilot run! We'd like to get this in AMO so you can start managing and pushing your updates. Below are the steps off the top of my head:

* [x] [Fix the add-on title, description, and drop the UpdateURL](https://github.com/mozilla/testpilot-containers/pull/758)
* [x] Drop Telemetry
* [x] Upload new add-on to AMO. The current edit page is at https://addons.mozilla.org/en-US/developers/addon/containers-experiment/edit. 
* [x] Change the "How to distribute" option to be "On this site" instead of "on your own"
  * [x] https://addons.mozilla.org/en-US/firefox/addon/containers-experiment/ should start working
* [ ] @clouserw: Disable the install button on Test Pilot (publish the graduation report)
* [ ] @clouserw: can redirect existing users to the AMO version. They'll get the new version as their browser checks for updates (once a day)
* [ ] @groovecoder: Get ratings/users for new AMO version and then submit for featuring on AMO

wresuolc:
Oh, sure.  Mostly just dropping the updateURL (it defaults to AMO).  You can see the checklist I made for them at https://github.com/internetarchive/wayback-machine-firefox/issues/84

jonathanKingston:
@wresuolc I assume that we can't ever rename the extension can we? The id will always be that it is a test pilot experiment if so.

wresuolc:
It's a lot more complex to change the ID (basically involves you making a whole new add-on and publishing a new version of the old one that installs the new one).  I'd leave the ID alone, personally. :)

groovecoder:
I was able to upload the 3.1.0 internally-signed 'Mozilla Extension' to AMO, so next week we should be able to upload the next and change the "How to distribute" to "On this site"

groovecoder:
@jonathanKingston - do you know if the description on the AMO page is pulled from the `package.json`/`manifest.json` contents? If so, we'll need to update that as well.

groovecoder:
@jonathanKingston - nevermind, looks like summary + description both come from the AMO admin interface.

grahamperrin:
From <https://discourse.mozilla.org/t/-/18918/3?u=grahamperrin> (2017-08-07): 

> … automation was probably with Firefox Sync. …

tl;dr if I understand correctly, you may find migrations occurring sooner than originally intended where users of Firefox Sync allow synchronisation of extensions.

[...]

PR #758, opened 2017-08-23
for #665: update title, desc, version for AMO

groovecoder:
r? but we can wait to merge this until others are merged.


jonathanKingston:
Whoops merged, we can do a follow up for version and name tweaks if we need it @groovecoder

PR #762, opened 2017-08-24
Remove tab counting code as also not needed, change tab counting to b…

### LINKED ISSUE #750: Selected container tab's window doesn't get focus
Steps:

1. Create a new container tab and move it to a new window.
2. Go back to the previous window, click Containers -> View <name> container and finally the name of the tab.

Expected:

The window focus should change to the new window and the selected tab should get focus as well.

Actual:

The new window doesn't get focus. The tab itself in the new window *gets* focus (when there's multiple of them) so it seems it's just the window switching.

macOS 10.12.6
Nightly 57.0a1 (2017-08-22) (64-bit)
Containers 3.1.0

### LINKED ISSUE #753: Moving containers should it account for only the current window
@groovecoder I think the extension only operates on the current window. We should only move the current windows tabs right?

This and #750 I think break our existing expectations of how the addon used to work.

### LINKED ISSUE #756: An infinite number of each hidden container is shown when choosing "Show this container" option in Private Browsing mode
**[Notes]:**
- On the latest Firefox Release 55.0.2 the issue is not reproducible. However, when choosing the "Show this container" option, two of the same container tabs are opened in the non-private window.

**[Affected versions]:**
- Firefox unbranded build (56.0b2-20170810113054)
- Nightly build (57.0a1-20170822171841)

**[Affected Platforms]:**
- All Windows
- All Mac OS
- All Linux

**[Prerequisites]:**
- Have a Firefox profile with the latest Containers add-on version (3.1.0) installed.

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites.
2. Open two or more tabs of the same container type.
3. Click on the "Containers" toolbar button.
4. Choose the "Hide this container" option for the current container type.
5. Open a New Private Window (CTRL+Shift+P).
6. Click on the "Containers" toolbar button.
7. Choose the "Show this container" option for the hidden container tabs.
8. Switch to the non-private Firefox window and observe the browser behaviour.

**[Expected result]:**
- Only one instance of each hidden container tab is shown in the non-private window.

**[Actual result]:**
- An infinite number of each hidden container tab is shown in the non-private window.

**[Additional Notes]:**
- Attached a screen recording of the issue:
containers_issue

jonathanKingston:
…e per window. Prevent reopening of hidden tabs doesn't call itself. Fixes #750, Fixes #753, Fixes #756 

Also changed window objects to not shadow the main window object.

### DIFF
--- a/webextension/js/background/backgroundLogic.js
+++ b/webextension/js/background/backgroundLogic.js
@@ -52,6 +52,16 @@ const backgroundLogic = {
   },
 
   async openTab(options) {
+    const userContextId = ("userContextId" in options) ? options.userContextId : 0;
+    const cookieStoreId = backgroundLogic.cookieStoreId(userContextId);
+    // Unhide all hidden tabs
+    this.showTabs({
+      cookieStoreId
+    });
+    return this.openNewTab(options);
+  },
+
+  async openNewTab(options) {
     let url = options.url || undefined;
     const userContextId = ("userContextId" in options) ? options.userContextId : 0;
     const active = ("nofocus" in options) ? options.nofocus : true;
@@ -64,10 +74,6 @@ const backgroundLogic = {
       url = undefined;
     }
 
-    // Unhide all hidden tabs
-    this.showTabs({
-      cookieStoreId
-    });
     return browser.tabs.create({
       url,
       active,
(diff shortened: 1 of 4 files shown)


webextension/js/background/backgroundLogic.js
```diff
@@ -76,60 +82,61 @@ const backgroundLogic = {
...
 
-    const containerState = await identityState.storageArea.get(options.cookieStoreId);
+    const containerState = await identityState.storageArea.get(cookieStoreId);
     // Nothing to do
     if (list.length === 0 &&
         containerState.hiddenTabs.length === 0) {
       return;
     }
-    const window = await browser.windows.create({
+    const windowObj = await browser.windows.create({
```
groovecoder: Call this `newWindowObj` to differentiate it from the `windowId` above.

groovecoder:
No obvious bugs nor errors. But this seems to introduce a new model of the Containers-Windows-Tabs relationships that may be different from users' current mental model.

I.e., "Hide this Container['s tabs]" now means "Hide this Container['s tabs in the current window]"? "Show this Container['s tabs]" now seems to mean "Show this Container['s tabs from any window where they were previously hidden"

Combined with the low usage of show/hide/move tab features in the Test Pilot, it almost makes me want to file a follow-up to simply remove the show/hide/move features from our add-on, and let a 3rd-party add-on implement those.

jonathanKingston:
> No obvious bugs nor errors. But this seems to introduce a new model of the Containers-Windows-Tabs relationships that may be different from users' current mental model.

@groovecoder this was the bug I raised however this was actually previous behavior.

groovecoder:
Then yeah, as soon as we make the small var name change I'm happy to r+ and merge.

Issue #711, opened 2017-08-03
Can we do something about the context menu entry?


emmyemi:
While I like having easy access to the "Always open in ..." feature via the context menu, the current way this is handled is rather ugly-looking, especially with other menu entries surrounding it. Example:

![screenshot](https://user-images.githubusercontent.com/14862440/28948639-fe9bb510-7884-11e7-8ecd-b8087e87763a.png)

Instead, maybe we could change the icon to the left to a version of the extension icon with a colored checkmark overlayed or something. Perhaps the entry could be changed to read "Opening in `x` container" or "Do not open in ..." when the option has been selected instead, that way the checkmark won't be necessary. This isn't exactly a high priority thing, but a more elegant solution would be much appreciated.

jonathanKingston:
Valid points, there is a limitation on the space and also that we can't have a top level tick like check spelling has. We also have the limitation of not always being present due to timing of when the context loads and us looking through our storage to check if the page was assigned.

This leaves us with trying to think of new text (which we tried but nothing could be agreed on / was too long).

Another option is open the prompt instead where you can now assign there. So the text could just read "manage container". (we could also jump this to the edit container screen where you can see the current assigned containers, we could also have an add button there also). This would be an extra click but it could be more useful.

The last option I can think of is making it have a child menu like "languages" does in the screenshot and provide more than just the assignment functionality.

emmyemi:
> Another option is open the prompt instead where you can now assign there. So the text could just read "manage container". (we could also jump this to the edit container screen where you can see the current assigned containers, we could also have an add button there also).

I think this would be much more useful, as you said, and certainly easier on my OCD. 


> The last option I can think of is making it have a child menu like "languages" does in the screenshot and provide more than just the assignment functionality.

Though this could also work just as well. Anything would be a bit better than current, IMO.

jonathanKingston:
@groovecoder pick one and I can implement, I think the top one won't be possible until 55 (which we will likely wait for anyways). Unless you think design should be involved.

groovecoder:
Yeah, let's have @ChenMorpheus and/or @fangshih make the call here.

ChenMorpheus:
Hi @jonathanKingston @groovecoder, please find the proposals below. The original solution was in my previous spec. If it's not feasible, then we can go with solution B. However, if it's mandatory to have icon for every add-on, then we can go with solution C. What do you think?

[...]

PR #776, opened 2017-08-29
Add move and hide to context menu and neaten using checkboxes. Fixes …

jonathanKingston:
…#711

### DIFF
--- a/webextension/js/background/assignManager.js
+++ b/webextension/js/background/assignManager.js
@@ -1,6 +1,10 @@
 const assignManager = {
   MENU_ASSIGN_ID: "open-in-this-container",
   MENU_REMOVE_ID: "remove-open-in-this-container",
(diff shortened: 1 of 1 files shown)


webextension/js/background/assignManager.js
```diff
@@ -270,19 +293,37 @@ const assignManager = {
     if (siteSettings === false) {
       return false;
     }
-    // ✓ This is to mitigate https://bugzilla.mozilla.org/show_bug.cgi?id=1351418
-    let prefix = "   "; // Alignment of non breaking space, unknown why this requires so many spaces to align with the tick
+    let checked = true;
```
groovecoder: This default needs to be `false` or the menu shows the site assigned even when it's not.

jonathanKingston: True dat thanks

Issue #692, opened 2017-07-24
make containers min version >= 53


wresuolc:
On this line:  https://github.com/mozilla/testpilot-containers/blob/master/webextension/manifest.json#L14

PR #800, opened 2017-09-08
Add strict min version and extension id and bump version to 4.0.2. Fi…

jonathanKingston:
…xes #692 

@groovecoder I'm hoping that adding the addon id will fix windows in Nightly which doesn't seem to install the browser action, however we might also think this is sandboxing.

Could you double check the upgrade process on Mac (I'm always nervous of changes like this)?

The min version is to address #692

### DIFF
--- a/install.rdf
+++ b/install.rdf
@@ -12,7 +12,7 @@
     <em:targetApplication>
       <Description>
         <em:id>{ec8030f7-c20a-464f-9b0e-13a3a9e97384}</em:id> <!--Firefox-->
-        <em:minVersion>51.0a1</em:minVersion>
+        <em:minVersion>53.0</em:minVersion>
         <em:maxVersion>*</em:maxVersion>
       </Description>
     </em:targetApplication>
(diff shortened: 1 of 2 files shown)


jonathanKingston:
This is the error I was seeing which looked like it was related to why I
can't install the signed version.

groovecoder:
For whatever reason, this branch `.xpi` won't install into Nightly on Windows (Gives the above "not compatible" message), while `master` branch `.xpi` will.

Until we can figure it out, I suggest we hold this PR. 52 (ESR) users are 6% of traffic to the AMO listing, while 57 users are 35%. So we should promote the current 4.0.1 add-on as soon as possible to get 57 users to install while it works for them.

jonathanKingston:
@groovecoder tested again after rebase, I can pull the id part out however this works for me. Should we cut another release soon as we fixed a hide tabs issue right?
I can fix some other things tomorrow if we want to cut a bigger release.

jonathanKingston:
I'm keen to test if pushing a 4.x release fixes my main accounts container extension

groovecoder:
Now I can't get this branch to install into my Windows *NOR* Mac Nightly (58)?

jonathanKingston:
Closing in favour of #853 (I'm not sure why you are having issues installing this though)

PR #834, opened 2017-09-20
Store only one of the current version opened. Fixes #833

### LINKED ISSUE #833: Duplicate storage of browser action clicks
The version number that we keep in "browserActionBadgesClicked" doesn't check the unique number used, which is logging each time the browser action is opened. This will cause a huge storage issue in time.

The solution here should be to remove all duplicates to clear out the versions the user already has. I'm not sure if we want to rethink how we store this in future also however this can be an interim bug.

jonathanKingston:
Cleared up the legacy storage stuff that was sat in localStorage, it's pretty safe to say users have had this update based on the numbers (this update was back in july).

### DIFF
--- a/webextension/js/popup.js
+++ b/webextension/js/popup.js
@@ -93,18 +93,7 @@ const Logic = {
     const data = await browser.storage.local.get([ONBOARDING_STORAGE_KEY]);
     let onboarded = data[ONBOARDING_STORAGE_KEY];
     if (!onboarded) {
-      // Legacy local storage used before panel 5
-      if (localStorage.getItem("onboarded4")) {
-        onboarded = 4;
-      } else if (localStorage.getItem("onboarded3")) {
-        onboarded = 3;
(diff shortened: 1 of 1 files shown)


webextension/js/popup.js
```diff
@@ -144,7 +133,11 @@ const Logic = {
     browser.browserAction.setBadgeBackgroundColor({color: ""});
     browser.browserAction.setBadgeText({text: ""});
     storage.browserActionBadgesClicked.push(extensionInfo.version);
-    browser.storage.local.set({browserActionBadgesClicked: storage.browserActionBadgesClicked});
+    // use set and spread to create a unique array
+    const browserActionBadgesClicked = [...new Set(storage.browserActionBadgesClicked)];
+    browser.storage.local.set({
+      browserActionBadgesClicked
+    });
```
groovecoder: You and your clever JS applications. Thanks for the comment - would never have understood the reasoning behind this. 😜

Issue #894, opened 2017-10-05
Containers extension "forgets" my customer containers configuration, and closes container tabs


terrycloth:
This has happened to me twice in the past week or so: Soon after opening Firefox, Containers closed all my container tabs, *and* also lost/deleted my container configurations, resetting everything to the default. I'm not sure what's causing this, except maybe extension updates? Have there been a couple recent updates?

My system setup:

* Fedora 26 Workstation x86_64 (Linux)
* KDE Plasma Workspaces 5.10.5
* Firefox Developer Edition 57.0b5 (auto-updating static binary downloaded from official web page, rather than a distro package)
* all enabled extensions:
    * Multi-account Containers
    * Context Plus
    * Test Pilot
    * Snooze Tabs
    * uBlock Origin
    * Wayback Machine

Possibly related to #431.

emmyemi:
I've had this issue a couple of times as well. My setup, for reference:

* Windows 10 64-bit
* Firefox Dev Edition 57.0b5
* Extensions:
  * bitwarden
  * Context Plus
  * Decentraleyes
  * Firefox Pioneer
  * Multi-Account Containers
  * Smart HTTPS
  * uBlock Origin

Sometimes when this issue occurs, the extension is removed entirely and must be re-downloaded. Others, my configuration just disappears and I have to redo it all from scratch.

Notably, the extension remembers when I've set a site to always open in a specific container, and will automatically do so even if I don't tell it to. However, if the order of the re-configured containers is different from before, it will automatically open the site in whichever container happens to be in the slot that previously held the correct container. Example:

**Containers:**

1. Container A
2. Container B
3. Container C

* Site X is configured to automatically open in Container B.
* Settings are wiped, forcing me to redo.
* As I'm setting up my containers, I place a different one after Container A, changing the order of those after it. The order is now as follows:

1. Container A
2. Container D (new)
3. Container B
4. Container C

Although Site X is configured to open in **Container B**, upon re-configuring my containers, it will now auto-open in **Container D**, since that container now takes up slot 2. I'm unsure of whether this will be of any help, but hopefully it will, because I'm growing tired of having to redo my settings every few days.

hoosteeno:
[...]

PR #901, opened 2017-10-09
Change (un)install changes to only work for >57. Fixes #858 and Fixes…

### LINKED ISSUE #858: Confusing setting "Enable Container Tabs" under Privacy
With a brand new Firefox profile, if I go to Preferences, Privacy, I do not see an "Enable Container Tabs" checkbox.

Once I have installed the [Firefox Multi-Account Containers add-on](https://addons.mozilla.org/en-GB/firefox/addon/multi-account-containers/) from the add-ons directory, that setting appears and is checked by default. This checkbox seems redundant, after all, if the add-on is enabled, surely the feature of containers should be enabled as well.

Here is the most confusing part: if I disable the Firefox Multi-Account Containers add-on, containers contain to work, and the checkbox "Enable Container Tabs" continues to appear under Preferences, Privacy.

I would expect either the enabling of the add-on to control activation of containers, or the checkbox under Preferences, Privacy, but not a weird mixture of both.

jonathanKingston:
… #900 

@groovecoder - This code needs to be tested in 56 and 57 for upgrades/downgrades. Leaving the code here as a FYI. I think it is fine though.


bootstrap.js
```diff
@@ -130,41 +130,58 @@ async function install() {
...
+  });
+}
+
+function checkLegacyFirefox() {
   const version = Services.appinfo.version;
   const versionMatch = version.match(/^([0-9]+)\./)[1];
-  if (versionMatch === "55"
-      || versionMatch === "56") {
+  if (Number(versionMatch) <= 56) {
+    return false;
```
groovecoder: This logic seems backwards, or the function name is confusing. "LegacyFirefox" would be <= 56?

jonathanKingston: Yeah sorry I renamed last minute, will work on this tomorrow.

jonathanKingston:
@groovecoder I updated the code, I still need to check if this fixes the update issue I have been seeing in: #894 however I am worried that is a central issue still.

groovecoder:
Need to remove all the .zip files from this PR.

groovecoder:
Superseded by https://github.com/mozilla/multi-account-containers/pull/971 which removes and ignores the web-ext-artifacts zip files.

PR #911, opened 2017-10-18
Add delete numbered containers button

ironyman:
Containers are pretty good for web development because they're more permanent than private windows and less permanent than not private windows. I often end up with a bunch of containers that I use to test some feature and never use again. It's a pain to delete them one by one. This is an attempt at addressing this issue. I realise it's not useful for the average Firefox user though. Comments?

### DIFF
--- a/webextension/css/popup.css
+++ b/webextension/css/popup.css
@@ -467,6 +467,24 @@ manage things like container crud */
   background: #f2f2f2;
 }
 
+#edit-containers-panel #delete-numbered-link {
+  align-items: center;
+  block-size: var(--block-url-label-size);
+  border: 1px solid #d8d8d8;
+  border-radius: var(--small-radius);
+  color: var(--title-text-color);
+  display: flex;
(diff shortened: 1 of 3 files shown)


jonathanKingston:
I don't personally think we should add this at all. This is only useful for your use-case really which means it shouldn't be part of the core add-on.

However, there it could be it's own add-on instead now as WebExtensions can manage this and be published to addons.mozilla.org in the same way ours is. This would give you more flexibility to change more things too.

jonathanKingston:
Thoughts @groovecoder

groovecoder:
I agree. We won't have the bandwidth to maintain this feature (and it's potential data loss for users) in the core add-on, but it would make a great independent add-on using the [`contextualIdentities` API](https://developer.mozilla.org/Add-ons/WebExtensions/API/contextualIdentities).

groovecoder:
But thank you for submitting the PR! I think it's a good start and a good idea.

PR #921, opened 2017-10-26
Add share feature

groovecoder:
image

## To spot-check ...

1. Install this version of XPI
2. Open 5 container tabs (will bump to 100 for release)
3. [x] You should see a green "NEW" badge on the browserAction icon.
4. [x] When you click it, you should see the "achievement panel" with links to rate, share, tweet, and done.


webextension/js/background/messageHandler.js
```diff
@@ -120,6 +128,30 @@ const messageHandler = {
...
+    if (countOfContainerTabsOpened === null) {
+      countOfContainerTabsOpened = 1;
+    } else {
+      countOfContainerTabsOpened += 1;
+    }
+    console.log("incrementCountOfContainerTabsOpened, countOfContainerTabsOpened: ", countOfContainerTabsOpened);
+    browser.storage.local.set({[key]: countOfContainerTabsOpened});
+
+    // When the user opens their _ tab, give them the achievement
+    if (countOfContainerTabsOpened === 5) {
```
groovecoder: NEED TO BUMP THIS TO 500 BEFORE MERGE!

webextension/js/popup.js
```diff
@@ -121,6 +122,34 @@ const Logic = {
...
+      }
+    }
+    if (showAchievements) {
+      this.showPanel(P_CONTAINERS_ACHIEVEMENT);
+    } else {
+      this.showPanel(P_CONTAINERS_LIST);
+    }
+  },
+
+  async setAchievementDone(achievementName) {
```
groovecoder: Need to figure out better logic here.

groovecoder: Updated the logic. This is ready to review and spot-check now, @jonathanKingston

TanviHacks:
Lets make it 100 container tabs when this hits release users.

TanviHacks:
Also, how are we counting?  Any tab with a usercontextId!=0?  There are many entrypoints for opening container tabs, so better to count the tabs than the entrypoints.

groovecoder:
It's counting any container tab opened by any of the entry points when the user context is not `firefox-default` or `firefox-private`.

webextension/js/popup.js
jonathanKingston: So done means "read" essentially? It's a little confusing in name between getting the achievement and marking it done.

webextension/js/background/messageHandler.js
```diff
@@ -120,6 +128,29 @@ const messageHandler = {
     });
   },
 
+  async incrementCountOfContainerTabsOpened() {
+    const key = "containerTabsOpened";
+    const count = await browser.storage.local.get(key);
```
jonathanKingston: I think you can use `get({[key]: 1})` right?

jonathanKingston: Instead of the presence checking below etc

webextension/js/popup.js
groovecoder: Essentially ... I put "Done" on the button because the user has to actually click it to clear the achievement. (Hopefully after they rate/tweet/share) So "Done" means whatever the user chooses to do: just read it, or read + rate, or read + share, etc.

groovecoder: I'll probably keep the name; `setAchievementRead` would imply that the user only has to *see* the achievement for it to go away.

webextension/js/background/messageHandler.js
```diff
@@ -112,6 +112,14 @@ const messageHandler = {
       // lets remember the last tab created so we can close it if it looks like a redirect
       this.lastCreatedTab = tab;
       if (tab.cookieStoreId) {
+        // Don't count firefox-default, firefox-private, nor our own confirm page loads
+        if (tab.cookieStoreId !== "firefox-default" &&
+            tab.cookieStoreId !== "firefox-private" &&
+            tab.url.indexOf("confirm-page.html") === -1) {
```
jonathanKingston: Can we check instead for moz-extension protocol instead perhaps?

webextension/js/background/messageHandler.js
groovecoder: updated.

webextension/js/background/messageHandler.js
groovecoder: 👍

jonathanKingston:
I left two comments which would be cool if we could fix those to be a little simpler. If you could add a comment explaining when achievements are gathered and "done" too just for future when we forget.

Other than that everything looks fine, I haven't tested at the moment though.

groovecoder:
Updated. @SoftVision-EmilPasca - it's a tedious task, but could you test this branch that the achievement panel shows in the pop-up after opening 100 container tabs? (Easiest way is probably to assign a site to automatically open in a container, "Remember my decision", and then just open tabs to that site very quickly.)

SoftVision-EmilPasca:
@groovecoder I have made a build from add-share-feature branch and tested on Windows 10 x64, Mac 10.12 and Ubuntu 14.04 with the latest Firefox release (56.0.2), the latest Nightly (58.0a1-20171030103605) and latest Beta(57.0b13-20171030163911).

I have tried different scenarios in which I've created 100 container tabs or more using combinations of custom and default containers, hidden container tabs, non-container tabs, using multiple windows, private windows, different sessions or disabling/enabling the add-on and restarting the browser.

During testing I've observed that the "achievement panel" is successfully shown regardless of the scenarios above if a number of 100 container tabs have been reached by the user.
Also the "Rate", "Share" and "Tweet" buttons correctly redirects to the desired pages and the "Done" button correctly closes the "achievement panel" when pressed.

If you think there are any more scenarios which I haven't covered, please let me know.

groovecoder:
That's great, thanks @SoftVision-EmilPasca !

groovecoder:
@jonathanKingston can I get an r+ here now?

jonathanKingston:
Out of interest why are we using shortened urls for the sharing links?
Eg:
https://bit.ly/fb-share-mac-addon

Also these have a mac utm in their source?
- If we are worried about length of HTML lines we can solve that with vars or something else
- If we are worried about utm's being exposed when the user hovers over the URL, I would rather be genuine and show it all (I didn't actually test if hovering shows as it normally does)

... more to come end of train ride.

groovecoder:
Yeah, I'm not super-happy with it. But the `utm` params didn't seem to work when I put them straight into the url. But I'll try a couple encoding tricks to avoid short urls if at all possible.

jonathanKingston:
**1.**
> Yeah, I'm not super-happy with it. But the utm params didn't seem to work when I put them straight into the url.

Hmm that is odd, anyway can we change the "mac" part, unless there is an explanation for it?
Perhaps it is the tweet text having spaces, you could pass it through the method we have that uses encodeUriComponent and some other changes (that covers weird edge cases that impact extensions)?

The code is in the assignManager, feel free to take the output and hard code it into the HTML
```
  encodeURLProperty(url) {
    return encodeURIComponent(url).replace(/[!'()*]/g, (c) => {
      const charCode = c.charCodeAt(0).toString(16);
      return `%${charCode}`;
    });
},
```
**2.**
Also we can now do async arrows: `async () => {` feel free to fix this or ignore (we can file a follow up to fix it all instead).

**3.**
Can you change `tab.url.indexOf("moz-extension") === -1` with `!tab.url.startsWith("moz-extension")` as it prevents any edge case URLs that contain that in a param for some bizarre reason.

**4.**
Don't we have the following as a function call anyway?
```
+      browser.storage.local.set({achievements});
+      browser.browserAction.setBadgeBackgroundColor({color: "rgba(0,217,0,255)"});
+      browser.browserAction.setBadgeText({text: "NEW"});
```

**5.**
Are "webextension/img/fb-icon.png" and "webextension/img/twitter-icon.png" open source? Also is there a SVG version instead?

**6.**
`const countOfContainerTabsOpened = count[key] += 1;` to me implies you need to increment count too, which from what I can see you don't.

I would use: `const countOfContainerTabsOpened = ++count[key];`
or just reference `count[key]`:
```
count[key] += 1;
browser.storage.local.set({[key]: count[key]});
```

Other than that we are all good, feel free to merge after these 6.
None are blockers if you want to make them follow ups, although likely 1. utm code can't change after releasing to users.

groovecoder:
1. The 'mac' part in this case means "Multi-Account Containers" - not Mac OS.

2. Let's do a follow-up enhancement for async arrow functions

3. fixed

4. There's a `badge.displayBrowserActionBadge` function, but it would need refactoring to be useful here and in `popup.js`. Probably another follow-up/fix-up?

5. Switched to CC-Attrib SVG icons

6. Updated/fixed.

TanviHacks:
What happens if a user clicks "Tweet".  Is the tweet content pre-populated for them to share or manually edit?

groovecoder:
It's pre-populated but the user can change/update the text if they want.

PR #981, opened 2017-11-17
Add shortcuts for hiding and showing container groups

totokaka:
This PR adds the following features:

- Icon buttons to toggle a container's visibility from the container overview popup
- A button to unhide all containers, next to the "Sort tabs" button on the popup
- A button to show only tabs from a container on the container info pane of the popup

This is my first work with web extensions, and I'm open to suggestions for improvements and the like.
If this would be better as three different PRs, I would be happy to split it up.

### DIFF
--- a/webextension/css/popup.css
+++ b/webextension/css/popup.css
@@ -450,7 +450,7 @@ manage things like container crud */
   margin-inline-start: var(--inline-item-space-size);
 }
 
-#container-panel #sort-containers-link {
+#container-panel .action-link {
   align-items: center;
   block-size: var(--block-url-label-size);
   border: 1px solid #d8d8d8;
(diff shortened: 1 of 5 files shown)


gene1wood:
@totokaka Could you update your branch to resolve the conflicts that have come up in the last couple months?

@groovecoder Any thoughts on what else is needed to move this forward, looks like a great feature that I personally could really use.

groovecoder:
@totokaka - thanks for the PR!

Since this PR adds a number of UI elements and UX workflows. Every UI or UX element adds maintenance costs. (Sometime exponentially when conflicting UI surface can create bugs.)

If you're up for it, you can submit a new PR with conflicts resolved, and we will get a UI/UX reviewer for it.

Issue #462, opened 2017-04-25
Option to open new Tabs in the same Container


MichelZ:
I often CTRL+T to open a new Tab. Currently, these are opened in the Default Container. It would be nice to at least have a configurable option to open new Tabs in the Context of the Tab I'm currently working on.

I am a system admin, and I do need to use different personas, often even for the same sites. Also, when I am working with one persona, I often need multiple tabs for the same persona, so this would help me not needing to change the persona on every new tab.

As mentioned above, saving the URL to always open in the same Container does not work for me, as the same site needs to be used with multiple personas.

Nainterceptor:
I've suggested in test pilot feedback to be able to open with Ctrl/Cmd + T a new tab without container, then add Ctrl/Cmd + T + n (where n is a number) to open a new tab in the numbered container (by order in the list).

Another option may be to show another "new tab controls" option to see possible containers, then bind a key to each container, so we can open with Ctrl/Cmd + T then the key (better for fingers & accessibility, because some combinations like Ctrl/Cmd + T + n are not easy with only one hand)

My 2 cents ;)

anewuser:
@Nainterceptor Your suggestions are as valid as anybody else's, but please don't take the focus away from @MichelZ's idea, which is simply to make new tabs open in the same container as the current tab with the 2-key shortcut we've been used to for many years.

Nainterceptor:
@anewuser Mhhh, nevermind, I've understood "vote for me" label as an open debate. I guess that is not one.

hobarrera:
Typing in the location bar and hitting <kbd>alt</kbd>+<kbd>enter</kbd> should **definitely** open in the same container, regardless of this particular request's outcome.

TitanNano:
Also just clicking the "new tab" button should open the tab in the same container. Maybe the color of the + can be changed to the current container color.

groovecoder:
I'm pretty sure this needs code in m-c. I can't seem to get the right combination of tab event handlers to make this work in a browser extension. Here's where I'm trying:

https://github.com/groovecoder/new-tab-same-container

jonathanKingston:
@groovecoder can you not combine:
https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webNavigation/onCreatedNavigationTarget

with a blocking onBeforeRequest?

[...]

PR #982, opened 2017-11-17
[WIP] Add key combination to open new tab in the same container.

totokaka:
I was let down that this addon contains no functionality to make Ctrl+T and Ctrl+Alt+Enter open the new tab in the same container. See issue #462.

I found that this is also tracked in [Bugzilla Bug 1406371](https://bugzilla.mozilla.org/show_bug.cgi?id=1406371).

This PR adds the key combination Ctrl+Y as a workaround until this is implemented in Firefox.

This is still [WIP], as the command is not called when the keys are pressed unless you have the popup open, for some reason. Did I place the command handler in the wrong place?


totokaka:
Closing this in favour of #990

PR #986, opened 2017-11-19
Make it possible to reopen current tab in other container #942

### LINKED ISSUE #942: Open Current URL in Container X
For security, privacy, and complexity reasons, we do not want the user to have the ability to "switch Containers" within a given tab.  We want don't want to imply that we have fixed the users mistake (i.e. maybe they logged into social.com in their Work Container, then realized it, and "switched Containers" to Social Container.  They are still logged in in the Work Container)

But users do often make mistakes and end up on the wrong link in the wrong Container.  This is not always detremental to security or privacy.  Perhaps they are visiting a relatively benign site with few trackers in the wrong Container, and realize that they aren't logged in.  To recover from this, right now the user has to copy the link, open a tab in the right Container (which may or may no be right next to the tab that they are already using - i.e. the new tab may end up at the end of a line of many tabs), paste the link and go to it.  We should make this a couple steps easier.

I propose that some where in the Location Bar, we have an option to "Open this link in Container X".  The best UX way I can picture this is a drop down on the colored Container Name and Icon in the URL bar.  When the user clicks the drop down, they can select which Container they want to open the given URL in.  Selecting the Container will not "switch" the current tab.  Instead, it will create a new tab right next to it in the selected Container.  The site will then be open in two tabs, side by side, with different Container types.

Morpheus, what do you think of this proposal?  Can we have a basic UI design for this?  Thanks!

totokaka:
Fixes #942. Implements @ChenMorpheus' design, though not fully.
Misses styling. Both in terms of of the drop down looks, and that the identities doesn't have colors and icons.

I do not have the css experience to style the drop-down menu of the select form. Just styling the form border proved difficult as I had to remove default behaviour, and then had problems reimplementing hover style. I found no way to display images in the dropdown either.

### DIFF
--- a/webextension/css/popup.css
+++ b/webextension/css/popup.css
@@ -514,8 +514,21 @@ span ~ .panel-header-text {
 }
 
 #current-tab > label {
-  display: contents;
   font-size: var(--small-text-size);
+  grid-column: 2 / 4;
+  grid-column-start: 2;
+}
+
+#current-tab select {
+  color: var(--text-normal-color);
+  font-size: var(--small-text-size);
+  inline-size: var(--inline-button-size);
+  justify-content: center;
+  text-decoration: none;
+}
+
+#current-tab > label[for="container-page-assigned"] {
+  display: contents;
 }
 
 #current-tab > label > input {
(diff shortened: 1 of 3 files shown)


groovecoder:
@totokaka - thanks for another PR!

If you're up for it, can you submit a new PR with conflicts resolved, and we will get a UI/UX reviewer for it.

Issue #943, opened 2017-11-10
Create Container Windows


TanviHacks:
Create an Option in the PopupMenu to open a Container "window".  If I create a Container Window for "Personal", then clicking CtrlT or the plus button will continue to open Personal tabs.

We can decide whether other Container types are permitted to open in that Personal Window or not.  (Deciding that they can't, may be more engineering work than it's worth.  If a user creates a container window and then explicitly opens another Container, then maybe we shoudl just let them and assume they know what they are doing and no longer want a Personal Window.  Once they do that, it becomes a regular window instead of a Personal Window.  That may be easiest from an implementation stand point.)

Morpheus, what do you think of this Proposal?

We would have to figure out where in the Popup Menu to include the option for Container Windows.

ChenMorpheus:
@TanviHacks The proposal you have here shares similar conceptual models with the prototype in this [document](https://docs.google.com/document/d/1_AQi0EvWLOYMdPQxuMzs489q3JqRJmUokld8906C-0E/edit#heading=h.hmger4fymd19), which is different than what we provide now. I'm concerned that the proposal would cause various use cases, which cause confusions to users as well. As a result, I would suggest not mixing these two models in one product. 

Besides, as Fang and I mentioned in #844 and #871, we strongly recommend following the visual spec to solve some UI issues before diving into new features. If you need any help from us for any measurement-related questions, please let us know, thank.

jordanweaver:
Without access to the mentioned document, I can only comment on the proposal here and that is to say that I hope this (in some form) makes it on the roadmap. At the moment, I'm finding it very cumbersome to manage just two containers (work and personal), where at any moment I can mistakenly open a non-container tab that starts generating its own history/prefs/downloads/etc. I work on multiple projects for various entities and so this is enough to make me hesitate on fully migrating over to Firefox (which I would love to do).

Having the option to make the container distinction at the Window-level feels more "container-y" so I don't see why this would confuse users... rather it feels just like a less granular application of the container.

As I'm sure you're aware, Chrome provides this UX via "Users" with independent profiles and I suspect you'd get more users to convert by offering something similar. I realize that Firefox still contains a Profiles feature, but it doesn't currently have a great UX (multiple app instances are required, requires the CLI, etc). 

When/if this feature is worked on, some nice UI additions could be:

- quick switching between Window Containers via keyboard shortcuts
- visual cues in the Window UI to indicate which Container

[...]

Issue #544, opened 2017-05-27
Add context popup for containers when creating a tab (particularly via keyboard shortcut)


clairity:
this is somewhat related to #245, #311, and #518. i'd love cmd-t (and equivalents on other OSes) to provide a context menu to choose the container type to open. so 

1. `cmd-t` to signal the desire for a new tab
2. display context menu of containers available, numbered so i can choose the type with a number
```
    0  Default
    1  Personal
    2  Work
    3  Financial
```
3. type number of container desired for the tab (e.g., type in `1` for a Personal container)

this might need a configuration setting for how `cmd-t` behaves since some people may not want the extra context menu, or this flow can be triggered by a more complicated keyboard shortcut (like `ctrl-cmd-t` on mac). 

this context menu also could be used when cmd-/ctrl-clicking links or converting tabs from one type of container to another (selectable via keyboard number, click or touch, and possibly auto-hiding after a couple seconds or via `esc`).

groovecoder:
As of this week, there's a keyboard shortcut to open the pop-up, and to tab down to a container. So the steps are:

1. Ctrl+y (or Cmd+y on Mac)
2. Tab down to the container you want
3. <enter> to open a tab in that container

jonathanKingston:
We have not announced this feature yet as if you have a text field focussed this doesn't work. This includes the URL bar too.

I'm looking into ways we can expose this more. Adding numbers for keyboard navigation is certainly something we can do for #512.

jonathanKingston:
I have also raised #546 to move this to Ctrl+Comma as it has URL bar issues.

clairity:
@groovecoder @jonathanKingston thanks for the update!

the only disadvantage of `ctrl-y` or `ctrl-,` is that it's another key combo to remember vs. overriding the already existing `ctrl-t` shortcut (from the user's perspective, the browswer just automagically realizes you are using containers and asks you to choose which one you would like).

and tabbing to select in a dropdown list is not as intuitive (to me) as either (1) using arrow keys or (2) choosing numbers, but if that's the implementation, then i can live with it.

in any case, focusing on the url bar afterwards is definitely key (pun intended) to an efficient keyboard-based workflow. 😊

jonathanKingston:
@clairity focusing the URL bar will come back (this is available in Nightly and I think Beta), we are not sure about the keyboard shortcut alternative as of yet.

terrycloth:
[...]

Issue #448, opened 2017-04-12
Opening a URL using enter key modifiers does not open in same container tab


milescrabill:
Steps to reproduce:

- Open a container tab
- In the URL bar of the container tab, enter a URL
- Press option+enter (open URL in new tab) or another URL open modifier

Expected Result:
URL opens in new container tab, the same container as the URL was opened from

Actual Result:
URL opens in new default tab, not in a container tab of the same container as opened from

grahamperrin:
Is this still an issue? 

I can't reproduce it. 

For me with Firefox `88`, neither <kbd>Alt</kbd>-<kbd>Return</kbd> nor <kbd>Alt</kbd>-<kbd>Enter</kbd> opens a new tab in the context of a URL entered at **about:newtab**

dannycolin:
Same here, Alt+Enter does nothing in 110. Closing since we didn't get an answer back from OP and 2 persons can't reproduce it.

jwbth:
I assume the OP just missed the step "Open any URL in the container tab", with which it's easily reproducable, while without it, Alt+Enter just proceeds to open the URL in the same tab (which is in itself a questionable behavior, but that would be offtopic). The issue is of the same class as #462, but  for opening a URL you have already typed in the address bar.

Issue #406, opened 2017-03-29
Typing an URL in a container's address bar does not open the new tab in the same container


klint:
My Firefox is set to open new bookmarked linked or manually entered URLS in the address bar within new tabs.
When I do this from a tab which is in a container, I would expect the new tab to be opened in the same container (unless of course there would be a rule assigning an URL pattern to a certain container, but that feature is not in place yet). At least this is what I was expecting without thinking to it... and I was surprised by the result:  the new tab opened within no container.

IMHO it should have opened in the same context/container as the source tab.

PR #990, opened 2017-11-20
Open new tabs in same container as previous tab

### LINKED ISSUE #462: Option to open new Tabs in the same Container

totokaka:
This PR adds functionality that will try to open new tabs in the same container as the previously active tab in that window, if the new tab doesn't already have a container, and the previously active tab did.

Some undocumented / implicity functionality in the webextension API is used:
 * `onCreated` is called for a new tab before `onActivated` for that tab is called
 * `onUpdated` will be called for all new tabs, also `about:newtab` and the like (sets the favicon)
 * The first `onUpdated` for a new tab, that is not an `about:`-tab, will set the URL

It is a little bit wonky, as the original new tab is created and displayed (although not loaded) before it can be closed and a new tab can be created. However, it is the best workaround I can find to add this functionality until [Bugzilla 1406371](https://bugzilla.mozilla.org/show_bug.cgi?id=1406371) is solved.

This would fix #462, #448 and #406 (I think)
Also relates to #943 and #544

### DIFF
--- a/webextension/js/background/assignManager.js
+++ b/webextension/js/background/assignManager.js
@@ -177,6 +177,93 @@ const assignManager = {
     browser.webRequest.onBeforeRequest.addListener((options) => {
       return this.onBeforeRequest(options);
     },{urls: ["<all_urls>"], types: ["main_frame"]}, ["blocking"]);
+
+    // We need to keep track over the last activated tab, in order to open new
+    // tabs in the correct container. See the comments in _onTabCreated for
+    // more details
+    this.lastActivatedTab = new Map();
+    browser.tabs.onActivated.addListener((info) => {
+      this._onTabActivated(info);
+    });
+    // We need to query for the currently active tabs, to support
+    // hot-reloading of the extensions, and possibly also for functioning on
+    // browser startup
+    browser.tabs.query({active: true}).then(active => {
+      active.forEach(tab => {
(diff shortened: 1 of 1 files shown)


mandawan:
Thanks for the PR ! Hope it'll be reviewed merged, and deployed on store soon.

beatgammit:
Also relates to #1000

jonathanKingston:
Thanks for the PR!
However we will be moving forward with the platform bug instead:  [Bugzilla 1406371](https://bugzilla.mozilla.org/show_bug.cgi?id=1406371) this hopefully should land in Nightly this week.

m4dc4p:
Been a year - the platform bug never landed AFAICT. Could this PR get revived? I would really like functionality similar to the "sticky containers" extension (which I would use if it wasn't so buggy).

PR #992, opened 2017-11-21
Adds shortcut to open containers.

crenwick:
Makes it more effective to open the first ten containers in the
containers list. Keys 1-0 will open containers 1-10, if that container
exists.

Note: the plugin's popup panel does not autofocus when opened. This
requires the user to focus on the panel by either clicking or pressing
TAB before using shortcut keys. This behavior is consistent with the previous shortcuts for
this addon.

### DIFF
--- a/webextension/js/popup.js
+++ b/webextension/js/popup.js
@@ -516,12 +516,48 @@ Logic.registerPanel(P_CONTAINERS_LIST, {
           previousElement.focus();
         }
       }
+      function select(num) {
+        const element = selectables[num];
+        if (element) {
+          element.click();
+        }
+      }
       switch (e.keyCode) {
       case 40:
         next();
         break;
       case 38:
         previous();
+        break;
+      case 49:
+        select(1);
+        break;
+      case 50:
+        select(2);
+        break;
+      case 51:
+        select(3);
+        break;
+      case 52:
+        select(4);
+        break;
+      case 53:
+        select(5);
+        break;
+      case 53:
+        select(6);
+        break;
+      case 53:
+        select(7);
+        break;
+      case 53:
+        select(8);
+        break;
+      case 53:
+        select(99);
+        break;
+      case 48:
+        select(10);
         break;
       }
     });
--- a/webextension/js/popup.js
+++ b/webextension/js/popup.js
@@ -529,35 +529,13 @@ Logic.registerPanel(P_CONTAINERS_LIST, {
       case 38:
         previous();
         break;
-      case 49:
-        select(1);
-        break;
-      case 50:
-        select(2);
-        break;
-      case 51:
-        select(3);
-        break;
-      case 52:
-        select(4);
-        break;
-      case 53:
-        select(5);
-        break;
-      case 54:
-        select(6);
-        break;
-      case 55:
-        select(7);
-        break;
-      case 56:
-        select(8);
-        break;
-      case 57:
-        select(9);
-        break;
       case 48:
         select(10);
+        break;
+      default:
+        if (e.keyCode >= 49 && e.keyCode <= 57) {
+          select(e.keyCode - 48);
+        }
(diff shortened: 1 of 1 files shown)


totokaka:
Nice! This seems like a nice way to open tabs in specific containers in a fast way.

However, It didn't quite work when I tried it: I had to manually click somewhere in the popup for the shortcuts to work.

crenwick:
@totokaka unfortunately that's a plugin behavior. For some reason, the popup doesn't get autofocused on open. However, pressing tab does focus it.

My general flow is now:
1. `Cntrl+.` to open the panel
2. `Tab` to focus the panel
3.  `1` to open a new tab with the first container on the list

The focus issue existed before this PR, and affects other shortcuts.

poletniy:
It's a great idea! Just do not limit yourself to the first 10 containers. Please make a hotstring. For example, Ctrl + 55 opens the container # 55.

jonathanKingston:
@groovecoder thoughts?

webextension/js/popup.js
```diff
@@ -516,13 +516,49 @@ Logic.registerPanel(P_CONTAINERS_LIST, {
...
+        break;
+      case 56:
+        select(8);
+        break;
+      case 57:
+        select(9);
+        break;
+      case 48:
+        select(10);
+        break;
```
groovecoder: Can this be consolidated into a `default` case instead?
```
case 48:
  select(10);
  break;
default:
  if (e.keyCode >= 49 && e.keyCode <= 57) {
    select(e.keyCode - 48)
  }
  break;
```

webextension/js/popup.js
```diff
@@ -516,13 +516,27 @@ Logic.registerPanel(P_CONTAINERS_LIST, {
...
+        }
+      }
       switch (e.keyCode) {
       case 40:
         next();
         break;
       case 38:
         previous();
         break;
+      case 48:
```
groovecoder: actually let's remove this case for `10` too; `0` as `10` may confuse folks.

crenwick: Done. But I'll be sad to see my tenth shortcut key leave :(

groovecoder:
Thanks!

hoshsadiq:
This doesn't seem to work on MacOS. As @crenwick said, I press <kbd>ctrl</kbd>+<kbd>.</kbd> then <kbd>tab</kbd> then <kbd>1</kbd> but nothing happens.

stoically:
@hoshsadiq It's not yet published (last published version of MAC is from Nov 17). Will probably come with the next version I guess.

groovecoder:
Yes, planning to release today or tomorrow.

ghostofkendo:
The `Ctrl+.` shortcut does not work when using an AZERTY-based keyboard layout, because this layout requires you to hold `Shift` to produce the `.` character, see https://en.wikipedia.org/wiki/AZERTY#/media/File:KB_France.svg

Maybe you could add `Ctrl+Shift+.` in addition to `Ctrl+.` as shortcuts to open the addon's popup menu?
Or allow end-users to change the shortcut?

groovecoder:
Changing shortcuts opens an exponentially larger number of potential issues. :/

Can you file a new issue to support AZERTY-based keyboard layout? I'll label it as a good first bug and help wanted.

dbdr:
For reference, there is now an issue for the AZERTY case: #1081

cecurin:
`Ctrl`+`.`+`Tab`+`1` is broken :(
There's no way to open a new tab from a container directly? I mean If I want the first container, I could press `Ctrl`+`.`+`1` maybe?

soumendrak:
IMO the shortcuts should be like `Ctrl + Shift + 1` for 1st container and `Ctrl + Shift + 2` for 2nd container, subsequently.

PR #1021, opened 2017-12-02
Maintain the active state of the redirected tab

### LINKED ISSUE #442: Middle click on assigned link forces focus to new tab
Usually, hen i use middle click, the corresponding tab is opened in background, which is what I want.
But when I middle click to an url for which there is an assignment different than current container context, the new tab has focus.
Consider or example:

*   `github.com` is assigned to `GH` container
*   I am on a page in no container
*   I middle click on the link `github.com`
*   the new (temp) tab opens in background (which seems fine)
*   the extension copy the tab in a new (final) `GH` tab and destroy the temp tab
*   the final tab is focused (this is not wanted)

### LINKED ISSUE #923: Unexpected switch to a new tab
When opening a new tab in Firefox you are not automatically redirected to it, you stay on the tab on which you were. But when opening a tab in a container, you are redirected to this tab.
For a more streamlined experience I think you should not be automatically redirected to the container when opening it.

Running Containers 4.0.3 on Firefox 56.0.2 on Windows 8.1

gaborluk:
This mini changeset ensures that redirected tabs that have been opened in the background (with a middle click, for instance) don't steal the focus from the current tab. Particularly useful for opening up several top sites in a row from the New Tab page, each requiring a redirect.

Fixes #442, fixes #923.

### DIFF
--- a/webextension/js/background/assignManager.js
+++ b/webextension/js/background/assignManager.js
@@ -144,7 +144,7 @@ const assignManager = {
       return {};
     }
 
-    this.reloadPageInContainer(options.url, userContextId, siteSettings.userContextId, tab.index + 1, siteSettings.neverAsk);
+    this.reloadPageInContainer(options.url, userContextId, siteSettings.userContextId, tab.index + 1, tab.active, siteSettings.neverAsk);
     this.calculateContextMenu(tab);
 
     /* Removal of existing tabs:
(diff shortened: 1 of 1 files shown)


I-am-me:
Thanks 😃

gaborluk:
Sure thing! Thanks, guys.

PR #1034, opened 2017-12-05
id:@testpilot-containers to manifest.json; 5.0.1

groovecoder:
`manifest.json` needs `id: @testpilot-containers` to make sure it updates the existing 4.x legacy add-on instead of making a new one.


groovecoder:
@jonathanKingston - please be sure to spot-check this that the web-ext update doesn't destroy any data. I checked on mine and it was fine, but want to be extra careful since it's a big change.

Issue #929, opened 2017-10-31
Disposable (throw-away) Containers


girst:
I'd like to have an option to generate a new temporary container on-the-fly - sort of a private tab instead of a private window. These disposable containers should be destroyed automatically, when the last tab is closed.

This is how I imagine the workflow: 
 * right click a link -> Open in New Container Tab -> Disposable Container
 * click tool bar icon -> (+) New Disposable Container

in https://github.com/mozilla/multi-account-containers/issues/429 (closed by the author), the addon [Containers on the Go](https://addons.mozilla.org/en-us/firefox/addon/containers-on-the-go/) has been mentioned, which seems to implement this feature, but is [proprietary software](https://addons.mozilla.org/en-us/firefox/addon/containers-on-the-go/license/1.14).

groovecoder:
I'm starting a new label called "good extra addon" for issues like this. We are pointing add-on authors to this repo for ideas for "follow-up" addons that extend on the core MAC add-on.

Narvey:
@groovecoder if that's going to be the model, why not make the core features built-in instead of an addon?

groovecoder:
@Narvey that's a good point.

Over the course of the Test Pilot experiment, we learned that Containers is still mostly a power-user feature. Distributing as an add-on works for power users, and it lets us work on the code-base outside of the Firefox release models.

stoically:
I'm currently working on a MIT version of [Temporary/Throw-away Containers](https://addons.mozilla.org/en-US/firefox/addon/temporary-containers). It lacks any possibility to set personal preferences yet, but I'm planning to add them soon. If you have any suggestions please let me know.

stoically:
@groovecoder I've implemented a feature for the Add-on where it'll automatically open tabs in temporary containers when Middle- and/or Ctrl+Left Mouse Clicking on links. Thing is that I have to [deal with race-conditions](https://github.com/stoically/firefox-add-on-temporary-containers/blob/master/test/background.mac-race-conditions.test.js) where MAC also wants to reopen the link, either for the confirm page ("always open in $") or directly in the target container ("remember my choice").

[...]

PR #1095, opened 2018-01-25
Allow webextensions with contextualIdentities permission to get assignments

stoically:
Hi,

thanks for your work on Multi-Account Containers!

I've developed an "[extra Add-on](https://github.com/mozilla/multi-account-containers/issues/929)" named Temporary Containers ([AMO](https://addons.mozilla.org/en-US/firefox/addon/temporary-containers) / [GitHub](https://github.com/stoically/firefox-add-on-temporary-containers)) and in order to support some of the available features (e.g. automatically open clicked links in a Temporary Container) I have to rely on fuzzy logic to make a best guess about when MAC is going to reopen an URL in a Container or Confirm Page. This leads to random bugs where Tabs get closed or multiple Tabs get opened - up to the point that people [decide against using my Add-on](https://github.com/stoically/firefox-add-on-temporary-containers/issues/15#issuecomment-360484034).

If there's a chance that this PR can get merged I could reduce the number of possible problems when MAC and my Add-on get used in combination.

Please let me know if there's anything I can do to make a merge more likely.

Usage example from a whitelisted extension:
```js
const assignment = await browser.runtime.sendMessage('@testpilot-containers', {
  method: 'getAssignment',
  url: 'http://example.com'
});
```
Best regards

### DIFF
--- a/src/js/background/messageHandler.js
+++ b/src/js/background/messageHandler.js
@@ -72,6 +72,42 @@ const messageHandler = {
       return response;
     });
 
+    // Handles external messages from webextensions
+    const externalExtensionAllowed = {};
+    browser.runtime.onMessageExternal.addListener(async (message, sender) => {
+      if (!externalExtensionAllowed[sender.id]) {
+        const extensionInfo = await browser.management.get(sender.id);
+        if (!extensionInfo.permissions.includes("contextualIdentities")) {
+          throw new Error("Missing contextualIdentities permission");
+        }
+        externalExtensionAllowed[sender.id] = true;
+      }
+      let response;
+      switch (message.method) {
+      case "getAssignment":
+        if (typeof message.url === "undefined") {
+          throw new Error("Missing message.url");
+        }
+        response = assignManager.storageArea.get(message.url);
+        break;
+      default:
+        throw new Error("Unknown message.method");
+      }
+      return response;
+    });
+    // Delete externalExtensionAllowed if add-on installs/updates; permissions might change
+    browser.management.onInstalled.addListener(extensionInfo => {
+      if (externalExtensionAllowed[extensionInfo.id]) {
+        delete externalExtensionAllowed[extensionInfo.id];
+      }
+    });
+    // Delete externalExtensionAllowed if add-on uninstalls; not needed anymore
+    browser.management.onUninstalled.addListener(extensionInfo => {
+      if (externalExtensionAllowed[extensionInfo.id]) {
+        delete externalExtensionAllowed[extensionInfo.id];
+      }
+    });
+
     if (browser.contextualIdentities.onRemoved) {
       browser.contextualIdentities.onRemoved.addListener(({contextualIdentity}) => {
         const userContextId = backgroundLogic.getUserContextIdFromCookieStoreId(contextualIdentity.cookieStoreId);
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -26,6 +26,7 @@
     "contextualIdentities",
     "history",
     "idle",
+    "management",
     "storage",
     "tabs",
     "webRequestBlocking",


groovecoder:
Thanks for the PR @stoically. I'm going to add @jonathanKingston and @TanviHacks on this to decide if we want to set this precedent of explicitly allow-listing additional add-ons to receive messages.

jonathanKingston:
I think we would be fine with this @stoically however I suggest using the management API: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/management/get

We would need the new permission for containers but I am ok with us adding that.
Once you have this check if the extension has the contextualIdentity permission, that way we don't need the allow-list.

stoically:
@jonathanKingston @groovecoder @TanviHacks That's great news, thanks! 

I've changed the code accordingly and also added throwing errors so webextensions know if something went wrong. Hope that's OK. Let me know if you want to see something changed.

stoically:
Added an allowed-cache to avoid calling `browser.management.get` for every allowed message. The allowed-cache gets cleared if:
* MAC gets updated or disabled/enabled
* An extension in the allowed-cache gets installed/updated (using `browser.management.onInstalled`)

LarfAndAHarf:
Requiring new permissions from all users, just to support a third party add-on? I understand the pro-extensibility attitude, but you're already getting negative reviews for it. Worse though is that PLUS never having real release notes on addons.mozilla.org; just a git log. We shouldn't have to switch sites and search through developer issues (or even know what github is) to understand why an extension was proactively blocked from updating, and how to make an informed decision on whether or not to keep using it. People choose foss and Firefox for a reason. This extension as great, its PR is ...wanting.

stoically:
> This extension as great, its PR is ...wanting.

Just to clarify, my initial PR proposed to whitelist other extensions and not adding a new permission.

LarfAndAHarf:
Ack... sorry @stoically . Temporary Containers is great and your PR was great.
> its PR is ...wanting

I actually meant **P**ublic **R**elations for this addon's release process. Terminology fail, sorry!

stoically:
> Terminology fail, sorry!

Ah, ok, no problem - and thanks.

I have to admit though that I should've at least raised my concerns with adding a new permission that needs confirmation from users "just for my third party add-on". On the other hand, there's a team behind MAC and I'm sure they put thought into this and didn't decide it light-heartedly - probably with exactly the pro-extensibility arguments you mentioned in mind. I do however agree that this should be more pro-actively communicated. Coincidentally, today was a mail sent out to Add-on developers including a section about "Establishing trust with your users" and recommending "Describe why your extension needs its permissions in your AMO description."

Here's a proposal to add to the MAC AMO description:

---

**Multi-Account Containers asks for permissions upon installation:**

* **"Access your data for all websites"**: Required to provide the feature that enables automatic reopening of websites in target containers by selecting "Always open in" in the toolbar popup. In order to do that, it is necessary to be able to intercept and potentially cancel every request inside the browser. This is the [all_urls](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Match_patterns#%3Call_urls%3E) permission in combination with the [webRequest and webRequestBlocking](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webRequest) permissions.
* **"Access browsing history"**: Required to remove entries from the history generated from Multi-Account Containers when showing the Confirm Page which lets you decide in which container you want the website to open in if the website if assigned using "Always open in". This is the [history](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/history) permission.
* **"Access browser tabs"**: Required to access the tabs URL, which is also needed to provide automatic reopening of websites in the correct target container in cases where the currently loaded URL in the tab is relevant to decide whether to stay in a container or open a new one. This is the [tabs](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs) permission.
* **"Monitor extension usage and manage themes"**: Required to provide interoperability with other container Add-ons by checking if they have the required permissions. This is the [management](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/management) permission.

More required and used permissions included in the Add-on that don't show up as special text when installing the Add-on are:

* **[contextMenus](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/contextMenus)**: Required to add entries to context menu entry when right clicking.
* **[contextualIdentities](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/contextualIdentities)**: Required to use the container feature itself, which is natively available in Firefox with this permission.
* **[cookies](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/cookies)**: Required to [manage tabs based on the "cookieStoreId"](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/query), which is the internal way of handling tabs in combination with the container feature.
* **[storage](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/storage)**: Required to store and read preferences/options and data needed to provide functionality.

[You can find tips for assessing the safety of an extension here](https://support.mozilla.org/en-US/kb/tips-assessing-safety-extension) and [more general explanation about permission requests here](https://support.mozilla.org/en-US/kb/permission-request-messages-firefox-extensions).

While we provide overviews of what these permissions messages mean and how to judge if they're appropriate

elewis:
Completely agree with @LarfAndAHarf . The only indication in the browser UI is a tooltip asking to approve the "Monitor extension usage and manage themes" permission. 'Monitor extension usage' sounds especially nasty without the extra context here (to me that indicates the addition of telemetry monitoring features to MAC, which obviously isn't ideal in a privacy oriented extension).

I don't want to gripe too much, but it's definitely something to keep in mind. I'm a user of both of these extensions and love them so far. Thanks to @stoically and the MAC team for your work!

PR #1106, opened 2018-01-30
Fix for siteSettings is null

timendum:
siteSettings can be null, with this change we can avoid an exception.


groovecoder:
Thanks for the PR @timendum. Is there an existing issue or bug this fixes?

timendum:
Hi @groovecoder,
I check the browser console sometimes and "`siteSettings is null`" message appears quite often.

https://i.imgur.com/EHDCoSz.png

PR #1125, opened 2018-02-19
Add 'Reopen in container' entry in context menu

### LINKED ISSUE #1124: [Feature Request] Add "Open current Tab/Link in New Container Tab" menu in Tab context menu.
Please look at this screenshot if you don't understand (my english is bad) the title. 
https://imgur.com/a/Goooa

tunefish:
Adds a new entry in the context menu of tabs and websites that allows the user to reopen the current tab in a different container.
This PR extends features introduced in #986 and fixes #1124.

There are three more ways how to improve my solution:

- In the context menu of a website, all containers are shown except the container of the current tab.
However, I could not find a way to achieve the same for the context menu in tabs, thus all containers are shown there.
- Make it possible to reopen a tab in the default container. I tried opening a new tab with `cookieStorageId = 'firefox-default'` and no `cookieStorageId`, but Firefox seems to automatically open the tab in the current container.
- The icons for the menu entries could be colourized.

### DIFF
--- a/src/js/background/assignManager.js
+++ b/src/js/background/assignManager.js
@@ -188,7 +188,7 @@ const assignManager = {
         active: info.frameId !== undefined,
         cookieStoreId: reopenCookieStoreId,
         index: tab.index + 1,
-        openerTabId: tab.id,
+        openerTabId: tab.openerTabId,
         pinned: tab.pinned,
         url: tab.url,
       }).then(() => {
@@ -360,6 +360,20 @@ const assignManager = {
       id: this.MENU_REOPEN_IN,
       title: "Reopen in container",
       contexts: ["all", "tab"],
+    });
+
+    browser.contextMenus.create({
+      id: this.getCookieStoreIdFromReloadMenuId("firefox-default"),
+      title: "Default",
+      contexts: ["all", "tab"],
+      parentId: this.MENU_RELOAD_IN,
+    });
+
+    browser.contextMenus.create({
+      id: "reopen-separator",
+      contexts: ["all", "tab"],
+      type: "separator",
+      parentId: this.MENU_RELOAD_IN,
     });
 
     const identities = await browser.contextualIdentities.query({});
--- a/src/js/background/assignManager.js
+++ b/src/js/background/assignManager.js
@@ -1,5 +1,5 @@
 const assignManager = {
-  MENU_REOPEN_IN: "reopen-in-container",
+  MENU_RELOAD_IN: "reopen-in-container",
   MENU_ASSIGN_ID: "open-in-this-container",
   MENU_REMOVE_ID: "remove-open-in-this-container",
   MENU_SEPARATOR_ID: "separator",
@@ -181,7 +181,7 @@ const assignManager = {
   },
 
   async _onClickedHandler(info, tab) {
-    const reopenCookieStoreId = this.menuId2cookieStoreId(info.menuItemId);
+    const reopenCookieStoreId = this.getReloadMenuIdFromCookieStoreId(info.menuItemId);
 
     if (reopenCookieStoreId) {
       const newTab = await browser.tabs.create({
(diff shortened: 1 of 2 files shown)


src/js/background/assignManager.js
```diff
@@ -222,6 +241,16 @@ const assignManager = {
     return backgroundLogic.getUserContextIdFromCookieStoreId(tab.cookieStoreId);
   },
 
+  cookieStoreId2menuId(cookieStoreId) {
```
jonathanKingston: Can you give a better name for the functions here?

src/js/background/assignManager.js
```diff
@@ -294,14 +323,51 @@ const assignManager = {
...
+      });
+    }).catch(() => {});
+    browser.contextMenus.remove(this.MENU_REOPEN_IN);
     browser.contextMenus.remove(this.MENU_REMOVE_ID);
     browser.contextMenus.remove(this.MENU_SEPARATOR_ID);
     browser.contextMenus.remove(this.MENU_HIDE_ID);
     browser.contextMenus.remove(this.MENU_MOVE_ID);
   },
 
+  addContainerMenuEntry(contextualIdentity, contexts) {
```
jonathanKingston: This is specific to the reloading menu so the function should have the reload name in it.

src/js/background/assignManager.js
```diff
@@ -294,14 +323,51 @@ const assignManager = {
...
+    browser.contextMenus.create({
+      id: this.cookieStoreId2menuId(contextualIdentity.cookieStoreId),
+      title: contextualIdentity.name,
+      // TODO: colorized icons?
+      icons: {
+        "16": contextualIdentity.iconUrl
+      },
+      // TODO: hide entry for current container in context menu of tabs
+      contexts: contexts,
```
jonathanKingston: Can the calling function just not call this function when the context id matches?

src/js/background/messageHandler.js
```diff
@@ -76,6 +76,14 @@ const messageHandler = {
...
         const userContextId = backgroundLogic.getUserContextIdFromCookieStoreId(contextualIdentity.cookieStoreId);
         backgroundLogic.deleteContainer(userContextId, true);
+
+        assignManager.removeContainerMenuEntry(contextualIdentity);
+      });
+    }
+
+    if (browser.contextualIdentities.onCreated) {
+      browser.contextualIdentities.onCreated.addListener(({contextualIdentity}) => {
+        assignManager.addContainerMenuEntry(contextualIdentity, ["all", "tab"]);
```
jonathanKingston: It might be simpler to have a `recalculateReloadMenu` function to call instead.

src/js/background/assignManager.js
```diff
@@ -1,4 +1,5 @@
 const assignManager = {
+  MENU_REOPEN_IN: "reopen-in-container",
```
jonathanKingston: When this lands we should file a follow up bug to have a JS file to control context menus rather than have it here.

tunefish:
I have applied the changes you asked for.
Regarding the TODOs:

- After thinking about it again, showing all containers but the one of the tab is not possible for the tab context menu, since the same context menu is shown regardless of the tab which is right-clicked. Removed the TODO from the code
- Tabs can now be opened in default container
- Colorized icons in the context menu are possible if we can manipulate an icon's SVG code. While the icons can be viewed by users by navigating to the correct `resource://` URL, access is blocked when opened from JS with `fetch()`. A solution would be to keep a copy of the icons within the extension. Thoughts?

ScreamingHawk:
Are there still issues preventing this pull request being merged? Happy to help get this across if needed.

jonathanKingston:
This was actually added to the platform, I think we decided to close this right @groovecoder?

groovecoder:
Yes, this has been fixed in the Firefox UI:

screen shot 2018-10-15 at 7 00 06 am

(Nightly 64 2018-10-13)

PR #1141, opened 2018-02-28
Remove identityState when deleting containers

### LINKED ISSUE #1140: IdentityState is not removed from storage when deleting containers
I'm not exactly sure whether not removing identityState from storage is intended, but in case it isn't I've opened PR #1141 to do so.

stoically:
Fixes #1140


stoically:
Some users of Temporary Containers are up to 100k+ opened (and deleted) containers, and hence with 100k entries in MACs storage.js. Would be nice if we could merge this to avoid it in the future.

groovecoder (dismissed):
It looks like #1131 is merged, but this is still red on Travis? I also tried to rebase this branch on latest master and got the same test failures. So I think this needs some updating before we can merge it.

stoically:
Fixed. The mock for `browser.contextualIdentities.remove` was missing. (#1464 introduces `sinon-chrome` to automatically mock all APIs)

liambluebox:
I am one of the  Temporary Containers users. Could this be related to https://github.com/mozilla/multi-account-containers/issues/1675 ?

stoically:
@liambluebox It's not. It generates other data trash. The one line fix is still relevant 4 years in.

liambluebox:
@groovecoder any chance this could be looked at?

dannycolin (approved):
Everything looks good. LGTM.

PR #1190, opened 2018-04-12
Testing 7.0.0

groovecoder:
Would like @SoftVision-PaulOiegas and team to check this alongside [Facebook Container 1.4.0](https://github.com/mozilla/contain-facebook/pull/130).

[firefox_multi-account_containers-7.0.0.zip](https://github.com/mozilla/multi-account-containers/files/1904967/firefox_multi-account_containers-7.0.0.zip)

~~[facebook_container-1.4.0.zip](https://github.com/mozilla/contain-facebook/files/1905022/facebook_container-1.4.0.zip)~~ outdated
[facebook_container-1.4.0.zip](https://github.com/mozilla/contain-facebook/files/1907439/facebook_container-1.4.0.zip)

### DIFF
--- a/src/js/background/assignManager.js
+++ b/src/js/background/assignManager.js
@@ -8,6 +8,7 @@ const assignManager = {
   storageArea: {
     area: browser.storage.local,
     exemptedTabs: {},
+    jailedUrls: [],
 
     getSiteStoreKey(pageUrl) {
       const url = new window.URL(pageUrl);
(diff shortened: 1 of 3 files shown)


jonathanKingston:
@stoically how much of this PR is still relevant. I think we should try and get it rebased and merged too.

stoically:
@jonathanKingston I'd say this is good to merge, for both the domain jailing and tests - went ahead and rebased.

Regarding the jailing: Because of the discussed possibility of a race, where MAC sends the `MACListening` message and FC isn't ready to receive the message yet, it might be an idea to respond to the [message in FC](https://github.com/mozilla/contain-facebook/blob/d6eb8d6b80a975955d7f54fd9793e6e4bcc47d2a/src/background.js#L49-L50) and in MAC check whether we got a response when [sending the message](https://github.com/mozilla/multi-account-containers/pull/1190/commits/14e2b6a36e51fd6b892dc9e1ce4c2aa0224ac0b9#diff-d0a9b69772d58952b3731e2d73d899a7R105), and if not retry resending the message until we get a response. This should also check using `management` if FC is actually installed and enabled.

Before something like that is implemented, we probably should avoid [removing the `getMACAssignment`](https://github.com/mozilla/contain-facebook/issues/303) from FC, to avoid ending up in an unknown state where e.g. tabs are duplicated. Handling the race isn't needed with the assignment check still in place, since it functions as a fallback.

stoically:
Moved the test only changes into #1464

PR #1215, opened 2018-05-21
Add Reopen in Container context menu for tab

### LINKED ISSUE #942: Open Current URL in Container X

arai-a:
Simple port of https://bugzilla.mozilla.org/show_bug.cgi?id=1376119 patch to extension.
which fixes #942.

This patch adds "Reopen in Container" menu into tab menu, which reopens the current tab into the new foreground tab with the specified container.
(doesn't touch/remove the current tab)

the difference between this and the patch above:
 * the menu item is shown at the bottom, instead of under "Duplicate Tab"
 * the container icon's color is not reflected (since WebExtensions' menu doesn't support it)

also, currently this patch doesn't have test.

### DIFF
--- a/src/js/background/reopen.js
+++ b/src/js/background/reopen.js
@@ -1,0 +1,69 @@
+const reopenIn = {
+  // Map from menuItemId to cookieStoreId.
+  cookieStoreIds: new Map(),
+
+  init() {
+    browser.menus.onShown.addListener(async (info, tab) => {
+      if (info.contexts.length !== 1) {
+        return;
+      }
+      if (info.contexts[0] !== "tab") {
+        return;
+      }
+
+      await this.rebuildMenu(tab);
+    });
+
+    browser.menus.onClicked.addListener((info, tab) => {
+      browser.tabs.create({
+        url: tab.url,
+        index: tab.index + 1,
+        cookieStoreId: this.cookieStoreIds.get(info.menuItemId)
+      });
+    });
+  },
+
+  async rebuildMenu(tab) {
+    browser.menus.removeAll();
+
+    const containers = await browser.contextualIdentities.query({});
+
+    const folderId = browser.menus.create({
+      title: "Reopen in Container",
+      contexts: ["tab"],
+    });
+
+    if (tab.cookieStoreId !== "firefox-default") {
+      const menuItemId = "openin-firefox-default";
+      this.cookieStoreIds.set(menuItemId, "firefox-default");
+      browser.menus.create({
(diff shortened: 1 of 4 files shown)


src/js/background/reopen.js
```diff
@@ -0,0 +1,69 @@
...
+      if (info.contexts[0] !== "tab") {
+        return;
+      }
+
+      await this.rebuildMenu(tab);
+    });
+
+    browser.menus.onClicked.addListener((info, tab) => {
+      browser.tabs.create({
+        url: tab.url,
```
Qwertovsky: New tab url is 'about:newtab'. Does not work. I think it needs to be replaced by NULL.
> Error: Illegal URL: about:newtab

arai-a: Indeed, thanks :)
but I think opening other open-able tab is not something we should do, since it contradicts with the menu label.

if we implement this as WebExtension feature, we shouldn't show the tab menu for tabs which WebExtensions cannot reopen.
(about:*, data:*, and maybe chrome:* etc?)

Qwertovsky: > we shouldn't show the tab menu for tabs which WebExtensions cannot reopen.

I agree.

arai-a:
This feature is going to be implemented in Firefox itself.

Issue #972, opened 2017-11-16
Refactor to async arrow functions


groovecoder:
See https://github.com/mozilla/multi-account-containers/pull/921#issuecomment-341670787 for details.

Soundharyaam:
Hi, Can I work on this bug?

groovecoder:
Yup.

Soundharyaam:
Thanks!

Soundharyaam:
PR: #1263

Soundharyaam:
@armenzg I have made the PR. Could you please review this?

Soundharyaam:
@groovecoder  I have made the PR. Could you please provide your review on this?

groovecoder:
See https://github.com/mozilla/multi-account-containers/pull/1263#issuecomment-423744345. This is a big refactor - not just throwing `async` onto a single function.

Paarmita:
@groovecoder So we need to assign arrow functions to all async functions in the code base?
And Can I work on this?

groovecoder:
@Paarmita yes - we can clean up some code (i.e., make it less verbose) by using `async (...) => { ... }` in many places.

jonathanKingston:
My intent on the previous bug was only to use `async () => {}` when it's something that is already inline, not changing named methods to be const async arrows.

This is less about changing just the code style in a blanket fashion and instead moving functions to be inlined whenever they are only used once.

The work might actually be check and see if there is any clean up that makes the code less verbose using async arrows. The answer might be it only makes it more verbose. IMO the latest PR makes it more verbose and less idiomatic.

When I suggested the use of these I believe we had small one line functions that are used only once.
```
async a(b) {
  await doThing();
},
async click() {
  for b in x {
    callback(await a(b));
  }
}
```

Which could be changed to something along the lines of:
```
async click() {
  for b in x {
    callback(await a(b) {
      await doThing();
    });
  }
}
```

Paarmita:
@jonathanKingston I changed the async function to arrow functions which makes it more verbose, I can work on the suggested lines as well. Do I need to change arrow functions for which I opened the PR?

ctgardner:
I understand the crux of this issue isn't to simply replace functions declared using the `function` keyword with arrow-functions, but to inline named async functions where sensible. I've reviewed all named async functions and I believe none would benefit from being inline, with the exception of a few handlers registered using `addListener`. Given the past confusion surrounding this issue, I believe this issue should be closed.

PR #1263, opened 2018-09-20
Refactor to async arrow functions #972

Soundharyaam:
deleteContainer() refactored to arrow function.


groovecoder:
#972 is a much bigger refactor than throwing `async` onto a single function.

#972 refers to using `async` arrow functions across the code-base. I.e., there are some places in the code where we defined a named `async` function to pass as an argument, where we could instead simplify the code by moving it an `async (...) => { ... }` inline function.

PR #1265, opened 2018-09-21
#1028 resolves focus name field on opening new container sub-panel

### LINKED ISSUE #1028: Focus name field on opening New Container sub-panel
In the sub-panel that appears on hitting the "+" (Create New Container) button: 
 * The `Name` field should be automatically focused
 * The default `Container #xx` text in that field should be selected so that typing removes it. (Alternative solution: Have no default text. I don't know why anyone would use the `Container #xx` scheme when naming containers, but maybe there's a reason for it?)

This is a continual "missing stair" issue for me: every time I create a new container, I start typing because I expect that field to be focused. I'm assuming that I'm not the only person this happens to.

ShivangiKakkar:
### AFTER:
webp net-gifmaker 2

### DESCRIPTION:
- Following the thread of the issue, I found the discussion is absolutely correct about the `focus()`

- After searching for hours the reason I found was interesting and weird; The solution was to "pause" the JavaScript execution to let the rendering threads catch up. And this is the effect that `setTimeout()` with a timeout of 0 does. For better understanding, you can refer this link [here](https://stackoverflow.com/questions/779379/why-is-settimeoutfn-0-sometimes-useful)
- For even better understanding, answer your curiosity by watching [this](https://www.youtube.com/watch?v=8aGhZQkoFbQ) 😄.
- The `select()` and `focus()` both work fine but `focus()` seems to be hidden by the input box-shadow.

### DIFF
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -1001,6 +1001,12 @@ Logic.registerPanel(P_CONTAINER_EDIT, {
 
     document.querySelector("#edit-container-panel-name-input").value = identity.name || "";
     document.querySelector("#edit-container-panel-usercontext-input").value = userContextId || NEW_CONTAINER_ID;
+    const containerName = document.querySelector("#edit-container-panel-name-input");
+    setTimeout(function()
+    { 
+      containerName.focus(); 
+      containerName.select();  
+    }, 0);
     [...document.querySelectorAll("[name='container-color']")].forEach(colorInput => {
       colorInput.checked = colorInput.value === identity.color;
     });
@@ -1015,7 +1021,6 @@ Logic.registerPanel(P_CONTAINER_EDIT, {
 
 // P_CONTAINER_DELETE: Delete a container.
 // ----------------------------------------------------------------------------
-
 Logic.registerPanel(P_CONTAINER_DELETE, {
   panelSelector: "#delete-container-panel",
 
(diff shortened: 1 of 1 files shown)


ShivangiKakkar:
@groovecoder Hope this works fine, Please review. :)

ShivangiKakkar:
@groovecoder I faced a problem of randomly failing tests while testing. I hope it is solved [here](https://github.com/mozilla/multi-account-containers/pull/1190). :)

jonathanKingston:
Does https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame fix the problem also? My understanding is it should and it explains better to what the intent is here.

If not it should probably have a code comment why this is here. I understand about the event loop but haven't really checked why it's needed here.

ShivangiKakkar:
@jonathanKingston Umm, I really didn't check that. I'll try and look into the API you pointed and let you know if it solves our issue here.  🙂
I just googled up and found a thread where using event loop solved a similar problem, tried it and I found it too work fine. Yes now that I think if its really needed here, I am not sure. 😅
Need some time to dig more into the issue.
Thanks!

ShivangiKakkar:
@jonathanKingston Yes! you were right `requestAnimationFrame()` seems to work fine. Please review.

ShivangiKakkar:
> Does https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame fix the problem also? My understanding is it should and it explains better to what the intent is here.
> 
> If not it should probably have a code comment why this is here. I understand about the event loop but haven't really checked why it's needed here.

@jonathanKingston Yes! you were right `requestAnimationFrame()` seems to work fine. Thanks :)

ShivangiKakkar:
@jonathanKingston Thanks for guiding me :)

PR #1270, opened 2018-09-24
#168, added ability to close a tab individually

### LINKED ISSUE #168: Add ability to close a tab without showing the hidden container.
When the user hides tabs, there is not a way to close all of them (or some of them) without showing them  all again.

ShivangiKakkar:
### Description
- Implemented individual closing (or removing) of tabs in active/hidden state as described in the issue discussion. Not sure if this is the best way to implement this enhancement, so any kind of reviews are welcome. :)
- The UX implemented is mentioned [here](https://github.com/mozilla/multi-account-containers/issues/168#issuecomment-279675783).

#### Problems/Issues faced: 
- @groovecoder @bakulf @jonathanKingston Everything seems to work fine, except one thing i.e. when the container `hasHiddenTabs` then calling the `browser.tabs.remove()` method doesn't seems to work, thus not closing the tabs in hidden state. (It closes the tabs perfectly fine when they are not hidden)
So can you please have look and let me know what am I missing here? Thank you!

#### TODO:
- Make closing of tabs in hidden state, working.


src/js/popup.js
```diff
@@ -777,9 +778,29 @@ Logic.registerPanel(P_CONTAINER_INFO, {
...
+        <td><img src="/img/container-close-tab.svg" style="width:50%;" id="close-tab" class="container-close-tab clickable" /></td>`;
       tr.querySelector("td").appendChild(Utils.createFavIconElement(tab.favIconUrl));
 
+      document.getElementById("container-info-table").appendChild(fragment);
+
+      const closeTab = document.querySelector("#close-tab");
+      closeTab.setAttribute("id", tab.id);
+
+      if(identity.hasHiddenTabs) {
+        closeTab.addEventListener("mouseover",function() {
```
jonathanKingston: `closeTab.addEventListener("mouseover",function() {`

Should likely be:
`closeTab.addEventListener("mouseover", () => {`

ShivangiKakkar: @jonathanKingston I made this change but it didn't solve the above problem of tabs not closing while they are in the hidden state.

ShivangiKakkar:
@jonathanKingston Would you please guide me a bit more into this. Thanks :)

jonathanKingston:
@ShivangiKakkar perhaps the quickest solution is to hide this behaviour on hidden tabs.

The hidden tabs aren't truely hidden, they are stored in background.js as a url only. Rather than the newer hidden tab implementation we have in the browser. So if we were to delete these we would need to call into that and then rerender the list after that.

The current logic ignores clicking on these anyway: `if (!tab.hiddenState) {` which maybe is worthwhile here for the time being. (I personally think it would be great to just restore one tab too)

jonathanKingston:
TL;DR: just implement this for visible tabs and then file follow up bugs for:

- Considering a real implementation of using hidden tabs
- Showing hidden tabs one at once
- Removing a hidden tab

Thank you.

ShivangiKakkar:
@jonathanKingston - Just implemented this just for visible tabs and will file follow up bugs soon.

webp net-gifmaker

src/js/popup.js
```diff
@@ -777,9 +777,20 @@ Logic.registerPanel(P_CONTAINER_INFO, {
       tr.classList.add("container-info-tab-row");
       tr.innerHTML = escaped`
         <td></td>
-        <td class="container-info-tab-title truncate-text" title="${tab.url}" >${tab.title}</td>`;
+        <td class="container-info-tab-title truncate-text" title="${tab.url}" >${tab.title}</td>
+        <td><img src="/img/container-close-tab.svg" style="width:50%;" id="close-tab" class="container-close-tab" /></td>`;
```
jonathanKingston: Please put all styles in the CSS rather than in the HTML. I think it would be better if the X is only visible on hover that way the title can be longer until the user hovers.

src/img/container-close-tab.svg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 21.0.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
```
jonathanKingston: Remove anything that isn't needed here like the adobe comment and also `<g` tags aren't needed. For a trimmed down svg look at src/img/blank-tab.svg

src/img/container-close-tab.svg
```diff
@@ -0,0 +1,10 @@
...
+	 viewBox="0 0 7 7" style="enable-background:new 0 0 7 7;" xml:space="preserve">
+<g id="Symbols">
+	<g id="close-glyph">
+		<polygon style="fill:#4C4C4C;" points="5.8,0 3.5,2.4 1.2,0 0,1.2 2.4,3.5 0.1,5.8 1.2,7 3.5,4.7 5.8,7 7,5.8 4.7,3.5 7,1.2 		"/>
```
jonathanKingston: Again we can remove `style=` here too:
```suggestion
		<polygon fill="#4c4c4c" points="5.8,0 3.5,2.4 1.2,0 0,1.2 2.4,3.5 0.1,5.8 1.2,7 3.5,4.7 5.8,7 7,5.8 4.7,3.5 7,1.2 		"/>
```

src/js/popup.js
```diff
@@ -777,9 +777,20 @@ Logic.registerPanel(P_CONTAINER_INFO, {
...
       tr.querySelector("td").appendChild(Utils.createFavIconElement(tab.favIconUrl));
 
+      document.getElementById("container-info-table").appendChild(fragment);
+
+      const closeTab = document.querySelector("#close-tab");
+      closeTab.setAttribute("id", tab.id);
```
jonathanKingston: Could this be in the template above instead? `id="{tab.id}"` ?

src/js/popup.js
```diff
@@ -777,9 +777,20 @@ Logic.registerPanel(P_CONTAINER_INFO, {
...
+
+      Logic.addEnterHandler(closeTab, async function(e) {
```
jonathanKingston: Can you move this logic inside line 759 rather than rely on CSS to prevent the click.

Most/All of the logic of this addon is within the JS and the current logic is already dictated by if it's `!tab.hiddenState`. By moving some of the logic into the CSS it will make it harder to maintain as developers will have to look in both places.

Also we should probably not even add the button / remove it if the row is a hidden tab. Making it `opacity: 0;` probably only makes it harder to click / it doesn't have a `tab.id` so does nothing perhaps?

src/css/popup.css
```diff
@@ -670,6 +670,23 @@ span ~ .panel-header-text {
...
+
+.container-info-tab-row:hover .container-close-tab {
+  opacity: 0.5;
+}
+
+.container-info-tab-row .container-close-tab:hover {
+  opacity: 1;
+}
+
+.container-info-tab-row:not(.clickable) .container-close-tab,
```
jonathanKingston: As mentioned elsewhere I reckon this `:not(.clickable)` should be removed.

src/js/popup.js
ShivangiKakkar: The only reason of putting inline CSS here is because width property seems to be blacklisted by lint, any other way to resize the image? And, yes X is only visible when someone hovers over the row.

ShivangiKakkar:
@jonathanKingston Review changes are done. Please review again. Thanks! :)

src/js/popup.js
```diff
@@ -777,20 +777,30 @@ Logic.registerPanel(P_CONTAINER_INFO, {
       tr.classList.add("container-info-tab-row");
       tr.innerHTML = escaped`
         <td></td>
-        <td class="container-info-tab-title truncate-text" title="${tab.url}" >${tab.title}</td>`;
+        <td class="container-info-tab-title truncate-text" title="${tab.url}" >${tab.title}</td>
+        <td><img src="/img/container-close-tab.svg" width="50%" id="${tab.id}" class="container-close-tab" /></td>`;
```
jonathanKingston: Can we move this into the title row instead of it being it's own cell? The idea is that it would be `display: none;` until the user mouse overs that row, then it becomes visible. This will allow more of the title to be visible as we are pretty limited for space here.

jonathanKingston: Also remove `width=50%` and plae all styles into the css file.

jonathanKingston: Rather than the logic to delete it and needing to use the `id=""` instead you could construct the button string before like this:
```javascript
let closeButton = "";
if (!tab.hiddenState) {
  closeButton = `<img src="/img/container-close-tab.svg" class="container-close-tab" />`;
}
tr.innerHTML = escaped`<td></td><td class="container-info-tab-title truncate-text" title="${tab.url}" >${tab.title}${closeButton}</td>`;
```

ShivangiKakkar:
@jonathanKingston In reference to your last review: If I use the `closeButton` as you have suggested it takes as a string than HTML, even after following the template literal rules, probably because we are using it between the `<td>` tags?
Thus, I used a different coloum for `closeButton`, plus I cant think off if it'd make much difference in the space even if move closeButton to a new row, can you please explain that? And, also using a diff row for `closeButton` seems to make `inline-size:50%` instead of `width:50%` work as @groovecoder pointed out.

Last thing, we need the logic i.e setting the closeTabs `id=${tab.id}` as to to use it when event handler is clicked and `browser.tabs.remove()` is executed, how else can we do that? 
Thank you and sorry if I am not clear somewhere.

groovecoder:
The current UI on this branch looks good to me - it doesn't seem to use unnecessary space.

Is that the only change left to make?

mac-close-tab-hover

ShivangiKakkar:
> The current UI on this branch looks good to me - it doesn't seem to use unnecessary space.
> 
> Is that the only change left to make?
> 
> mac-close-tab-hover

Yes! Actually, I am able to add the image block dynamically as @jonathanKingston suggested, But I am facing following problems in applying that:
-Tabs with a shorter title are doing fine but they seem messy as I hover and they add that `image icon block`. 
-The tabs with a longer title are not able to show the `image icon block`, I checked the console it is added but after the text ends. So, I tried to change some code in css with `truncate-text` to  make it appear but I failed.

ShivangiKakkar:
@groovecoder Yes! Actually, I am able to add the image block dynamically as @jonathanKingston suggested, But I am facing following problems in applying that:
-Tabs with a shorter title are doing fine but they seem messy as I hover and they add that image icon block.
-The tabs with a longer title are not able to show the image icon block, I checked the console it is added but after the text ends. So, I tried to change some code in css with truncate-text to make it appear but I failed.

ShivangiKakkar:
@groovecoder @jonathanKingston Finally did all the changes. Let me know your views. :)
final

ShivangiKakkar:
@jonathanKingston Conflicts resolved with the changes we made today. Hopefully, everything seems fine. Please have a final look.

jonathanKingston:
Thanks @ShivangiKakkar well done!

ShivangiKakkar:
Thanks a lot for supporting throughout! :) @jonathanKingston

PR #1274, opened 2018-09-27
Refactor async arrow functions

### LINKED ISSUE #972: Refactor to async arrow functions

Paarmita:
Refractor all async function to arrow functions 
Fixed #972

### DIFF
--- a/src/js/background/assignManager.js
+++ b/src/js/background/assignManager.js
@@ -69,12 +69,12 @@ const assignManager = {
       return this.area.remove([siteStoreKey]);
     },
 
-    async deleteContainer(userContextId) {
+    const deleteContainer = async (userContextId) => {
       const sitesByContainer = await this.getByContainer(userContextId);
       this.area.remove(Object.keys(sitesByContainer));
     },
 
-    async getByContainer(userContextId) {
+    const getByContainer = async (userContextId) => {
       const sites = {};
       const siteConfigs = await this.area.get();
       Object.keys(siteConfigs).forEach((key) => {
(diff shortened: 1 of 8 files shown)


Paarmita:
@groovecoder Please review the PR :)

ShivangiKakkar:
@jonathanKingston @groovecoder cc @Paarmita 
I went through the above PR and the travis build shows the syntax errors for the changes in all the files except in `popup.js` and `content-script.js`
In the other files, I guess the arrow function syntax isn't supported because the async functions are defined inside an object which is itself initialized in ES5 syntax.
What is your take on this?

groovecoder:
The intent of https://github.com/mozilla/multi-account-containers/issues/972 and https://github.com/mozilla/multi-account-containers/pull/921#issuecomment-341670787 (like @jonathanKingston [says](https://github.com/mozilla/multi-account-containers/pull/921#issuecomment-425271223)) is **NOT** to change all named methods and functions into defined async arrow functions.

Then intention is to find async functions that are only ever used once, and move them into inline async arrow functions.

E.g., `assignManager` has an `_onClickedHandler` method. But that logic is only used in one place - during `init()`. The only reason we made `_onClickedHandler` a method is because, when we wrote it, JS didn't support using `async`/`await` in inline `=>` functions.

Now that JS supports `async () => {}` functions, we could move the `_onClickedHandler` logic from its own named function into an inline `async (info, tab) => {` ... `}` function in `init`.

Does that make sense?

With all that said, we may just close this particular PR and instead clean up unnecessary defined `async` functions as we make changes.

jonathanKingston:
I think this makes sense to close for now. Thanks for the submission.

PR #1283, opened 2018-10-07
list sites which always load with container

jhford:
Thanks for building this addon!  It's really great and I really love the
functionality it provides.  I have a small change that I'd like to make.
It's motivated by my personal frustration with trying to figure out what
is supposed to load in a specific container.

The data storage system used is optimized for quick lookups on webpage
load, which is the right choice.  Instead of changing that, I decided
instead to iterate through the full data storage and pick out the
relevant items and trace them back to the container that they belong to
for display.  I suspect that with the size of the data storage, it's
probably fine.  Other than building a mapping in a different key of the
local storage when adding a site to a container, I don't see any way to
make this quicker on lookup.

I'm not at all an extension developer, nor do I write a lot of html or
css.  I probably butchered the display portion of this, but it mostly
works...  I would not expect this patch to merge in this state, but I'd
be interested in working on making it possible to merge.  Maybe hiding
this functionality behind an addon preference?

If this functionality isn't desirable in the official addon, is there
anything which would block me from submitting a fork to
addons.mozilla.org?  It's primarily because I'd like to use it on a
release copy of firefox, and so I'd need it signed by addons.

### DIFF
--- a/src/css/popup.css
+++ b/src/css/popup.css
@@ -685,6 +685,16 @@ span ~ .panel-header-text {
   padding-inline-start: 16px;
 }
 
+ul.container-info-url-list {
+  font-size: 14px;
+  list-style-type: none;
+  padding-left: 16px;
+}
+
+li.container-info-url-item {
+  padding-left: 16px;
+}
+
 .container-info-has-tabs img,
 .container-info-tab-row img {
   block-size: 16px;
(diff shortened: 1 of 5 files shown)


groovecoder:
Thank you for your previous contributions to Multi-Account Containers. We're closing PRs with conflicts to prepare this repo for some work in the near future. If you would like to re-submit the changes in a branch from the latest `master` code, we would love to take another look at this!

PR #1284, opened 2018-10-11
Adds proxy support as per #313

samuel-crypto:
Dear Maintainers!
Thank you very much for being extremely helpful and friendly regarding this request.

This pull request adds support for HTTP proxies as per issue #313 
I am submitting the raw working (tested) code, and am hoping it works for everybody and does not break any already existing features.

Some tests seem to fail and I am not quite sure why (tests seemingly unrelated whatsoever with my additions). Personally I didn't really change any logic of the add-on (except additions where necessary) and would appreciate help where possible to fix these!

Thanks again,
I sincerely hope this will eventually get merged.
Samuel.

### DIFF
--- a/src/js/background/assignManager.js
+++ b/src/js/background/assignManager.js
@@ -119,7 +119,7 @@ const assignManager = {
 
   async handleProxifiedRequest(requestInfo) {
     if(requestInfo.tabId === -1)
-      return {type: "direct"};
+      return Utils.DEFAULT_PROXY;
 
     const tab = await browser.tabs.get(requestInfo.tabId);
     const proxy = await proxifiedContainers.retrieveFromBackground(tab.cookieStoreId);
--- a/src/js/background/assignManager.js
+++ b/src/js/background/assignManager.js
@@ -118,8 +118,10 @@ const assignManager = {
   },
 
   async handleProxifiedRequest(requestInfo) {
+
+    // The following blocks potentially dangerous requests for privacy that come without a tabId
     if(requestInfo.tabId === -1)
-      return Utils.DEFAULT_PROXY;
+      return Utils.getBogusProxy();
 
     const tab = await browser.tabs.get(requestInfo.tabId);
     const proxy = await proxifiedContainers.retrieveFromBackground(tab.cookieStoreId);
--- a/src/js/background/assignManager.js
+++ b/src/js/background/assignManager.js
@@ -122,7 +122,7 @@ const assignManager = {
       return {type: "direct"};
 
     const tab = await browser.tabs.get(requestInfo.tabId);
-    const proxy = await window.proxifiedContainers.retrieveFromBackground(tab.cookieStoreId);
+    const proxy = await proxifiedContainers.retrieveFromBackground(tab.cookieStoreId);
 
     return proxy;
   },
--- a/src/js/background/assignManager.js
+++ b/src/js/background/assignManager.js
@@ -117,6 +117,16 @@ const assignManager = {
     return true;
   },
 
+  async handleProxifiedRequest(requestInfo) {
+    if(requestInfo.tabId === -1)
+      return {type: "direct"};
+
+    var tab = await browser.tabs.get(requestInfo.tabId);
+    var proxy = await window.proxifiedContainers.retrieveFromBackground(tab.cookieStoreId);
+
+    return proxy;
+  },
+
   // Before a request is handled by the browser we decide if we should route through a different container
   async onBeforeRequest(options) {
     if (options.frameId !== 0 || options.tabId === -1) {
@@ -223,6 +233,10 @@ const assignManager = {
     browser.contextMenus.onClicked.addListener((info, tab) => {
       this._onClickedHandler(info, tab);
     });
+
+    // Before anything happens we decide if the request should be proxified
+    browser.extension.getBackgroundPage().console.log('[SAMUEL CODE] Adding proxy.onRequest listener');
+    browser.proxy.onRequest.addListener(this.handleProxifiedRequest, {urls: ["<all_urls>"]});
 
     // Before a request is handled by the browser we decide if we should route through a different container
     this.canceledRequests = {};
--- a/src/js/background/assignManager.js
+++ b/src/js/background/assignManager.js
@@ -121,8 +121,8 @@ const assignManager = {
     if(requestInfo.tabId === -1)
       return {type: "direct"};
 
-    var tab = await browser.tabs.get(requestInfo.tabId);
-    var proxy = await window.proxifiedContainers.retrieveFromBackground(tab.cookieStoreId);
+    const tab = await browser.tabs.get(requestInfo.tabId);
+    const proxy = await window.proxifiedContainers.retrieveFromBackground(tab.cookieStoreId);
 
     return proxy;
   },
@@ -235,7 +235,6 @@ const assignManager = {
     });
 
     // Before anything happens we decide if the request should be proxified
-    browser.extension.getBackgroundPage().console.log('[SAMUEL CODE] Adding proxy.onRequest listener');
     browser.proxy.onRequest.addListener(this.handleProxifiedRequest, {urls: ["<all_urls>"]});
 
     // Before a request is handled by the browser we decide if we should route through a different container
--- a/src/js/background/backgroundLogic.js
+++ b/src/js/background/backgroundLogic.js
@@ -32,6 +32,10 @@ const backgroundLogic = {
       await browser.contextualIdentities.remove(this.cookieStoreId(userContextId));
     }
     assignManager.deleteContainer(userContextId);
+
+    // Now remove the identity->proxy association in proxifiedContainers also
+    proxifiedContainers.delete(this.cookieStoreId(userContextId));
+
     return {done: true, userContextId};
   },
 
--- a/src/js/background/backgroundLogic.js
+++ b/src/js/background/backgroundLogic.js
@@ -47,11 +47,11 @@ const backgroundLogic = {
     } else {
       donePromise = browser.contextualIdentities.create(options.params);
 
-      //We cannot yet access the new cookieStoreId via this.cookieStoreId(...), so we take this from the resolved promise
+      // We cannot yet access the new cookieStoreId via this.cookieStoreId(...), so we take this from the resolved promise
       donePromise.then((identity) => {
         window.proxifiedContainers.set(identity.cookieStoreId, options.proxy);
       }).catch(() => {
-        //Empty because this should never happen theoretically.
+        // Empty because this should never happen theoretically.
       });
     }
     await donePromise;
@@ -165,7 +165,7 @@ const backgroundLogic = {
         index: -1
       });
     } else {
-      //As we get a blank tab here we will need to await the tabs creation
+      // As we get a blank tab here we will need to await the tabs creation
       newWindowObj = await browser.windows.create({
       });
       hiddenDefaultTabToClose = true;
--- a/src/js/background/backgroundLogic.js
+++ b/src/js/background/backgroundLogic.js
@@ -46,6 +46,11 @@ const backgroundLogic = {
       window.proxifiedContainers.set(this.cookieStoreId(options.userContextId), options.proxy);
     } else {
       donePromise = browser.contextualIdentities.create(options.params);
+
+      //We cannot yet access the new cookieStoreId via this.cookieStoreId(...), so we take this from the resolved promise
+      donePromise.then((identity) => {
+        window.proxifiedContainers.set(identity.cookieStoreId, options.proxy);
+      });
     }
     await donePromise;
     browser.runtime.sendMessage({
--- a/src/js/background/backgroundLogic.js
+++ b/src/js/background/backgroundLogic.js
@@ -50,6 +50,8 @@ const backgroundLogic = {
       //We cannot yet access the new cookieStoreId via this.cookieStoreId(...), so we take this from the resolved promise
       donePromise.then((identity) => {
         window.proxifiedContainers.set(identity.cookieStoreId, options.proxy);
+      }).catch(() => {
+        //Empty because this should never happen theoretically.
       });
     }
     await donePromise;
--- a/src/js/background/backgroundLogic.js
+++ b/src/js/background/backgroundLogic.js
@@ -42,6 +42,8 @@ const backgroundLogic = {
         this.cookieStoreId(options.userContextId),
         options.params
       );
+
+      window.proxifiedContainers.set(this.cookieStoreId(options.userContextId), options.proxy);
     } else {
       donePromise = browser.contextualIdentities.create(options.params);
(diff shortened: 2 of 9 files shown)


src/js/proxified-containers.js
```diff
@@ -0,0 +1,148 @@
...
+      }).catch((error) => {
+        reject(error);
+      });
+    });
+  },
+
+  report_proxy_error: function(error, identifier = null) {
+    //Currently I print to console but this is inefficient
+    const relevant_id_str = identifier === null ? "" : " call supplied with id: " + identifier.toString() + " ";
+    browser.extension.getBackgroundPage().console.log("proxifiedContainers error occured" + relevant_id_str + ": " + JSON.stringify(error));
```
jonathanKingston: Can you use a template string here?
```
browser.extension.getBackgroundPage().console.log(`proxifiedContainers error occured ${relevant_id_str}: ${JSON.stringify(error)}`);
```

And the same for the above too:
```
` call supplied with id: ${identifier} `
```

jonathanKingston:
It doesn't look like the code is clearing up the storage when the containers are cleared. Is that the case?

Also we need to verify that all requests in the container are either blocked or go through the proxy settings. Things like service workers don't have a tab, this is something that we need to check works or it's a privacy leak.

samuel-crypto:
This is certainly a big concern which I also had.
Currently whenever the proxy doesn't work this error is simply displayed by
Firefox itself with its 'Proxy server not responding' error page, so there
is no re-routing to no proxy (which would have been disasterous), however
blocking of non-proxified requests is certainly necessary (but could it
possibly break some functions?). I will also confirm what exactly can be
proxified.

samuel-crypto:
> It doesn't look like the code is clearing up the storage when the containers are cleared. Is that the case?
> 
> Also we need to verify that all requests in the container are either blocked or go through the proxy settings. Things like service workers don't have a tab, this is something that we need to check works or it's a privacy leak.

Thank you very much for your response, I really appreciate the time taken for all these reviews.

Indeed this is the case, I'm not sure how I allowed this design flaw to happen. Possibly we can have a scan for orphan containers when the containers are listed in the 'edit containers' popup (or at startup / some other place), and deletion of them afterwards? This can probably be implemented as a function in `proxifiedContainers`.

Regarding blocking and proxies - this is certainly a big concern which I also had. 

Currently whenever the proxy doesn't work this error is simply displayed by Firefox itself with its 'Proxy server not responding' error page, so there is no re-routing to no proxy (which would have been disasterous), however, blocking of non-proxified requests is certainly necessary (but could it possibly break some functions?). 

The possibility of proxifying these background/service/non-page requests may exist but then an obvious problem resurfaces - what if you are in a tab with a container with different proxy settings when the background/service code requests something than you were in a minute ago? Since this would confuse other code (and probably give very mystical and interesting errors and bugs) blocking indeed might be the only option.

src/js/background/backgroundLogic.js
```diff
@@ -41,8 +41,17 @@ const backgroundLogic = {
         this.cookieStoreId(options.userContextId),
         options.params
       );
+
+      window.proxifiedContainers.set(this.cookieStoreId(options.userContextId), options.proxy);
     } else {
       donePromise = browser.contextualIdentities.create(options.params);
+
+      //We cannot yet access the new cookieStoreId via this.cookieStoreId(...), so we take this from the resolved promise
```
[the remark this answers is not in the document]
samuel-crypto: Oops sorry :)

samuel-crypto:
I have discovered another issue - according to [Mozilla Docs]() the `username` of `ProxyInfo` isn't actually usable as authentication for HTTP proxies - instead "For HTTP proxy authorizations, use webRequest.onAuthRequired". Will have to refactor relevant parts

src/js/proxified-containers.js
```diff
@@ -0,0 +1,148 @@
...
+    configurable: true,
+    writable: true
+  });
+
+
+// This object allows other scripts to access the list mapping containers to their proxies
+proxifiedContainers = {
+
+  // Slightly modified version of 'retrieve' which returns a direct proxy whenever an error is met.
+  retrieveFromBackground: function(cookieStoreId = null) {
```
jonathanKingston: `retrieveFromBackground: function(` can be `retrieveFromBackground(` here :)

src/js/proxified-containers.js
```diff
@@ -0,0 +1,148 @@
...
+    return new Promise((resolve, reject) => {
+      proxifiedContainers.retrieve(cookieStoreId).then((success) => {
+        resolve(success.proxy);
+      }, function() {
+        resolve({
+          type: "direct"
```
jonathanKingston: Another case where the constant could be used:
```suggestion
          type: Utils.DEFAULT_PROXY
```

src/js/background/assignManager.js
```diff
@@ -113,6 +113,16 @@ const assignManager = {
     return true;
   },
 
+  async handleProxifiedRequest(requestInfo) {
+    if(requestInfo.tabId === -1)
+      return {type: "direct"};
```
jonathanKingston: Move this constant into Utils and use it everywhere, even in comparisons I reckon.
```suggestion
      return Utils.DEFAULT_PROXY;
```

src/js/popup.js
```diff
@@ -1016,18 +1004,44 @@ Logic.registerPanel(P_CONTAINER_EDIT, {
...
     [...document.querySelectorAll("[name='container-icon']")].forEach(iconInput => {
       iconInput.checked = iconInput.value === identity.icon;
     });
 
+    // Clear the proxy field before doing the retrieval requests below
+    document.querySelector("#edit-container-panel-proxy").value = "";
+
+    const edit_proxy_dom = function(result) {
+      if(result.type === "http")
+        document.querySelector("#edit-container-panel-proxy").value = result.host.toString() + ":" + result.port.toString();
```
jonathanKingston: I suspect this works the same and is a little more readable:
```suggestion
        document.querySelector("#edit-container-panel-proxy").value = `${result.host}:${result.port}`;
```

src/js/popup.js
```diff
@@ -1016,18 +1004,44 @@ Logic.registerPanel(P_CONTAINER_EDIT, {
...
+
+
+    proxifiedContainers.retrieve(identity.cookieStoreId).then((result) => {
+      edit_proxy_dom(result.proxy);
+    }, (error) => {
+      if(error.error === "uninitialized" || error.error === "doesnotexist") {
+        proxifiedContainers.set(identity.cookieStoreId, DEFAULT_PROXY, error.error === "uninitialized").then((result) => {
+          edit_proxy_dom(result);
+        }, (error) => {
+          proxifiedContainers.report_proxy_error(error, "popup.js: error 1");
```
jonathanKingston: Make sure you fill in these with useful error names please :)

src/js/proxified-containers.js
```diff
@@ -0,0 +1,148 @@
+// Below lets us print errors, huge thanks to Jonothan @ https://stackoverflow.com/questions/18391212/is-it-not-possible-to-stringify-an-error-using-json-stringify
```
jonathanKingston: *Jonathan

src/js/proxified-containers.js
```diff
@@ -0,0 +1,148 @@
...
+        resolve(proxifiedContainersStore[index]);
+      }, (errorObj) => {
+        reject(errorObj);
+      }).catch((error) => {
+        throw error;
+      });
+    });
+  },
+  parseProxy: function(proxy_str) {
+    const regexp = /(\b(\w+):(\w+)@)?(((?:\d{1,3}\.){3}\d{1,3}\b)|(\b(\w+)(\.(\w+))+))(:(\d+))?/;
```
jonathanKingston: Add a comment explaining what this is parsing so we don't forget please :)

jonathanKingston:
> The possibility of proxifying these background/service/non-page requests may exist but then an obvious problem resurfaces - what if you are in a tab with a container with different proxy settings when the background/service code requests something than you were in a minute ago?

Currently firefox routes requests through their own containerised serviceworker which means that there will be a leak outside of the proxy into a normal request. The problem being is that these requests don't have the container id. I suspect we should try and fix that before landing this code.

Either that or at least add a comment to this effect where you do `if(requestInfo.tabId === -1)` to make it clear and file a follow up bug with the comment linking to the bug.

Thanks for this! I will load it into the browser and do a final review once you have done the bug you mentioned.

src/js/proxified-containers.js
samuel-crypto: Should be resolved now. Thanks

src/js/background/assignManager.js
samuel-crypto: Ok I've fixed it, currently it is implemented using `Object.defineProperty` (as otherwise the constant isn't a real constant and can be deleted and modified, e.t.c.)

samuel-crypto:
> It doesn't look like the code is clearing up the storage when the containers are cleared. Is that the case?
> 
> Also we need to verify that all requests in the container are either blocked or go through the proxy settings. Things like service workers don't have a tab, this is something that we need to check works or it's a privacy leak.

Clean up issue is resolved. Re privacy, with blocking non-page requests it can be tested somehow (I'm not entirely sure what `requestInfo.parentFrameId != -1` does, but you mentioned that the undesirable requests do not have a tab id, so that can be used), and then it must be somehow blocked. For now I'm just going to leave a reference to this problem in a comment.

ecx86:
I hope this gets merged. I'm looking forward to it. I have a question about WebSockets. Will this proxy those connections too?

stoically:
@samuel-crypto The tests go red because the call to `browser.proxy.onRequest.addListener` isn't defined in the [browser-mock](https://github.com/mozilla/multi-account-containers/blob/6f4d3c43274701e8c166f3ed80e4d453860fd31d/test/browser.mock.js#L5). They will go green again when #1190 is merged since it includes a refactoring that replaces the manual browser-mock with [sinon-chrome](https://github.com/acvetkov/sinon-chrome).

fsdjfhadflkjhsd:
Hi Samuel, I just signed up on github, as I found out about your per-tab-proxy allowing plugin. I hope you will release it soon.

xavier83:
Any update on the status of the PR?

samuel-crypto:
> Any update on the status of the PR?

I was (unfortunately) delayed for a long time because of other things but am now dedicating myself to finishing it, so hopefully it should be in soon!

samuel-crypto:
@jonathanKingston - excuse me for the prolonged activity on this PR, one question - I am unsure how exactly to implement the cleanup, as I see it there are a number of options:
 - The best would probably be to somehow hook up to the container delete event, but as I don't know or fully understand the internal structure of the plugin, your help would be appreciated here if possible
 - Passes that check for orphan proxy containers and clean them up with certain regularity can be organized (and this is much simpler coding wise, but generally a bit messier)

If I go with option 2 the other issue is when to do these checks - a logical solution might be every time the popup is opened (as normally one can't delete a container without opening the popup).

Looking forward to any input as I'm eager to get it finished.

src/js/background/assignManager.js
```diff
@@ -114,8 +114,10 @@ const assignManager = {
   },
 
   async handleProxifiedRequest(requestInfo) {
+
+    // The following blocks potentially dangerous requests for privacy that come without a tabId
     if(requestInfo.tabId === -1)
-      return Utils.DEFAULT_PROXY;
+      return Utils.getBogusProxy();
```
stoically: Instead of generating a bogus proxy, which might have unexpected side-effects, an alternative could be to store the [`requestId`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/proxy/RequestDetails#Type) as "blocked" at this point, register a new `onBeforeRequest` handler, which listens to all request types, and return `cancel: true` in both `onBeforeRequest` handlers if the `requestId` is seen in there.

samuel-crypto: I agree this is a cleaner option in general - two questions, however:
 - Is onBeforeRequest already used somewhere in the extension? If so it might be a good idea to seamlessly integrate it into the call present there
 - Is it guaranteed that the request will be blocked before it takes place if `requestId` is stored after the proxy addListener call?
 - Finally, has it been tested that non-page requests don't actually proxy? Seeing they can be detected in the addListener method itself, perhaps they do get proxified? Might require some elaborate testing with other addons

stoically: > Is onBeforeRequest already used somewhere in the extension?

It is, but it only listens to `main_frame` requests, and service worker requests are AFAIK not `main_frame`.

> Is it guaranteed that the request will be blocked

Didn't test it myself, but it would only make sense that the proxy listener is called and fully await'ed before `onBeforeRequest` fires, because otherwise the request wouldn't have the correct [`proxyInfo`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/onBeforeRequest#details) attached.

> Finally, has it been tested that non-page requests don't actually proxy?

Didn't test that either, but on `about:serviceworkers` the service workers have `userContextId`s attached, so the requests generated from them could've the correct `cookieStoreId`, might be worth checking indeed.

jonathanKingston:
Hey @samuel-crypto sorry for my lack of response here. Are you interested in rebasing the patch, perhaps I can get some time to make some follow up commits perhaps?

samuel-crypto:
> Hey @samuel-crypto sorry for my lack of response here. Are you interested in rebasing the patch, perhaps I can get some time to make some follow up commits perhaps?

Thanks a lot for the reply! Yes, I'm all in for rebasing and getting it up to speed again, also don't forget to let me know re the proxy blocking (see above).

samuel-crypto:
@jonathanKingston I have just rebased it and noticed a few flaws meanwhile:
 - `localhost:3000` is not accepted as a valid proxy string (while of course `127.0.0.1:3000` is)
 - If you enter an invalid proxy string it doesn't tell you anything, it simply does not store what you entered - perhaps error colouring can be used on the input box?
 - I will look into whether clean up occurs, but last time I checked there was a design issue: I was unsure whether to add the deletion of the extra storage for proxies for a container on a routine basis (e.g. a cleanup scan occurs every time the add-on is opened) - something that would seem a bit redundant to me, or whether to attach it as one of the things that occurs in the delete dialog

xavier83:
any update +@groovecoder

xavier83:
@jonathanKingston , @groovecoder  doesn't seem to be active here, can you please request another reviewer and get it merged. Thanks.

groovecoder:
Howdy @xavier83 . We're looking at M-A-C issues again and this is another one we're considering finishing in the next month or so.

tstdin:
Hey @groovecoder, any update whether there is possibility of merging this PR?

groovecoder:
Note: I'm working on an updated version of this PR here: https://github.com/mozilla/multi-account-containers/pull/1903

Still not sure if/when we might be able to merge it, but I have an idea for this to combine with Mozilla VPN.

PR #1327, opened 2019-01-18
exempting `always open in` functionality for non-standard ports

### LINKED ISSUE #985: Allow default container setting to differentiate between ports
Basically, my email service and something I use for my job run on the same physical server, but at different port addresses. I am unable, as it stands, to get it to use different default containers for these different services.

e.g. 
example.org:2096 -> email -> Personal Container
example.org:2083 -> work site -> Work Container

I can do it using subdomains, but if I do this I lose SSL/TLS, and this is not something I can work around. As you can imagine I'd rather have TLS than Default Containers, but ideally I'd like both.

ShivangiKakkar:
#### Exempting `always open in` functionality for all ports other than 80 and 443
- Not sure if this is what we wanted, but it seems to work as far as I tested it.

### DIFF
--- a/src/js/background/assignManager.js
+++ b/src/js/background/assignManager.js
@@ -12,7 +12,11 @@ const assignManager = {
     getSiteStoreKey(pageUrl) {
       const url = new window.URL(pageUrl);
       const storagePrefix = "siteContainerMap@@_";
-      return `${storagePrefix}${url.hostname}`;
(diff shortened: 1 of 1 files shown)


ShivangiKakkar:
The particular test failing is also failing on master.

jonathanKingston:
I haven't tested this but the solution seems the correct one at least.

Issue #330, opened 2017-03-03
Add ability to reorganize Containers


gwynnarth:
It would be useful if users could change the order of Containers after they are created.

jonathanKingston:
I actually think this could be solved with drag and drop when in edit mode using the browser action.

We would need some form of index adding to the storage to remember what to display them into though.

erickt:
It would also be nice to alphabetically sort the containers. I use a lot of containers to manage a bunch of Amazon environments, so it's always a bit of a pain to find the right one.

billturk:
Yes being able to alphabetically sort the containers would help a lot!!!

smichel17:
Perhaps it would make more sense to add this functionality to about:preferences#containers ?

Narvey:
Yes @smichel17 , we should be able to manually order our containers, and have that affect the Sort Containers button (which seems to be useless at the moment, somebody correct me if I'm wrong).

t4777sd:
Definitely agree. Not being able to sort the containers makes them a lot more difficult to work with.

psychoslave:
I use containers as tabgroups replacement, I already defined more the 30 containers and I only use them since yersterday. Grouping alphabeticaly would be fine for me as I use a "theme/subtheme" nomemclature, however being able to drag&drop to move them around would be a big plus.

Note really the issue topic, but being able to switch context and only display included tabs with one click would be great. The current behaviour seems to be "open a new tab in this container" which is really odd for my use. To achieve what I want to do, I have to go in the target container submenu, select one tab, then go back to the container menu, find the container I no longer want to display, go in the submenu and finally hide it.  What I'm accustomed to with the grouptabs was "drop down list of group, select, tada" for the same result.

Having better isolation is fine, but in my priority list it's really far, far, far behind being able to easily switch context, that is having a group of tabs.

RadonX:
@psychoslave You will like this addon https://addons.mozilla.org/en-US/firefox/addon/conex/

psychoslave:
Thank you @RadonX for the suggestion. I already tested it, but it seems that currently it brings only more clutter in the toobar with 0 additional feature until some bug is fixed.

br3n065:
I've deleted/repurposed far to many containers, an option to rearrange them manually or alphabetically would be awesome.

o-l-a-v:
Please implement this. Either alphabetically or custom/ manually, or both. 
To only sort by creation time, is pretty useless.

[...]

PR #1347, opened 2019-02-24
added code to sort the containers list

gvlachos:
I have been using this extension for a while now (awesome work!) and one feature I was missing is the ability to sort the containers list, so I went ahead and added it.

There is a new button with functionality to sort the containers list alphabetically, if not already sorted, or toggle the sort order if they are previously sorted. The two icons I used are from [Feather icons](https://feathericons.com) and they are MIT licensed.


IOAyman:
Fixes #330

ReporterX:
The sort is case sensitive. It would be better not to be case sensitive.

gvlachos:
@ReporterX that is a good point, I have made the change.

To the maintainers: I've made sure my commits are after any latest changes so there are currently no conflicts (rebase in my branch). I would appreciate a review or some feedback.

Lunique:
How can I get this function I really need it!

gvlachos:
Hi @maxxcrawford , thanx for the feedback! I am bit swamped at work at the moment but I expect I'll be able to have a go at it in about a month.

maxxcrawford:
@gvlachos Sounds great. Thanks again for contributing!

jonathanKingston:
My hope was that we could change the index within firefox itself so everyone could benefit. I can probably help with what changes would be needed if you were interested in helping. Essentially the changes would be to add an index within the container data within firefox and expose that through the extension API rather than store it internally in the extension. As the change within the addon will only impact the browser doorhanger rather than the file and + menus that are native within firefox.

gvlachos:
@jonathanKingston using the firefox internal index makes a lot of sense! 

It would be great if you could give me some hints (or links to docs) on how to do that. I am not familiar with the way container are set up within firefox, though quite curious to learn  :)

jonathanKingston:
> @jonathanKingston using the firefox internal index makes a lot of sense!
> 
> It would be great if you could give me some hints (or links to docs) on how to do that. I am not familiar with the way containers are set up within firefox, though quite curious to learn :)

For advice on building Firefox: https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Build_Instructions/Simple_Firefox_build#Building TL;DR clone the repo with mercurial `hg clone https://hg.mozilla.org/mozilla-unified` install the dependencies with `./mach bootstrap`, build with `./mach build` and run the browser with `./mach run`

The core of the code within Firefox for containers is within the ContextualIdentityService which is the internal name we use for containers: https://searchfox.org/mozilla-central/source/toolkit/components/contextualidentity/ContextualIdentityService.jsm

So I would advise trying to get a build working and maybe make that function randomise the output or alphabetically sort perhaps. This should change the long press + button results and also the file menu. From there we can think about how to interface this to the web extensions.
As you can see there are a few default containers that we have there and there are various keys of data. The change would require a sorting index to be added to the `_identities` object along with saving that or simply using the array index and allowing that to be modified depending on which is simpler (You will notice that any method within this service eventually will call save() which actually writes it out to disk so it can be restored).

`getPublicIdentities` is as far as I can tell is exclusively the only method we would care about for sorting within the interface components of Firefox: https://searchfox.org/mozilla-central/search?q=getPublicIdentities&case=false&regexp=false&path= so whatever that returns would be reflected whenever the user clicks on a container menu.

PR #1368, opened 2019-04-15
Possible changes to fix #1367

### LINKED ISSUE #1367: Hard coded cookie store id prefix breaks extension functionality on re-branded forks
- Multi-Account Containers Version: 6.1..0
- Operating System + Version: Trisquel GNU/Linux 7.0

### Actual behavior
Editing a container creates a copy of the expected result and the original container remains.  Removing a container doesn't work.

### Expected behavior
Editing and removing containers should work.

### Steps to reproduce
1. Install a re-branded fork of Firefox such as Abrowser on Trisquel
2. Install Multi-Account Containers extension
3. Attempt to edit or remove a container

### Notes
The code in these files assume that the prefix of cookieStoreId's from browser.contextualIdentities.create() is "firefox" when it may be different in re-branded forks of Firefox.

src/js/background/backgroundLogic.js
src/js/background/messageHandler.js
src/js/popup.js

I understand if this is considered outside the scope of this particular extension, and I could always fork it, but if it could be made to work generically on re-branded Firefox versions that are common in other operating systems that would probably be best.  I'm not currently setup to really try to work on the code myself, but I have created at least a sample of changes that might fix the issue as a pull request.

Aaron-P:
Using a dynamic method to get the container cookie id prefix to avoid hard coded prefixes that are incorrect in re-branded Firefox forks.

### DIFF
--- a/src/js/background/backgroundLogic.js
+++ b/src/js/background/backgroundLogic.js
@@ -1,6 +1,6 @@
 const DEFAULT_TAB = "about:newtab";
-const BROWSER_NAME = (await browser.runtime.getBrowserInfo()).name.toLowerCase();
 const backgroundLogic = {
+  BROWSER_NAME: (await browser.runtime.getBrowserInfo()).name.toLowerCase(),
   NEW_TAB_PAGES: new Set([
     "about:startpage",
     "about:newtab",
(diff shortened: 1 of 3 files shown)


groovecoder:
Thanks for the PR @Aaron-P. Have you verified that these changes fix this issue on Firefox forks?

Aaron-P:
It apparently doesn't, the code is throwing errors for some reason that I currently can't figure out, but I just slapped this together really fast.  Either way though after thinking about I have a much better idea for how to fix this, so this pull should be ignored.  I'll make a new pull request when I have a better fix.

PR #1369, opened 2019-04-15
Refactor anonymous async functions into arrow-functions

ctgardner:
Fixes https://github.com/mozilla/multi-account-containers/issues/972

I've refactored all anonymous async functions declared using the `function` keyword into arrow-functions.

I understand the crux of this issue isn't to simply replace functions declared using the `function` keyword with arrow-functions, but to inline named async functions where sensible. I've reviewed all named async functions and I believe none would benefit from being inline, with the exception of a few handlers registered using `addListener`. Given the past confusion surrounding this issue, I believe this issue should be closed.

### DIFF
--- a/test/issues/940.test.js
+++ b/test/issues/940.test.js
@@ -160,7 +160,7 @@ describe("#940", () => {
       await promise2;
 
       background.browser.tabs.create.should.have.been.calledTwice;
-    }).timeout(2005);
+    }).timeout(2002);
 
     it("should not influence the canceled url in other tabs", async () => {
       const newTab = {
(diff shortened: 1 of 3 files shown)


test/issues/940.test.js
```diff
@@ -160,7 +160,7 @@ describe("#940", () => {
       await promise2;
 
       background.browser.tabs.create.should.have.been.calledTwice;
-    }).timeout(2002);
```
jonathanKingston: Why this change out of interest?

ctgardner: This test was passing before the refactor and began failing after making the changes. As the test seems unrelated to the changes, I increased the timeout. (Two millis is the minimum increase necessary to make the test pass on my machine.)

ctgardner: The test is passing on CI after reverting the timeout change 🤷‍♂

jonathanKingston: Heh thanks!

jonathanKingston:
Merged thanks for this!

PR #1377, opened 2019-04-18
Sort tabs with less steps

### LINKED ISSUE #1373: [Feature request] Sort tabs with less steps
Currently "Sort Tabs" command moves all tabs always, even if they don't need to be moved:
https://github.com/mozilla/multi-account-containers/blob/6f4d3c43274701e8c166f3ed80e4d453860fd31d/src/js/background/backgroundLogic.js#L278-L286

 * This operation produces too much tab moves. When there are too many tabs (for example, over 1000+) it may take much time.
 * It scrambles tabs unexpectedly and unnecessarily. It is critically dangerous on some kinds of addons like Tree Style Tab. 

So, instead only minimum tabs should be moved based on the algorithm of "diff". Actually [Tree Style Tab addon does that](https://github.com/piroor/treestyletab/blob/ce805bdfdd00d700400d98f17b57f4679de7c3ee/webextensions/background/tabs-move.js#L336). TST uses a [diff library ported from Python](https://github.com/piroor/treestyletab/blob/master/webextensions/common/diff.js), but there are more alternatives, for example:

* https://github.com/kpdecker/jsdiff
* https://github.com/jhchen/fast-diff
* https://johnresig.com/projects/javascript-diff-algorithm/

How about such an improvement?

piroor:
This should fix #1373,

### DIFF
--- a/src/js/.eslintrc.js
+++ b/src/js/.eslintrc.js
@@ -6,6 +6,7 @@ module.exports = {
     "assignManager": true,
     "badge": true,
     "backgroundLogic": true,
+    "differ": true,
     "identityState": true,
     "messageHandler": true
   }
--- a/src/js/background/backgroundLogic.js
+++ b/src/js/background/backgroundLogic.js
@@ -241,12 +241,12 @@ const backgroundLogic = {
     const windows = await browser.windows.getAll({populate: true});
     for (let windowObj of windows) { // eslint-disable-line prefer-const
       // First the pinned tabs, then the normal ones.
-      await this._sortTabsInternal(windowObj, true);
-      await this._sortTabsInternal(windowObj, false);
+      this._sortTabsInternal(windowObj, true);
+      this._sortTabsInternal(windowObj, false);
     }
   },
 
-  async _sortTabsInternal(windowObj, pinnedTabs) {
+  _sortTabsInternal(windowObj, pinnedTabs) {
     const tabs = windowObj.tabs;
     const sortedTabs = [];
     let offset = 0;
@@ -289,11 +289,6 @@ const backgroundLogic = {
       if (newIndex < 0)
         newIndex = beforeIds.length;
       let oldIndices = movingIds.map(id => sortedIds.indexOf(id));
-      // Reject already removed tabs.
-      movingIds = movingIds.filter((id, index) => oldIndices[index] > -1);
-      if (movingIds.length === 0)
-        continue;
-      oldIndices = oldIndices.filter(index => index > -1);
       if (oldIndices[0] < newIndex)
         newIndex--;
       browser.tabs.move(movingIds, {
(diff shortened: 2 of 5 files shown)


groovecoder:
This is a big addition and change ... will probably take a long while to get to review this. 😢

piroor:
Here are some additional information for reviewing:

* The largest addition `differ.js` is a ported version from [the `diff` module of npm](https://github.com/kpdecker/jsdiff). I just modified it to remove needless codes for this usecase.
* The differ detects three type differences: "added", "removed", and "unchanged". Rearranging of tabs is based on a logic:
   1. Prepare three arrays: initial order (`beforeIds`), final order (`afterIds`), and the current order (`sortedIds`: this is required because `tabs.move()` is asynchronous and we need to track the order of tabs after tab moves.)
   2. Extract only "added" differences. Any "removed" difference should be ignored, because no tab is removed while rearranging.
   3. Calculate destination position of "added" tabs, based on the current order (`sortedIds`).
   4. Move "added" tabs actually via `tabs.move()`.
   5. Update the current order (`sortedIds`) for the next calculation.

piroor:
Here is a blog post describing the logic: [How to sort tabs smartly on Firefox addons and Chromium extensions? - Qiita](https://qiita.com/piroor/items/5e338ec2799dc1d75e6f)

jonathanKingston:
In the article you mention that the native API tries to move tabs if it's not required. I haven't tested but the code in Firefox at least seems that isn't the case: https://searchfox.org/mozilla-central/rev/e7d9a8749303b39dadcc0e18ea0d60a570a68145/browser/base/content/tabbrowser.js#3710

I suspect the existing algorithm could be improved greatly:
- Reduce the move distance within the container by preferring to keep the tab's position
- Flatten the sort map to create an ordering array for all tabs in the window
- Call one move API operation per window for all tabs and await the move completing
- Sort the focused window first and then sort other windows after to reduce the feedback lag for users

If this difference algorithm still has improvements over that, perhaps we should be including it natively. It seems to me that just calling one api request `await browser.tabs.move([1,12,8,7,2....], {windowId});` would solve the problem greatly.

piroor:
Yes, Firefox doesn't move tab if the specified index is same to the current one. However, it is not well-optimized. This is a worst case example:

```javascript
// originally tabs are ordered as: [7,1,2,3,4,5,6]
browser.tabs.onMoved.addListener(tabId => console.log('moved ', tabId));
browser.tabs.move([1,2,3,4,5,6,7], { index: 0 });
```

This case actually triggers 6 tab moves as:

```
move  1
move  2
move  3
move  4
move  5
move  6
```

Such a case will happen for example when tabs 1-6 are opened with the "default" container and the tab 7 is opened with a specific container. I think that implementing a FMAC-specific algorithm to calculate the minimum tab moves on various cases looks hard, and this is the reason why I proposed the differential algorithm as the universal algorithm.

kendallcorner:
Thanks for this contribution. However, we don’t plan to merge this, and now that we’ve launched the [Sync](https://blog.mozilla.org/security/2020/02/06/multi-account-containers-sync/) functionality, we’re closing some open PRs to make room for a couple more bug-fixes and updates.

piroor:
OK, I agree that this approach costs too expensive for the purpose.

Hmm, but I still need something solution to solve problems from such bulk tab move on switching of visible tabs...

PR #1387, opened 2019-05-01
Bug 1539507 - Fix targeting of button click when button element isn't…

jonathanKingston:
… clicked.

TL;DR is that the `explicitOriginalTarget` appears to have been changed to not be always the button. A better fix might be to stop using it and use a global to store the last clicked button on the form. I'm pretty suprised there isn't a more elegant way to get this info :/

### DIFF
--- a/src/js/confirm-page.js
+++ b/src/js/confirm-page.js
@@ -20,13 +20,20 @@ async function load() {
 
   document.getElementById("redirect-form").addEventListener("submit", (e) => {
     e.preventDefault();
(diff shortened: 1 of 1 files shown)


jonathanKingston:
Oh sorry @stoically I didn't see that. I think this preserves the enter behaviour so we should keep this. However as mentioned I'm happy to take another follow up clean up if anyone has anything better.

stoically:
@jonathanKingston @groovecoder If you get a chance to prepare a new release with this bugfix, that'd be great

PR #1435, opened 2019-06-09
Support dark themes on icon - Fixes #1424

### LINKED ISSUE #1424: Addin icon almost invisible when using Dark theme
### Actual behavior
The Multi Account Container icon in the FF toolbar almost blends in with the default Dark theme FF making it difficult to see on certain monitors.  Even in Compact Dark mode the icon is still too dark to see.

### Expected behavior
The icon should stand out and be more visible when users are using Dark theme.

Normal Dark Mode:
firefox_2019-05-24_09-43-20

Compact Dark Mode:
firefox_2019-05-24_09-46-00

silverwind:
On light theme: light

On dark theme: dark


src/manifest.json
```diff
@@ -47,7 +47,12 @@
     "browser_style": true,
     "default_icon": "img/container-site.svg",
     "default_title": "Multi-Account Containers",
-    "default_popup": "popup.html"
+    "default_popup": "popup.html",
+    "theme_icons": [{
+        "light": "img/container-site-light.svg",
+        "dark": "img/container-site.svg",
+        "size": 32
```
silverwind: Note: Size is mandatory according to [this](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_action#theme_icons). I don't know why thought.

ThinkWithPbody:
Will color.firefox.com get supported in the future?

silverwind:
Firefox Color is not supported by this, so that's likely a separate issue. I found [this API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/theme/getCurrent) but it would require the icon the be generated dynamically via JavaScript instead of static SVGs.

PR #1464, opened 2019-07-17
Refactor tests

stoically:
Here's a test change only - updated with the latest dependencies - which was previously in #1190, but since it doesn't touch production code, I think it's good to merge into master.


maxxcrawford:
@stoically I'm not able to run this patch with all passing tests. Is there something wrong on my end or is that a known issue with this patch?

stoically:
@maxxcrawford You're right, it seems something introduced a regression since I've opened the PR (travis was green back then). Reopening to trigger a red build and looking into it when I get the chance.

stoically:
Interesting. So travis is green again, but locally `creating a new container should create it in the browser as well` fails for me. Odd.

stoically:
@maxxcrawford rebased and updated to latest [`webextensions-jsdom`](https://github.com/stoically/webextensions-jsdom) (which depends on latest [`webextensions-api-fake`](https://github.com/stoically/webextensions-api-fake/) and now works for me locally as expected. Could you give it another look? Thanks!

stoically:
Separated the changes into multiple commits as requested, hope this helps with readability. As an aside: Facebook Container uses the dependency `webextensions-jsdom` as well.

Looking forward to see progress in MAC!

Issue #948, opened 2017-11-10
Extra scrollbars are wrongly displayed while editing a container if it has more than 9 websites assigned


Softvision-CristinaBadescu:
[Affected versions]:
- Firefox 56.0.2 and up
- Multi-Account Containers v4.0.3

[Affected Platforms]:
- All Windows

[Prerequisites]:
- Have a clean Firefox profile with the Multi-Account Containers v4.0.3 installed.
- Have one container populated with minimum 10 assigned websites.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites.
2. Click the addon button from the browser toolbar and choose the "Edit container" button.
3. Click on the container's pencil icon and scroll all the way down.
4. Observe the assignments panel.

[Expected result]:
- One vertical scrollbar is displayed in the panel.

[Actual result]:
- An additional horizontal scrollbar is displayed.

[Notes]:
- If the horizontal scrollbar is moved to the right, an extra vertical scrollbar appears.
- This issue is not reproducible if you have under 9 assigned websites to a container, see below a screenshot:
![9 websites assigned](https://user-images.githubusercontent.com/31244235/32669121-275365f0-c648-11e7-89ab-b29245cdd7b4.png)

- Here is a screen recording of the issue:
![scrollbar](https://user-images.githubusercontent.com/31244235/32669244-7fec1964-c648-11e7-9b82-0406c882b598.gif)
![scrollbar](https://user-images.githubusercontent.com/31244235/32669252-869df0ca-c648-11e7-839d-fc616e6f7046.png)
- This issue is not reproducible on Mac, nor on Linux.

groovecoder:
The fix probably involves scrollable class.

Paarmita:
@groovecoder Can I take up this issue?

groovecoder:
Yes, this is another good first bug.

Paarmita:
@groovecoder Thanks! I will soon add a PR.

groovecoder:
This was fixed in https://github.com/mozilla/multi-account-containers/pull/1476

PR #1476, opened 2019-08-06
Bugfix: incorrect scrollbars/spacing if edit container with many assignments

mckenfra:
#948 Fixes big empty gap above assignments list and multiple scrollbars, when editing container with many assignments.

### DIFF
--- a/src/css/popup.css
+++ b/src/css/popup.css
@@ -23,6 +23,15 @@ body {
   max-inline-size: 300px;
 }
 
+/**
+ Bugfix: incorrect scrollbars/spacing when editing container with many assignments
+ https://github.com/mozilla/multi-account-containers/issues/948
+ */
+html,
+body {
+  block-size: 100%;
+}
+
 :root {
   --primary-action-color: #248aeb;
   --title-text-color: #000;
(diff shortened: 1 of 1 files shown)


src/css/popup.css
```diff
@@ -783,8 +792,20 @@ span ~ .panel-header-text {
...
+ */
+.edit-container-panel .columns {
+  overflow: hidden;
+}
+
 #edit-sites-assigned {
-  flex: 1;
+  /**
+   Bugfix: incorrect scrollbars/spacing when editing container with many assignments
+   https://github.com/mozilla/multi-account-containers/issues/948
```
jonathanKingston: I think these code comments are a little excessive, I would remove them.

src/css/popup.css
```diff
@@ -783,8 +792,20 @@ span ~ .panel-header-text {
...
+   */
+  flex: 1000;
```
jonathanKingston: Can you explain why the 1000 value, it seems like a code smell from what I can tell?

src/css/popup.css
mckenfra: Ok sure, I'll remove the comments

src/css/popup.css
mckenfra: The value needs to be bigger than 1, so that the assignments part of the panel grows to fill the available space in preference to the top part of the panel (containing the name, color, icon). Exactly how much bigger than 1 it needs to be is debatable. Possibly 2 would have been sufficient. But I chose 1000 so that it is obvious that this part of the panel must take ALL free space, and not share even the slightest bit of it.

Hopefully the below pictures will give you an idea of the issue. The first picture shows the existing bug. The last picture shows my solution, using `flex: 1000`. You can see it's identical to the picture that uses `flex: 2`. Personally, I would stick with 1000.

As one final side note, usually you would use `flex: 0` and `flex: 1` to control what grows to fill empty space, and what doesn't. But this isn't possible in this situation, because the panel for adding a new container shares the same HTML as this one for editing an existing container, and although `flex: 0` and `flex: 1` works for the latter (edit container), it messes up the spacing on the former (add container).

flex-1

flex-1 2

flex-1 5

flex-2

flex-1000

jonathanKingston: Cool thanks for the explanation, thats good enough for me. We could have overriden the flex value for the top panel with a parent class I guess but it's debatable if it's worth it.

jonathanKingston:
Thanks for fixing this.

Issue #1375, opened 2019-04-18
Add Dark Mode for "Open this site in your assigned container?" page


dannycolin:
- Is "Firefox will: Never remember history" in the Firefox Preferences/Options under "Privacy & Security > History" selected? Yes/No: **no**
- Are you using Firefox in a Private Window? Yes/No: **no**
- Can you see a grayed out but ticked Checkbox with the description "Enable Container Tabs" in the Firefox Preferences/Options under "Tabs"? Yes/No: **yes**
- Multi-Account Containers Version: 6.1.0
- Operating System + Version: Debian 9
- Firefox Version: 68
- Other installed Add-ons + Version + Enabled/Disabled-Status:
  - HTTPS Everywhere 2019.01.31
  - UBlock Origin 1.18.16
  - UMatrix 1.3.16
  - Privacy Badger 2019.02.19

### Actual behavior

"Open this site in your assigned container?" has a light theme even if firefox dark theme is actived.

### Expected behavior

"Open this site in your assigned container?" should match the firefox theme actived

### Steps to reproduce
1. Open a website you've assigned for the first time to a container
2. "Open this site in your assigned container?" page appear

stoically:
Add-ons can't detect the currently activated theme in Firefox. So this would need a setting in the Add-on.

dannycolin:
> Add-ons can't detect the currently activated theme in Firefox. So this would need a setting in the Add-on.

@stoically according to the doc there's a theme.getCurrent() (see : https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/theme).

stoically:
That unfortunately doesn't give you the information whether the "default", "dark" or "light" theme is activated, just some [metadata](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/theme).

dannycolin:
@stoically that's enough information to guess if the theme is a light or dark one by calculating the contrast ratio. This is exactly how the Firefox devs have implemented the `prefers-color-scheme` on Linux. :)

jonathanKingston:
We should probably just use `@prefers-color-scheme` in CSS if we can help it rather than the API call.

jonathanKingston:
Also all the Firefox error pages are still light themed right? This is where the design for this came from, so until we have more examples I don't think we can easily more forward here.

dannycolin:
> Also all the Firefox error pages are still light themed right?

[...]

PR #1487, opened 2019-08-29
#1375 - Correct colors on confirmation page

htwyford:
Firefox's dark mode sets the background of the confirmation page to Grey 80 (`#2a2a2e`). As pointed out in [this comment](https://github.com/mozilla/multi-account-containers/issues/1375#issuecomment-500107995), the `#redirect-url` background remains `#efefef`, making the redirect url unreadable. 

This patch corrects the background color in dark mode. It also slightly changes the color in light mode to make it consistent with the [Photon colors](https://design.firefox.com/photon/visuals/color.html).

### DIFF
--- a/src/css/confirm-page.css
+++ b/src/css/confirm-page.css
@@ -47,7 +47,7 @@ html {
 }
 
 #redirect-url {
-  background: #efefef;
+  background: #efedf0; /* Grey 20 */
   border-radius: 2px;
   line-height: 1.5;
   padding-block-end: 0.5rem;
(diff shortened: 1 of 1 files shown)


stoically:
Since we run an old version of `stylelint` it doesn't know about `prefers-color-scheme` yet, however upgrading it introduces a bunch of other lint errors - so we should probably upgrade in a dedicated PR. Would you mind disabling the linting for `prefers-color-scheme` with `/* stylelint-disable */` so that we have a green build?

jonathanKingston:
Does this still repro in nightly? I wasn't able to get dark themed pages there?

stoically:
Can reproduce in Nightly 71.0a1 (2019-09-23) with `ui.systemUsesDarkTheme` set to `1`.

jonathanKingston:
Ah I can now too, I'm not sure why it wasn't working before maybe I fat fingered the pref name.

Issue #1511, opened 2019-10-03
Would like to suggest making the extension appear dark/light mode according to user preference on browser


silvererudite:
### behavior
I believe the extension would be more accessible if we could implement dark/light mode automatically according to user preference on browser. For ex: I have my browser set to dark theme and when I install the Add-on it should appear in dark mode and vice versa.

## Note:
asking for attention of @groovecoder and @maxxcrawford. If this is a desirable feature according to you guys i would love to work on it as part of my Outreachy contribution.

silvererudite:
@groovecoder may I work on this as part of my Outreachy contribution?

maxxcrawford:
@silvererudite Assigned!

silvererudite:
@maxxcrawford currently I'm being able to customize according to user preference on his/her local machine's OS. You know with CSS property  prefers-color-scheme: . however I wanted to detect the browser preference and make the addon automatically adjust to that which this property doesn't. 

Do you know if there's a way to detect what theme the browser is using?

groovecoder:
You can use https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/theme/getCurrent to get the current theme, and that object might have enough to tell whether it's "dark" or "light"?

silvererudite:
@groovecoder thanks for the resource but **what is the correct way to use the getCurrent() function**. Before changing the addon files i was trying to see if I can retrieve my browser theme info on a separate extension I made for testing.  But when I try to naively call the getCurrent() function it outputs nothing in the console. I also tried the example from the link's page and it also shows nothing. 

I'm attaching the ss of my test extension files for reference
![Screenshot (21)](https://user-images.githubusercontent.com/24827548/66254574-f51f6a00-e799-11e9-9c64-fa5ec07a3d7a.png)
![Screenshot (24)](https://user-images.githubusercontent.com/24827548/66254576-f81a5a80-e799-11e9-89c8-2060ad3a20da.png)

maxxcrawford:
@silvererudite In the [Contain Facebook project](https://github.com/mozilla/contain-facebook), we set two different icons within the manifest to cover this. 

See file entry here: https://github.com/mozilla/contain-facebook/blob/master/src/manifest.json#L29-L40

silvererudite:
[...]

PR #1521, opened 2019-10-09
added the ability to change theme based on user's browser preference …

silvererudite:
Hi @maxxcrawford and @groovecoder. I have opened this PR to show you guys my progress on #1511 .
The theme-changing ability works fine on the browser's light and custom theme but I need review on the dark theme. I'm attaching some ss for reference below
Screenshot (5)
Screenshot (6)
Screenshot (7)
Screenshot (9)
Screenshot (10)
Screenshot (11)
Screenshot (12)

as you can see even though the extension turns dark based on toolbar color, the middle list of containers still remains white. I don't really know why. i have tried to use js code selecting that part of the popup.html but it still doesn't work. I have commented the code on check-theme.js file for reference as well. Please suggest a fix for this (if you think it's required) and some feedback. 
I would gladly make changes with your review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,26 @@ For more info, see:
 1. `npm install`
 2. `./node_modules/.bin/web-ext run -s src/`
 
+or
+
+you can also test your changes without having to install web-ext as step #2 above
+
+**Debugging in the browser**
+
+Visit `about:debugging` in your browser.
+
+Then select the option `load temporary Add-on`
+and load the extension by selecting any file from the Web Extensions' dir. In our case, e.g. select manifest.json from the src dir. 
+
+For reference, [watch this video](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Your_first_WebExtension#Installing)
+
+After you make your changes, simply press the `reload` button to see them in effect.
+
+
+
+
+
+
 ### Testing
 TBD
 
--- a/README.md
+++ b/README.md
@@ -21,26 +21,6 @@ For more info, see:
 1. `npm install`
 2. `./node_modules/.bin/web-ext run -s src/`
 
-or
-
-you can also test your changes without having to install web-ext as step #2 above
-
-**Debugging in the browser**
-
-Visit `about:debugging` in your browser.
-
-Then select the option `load temporary Add-on`
-and load the extension by selecting any file from the Web Extensions' dir. In our case, e.g. select manifest.json from the src dir. 
-
-For reference, [watch this video](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Your_first_WebExtension#Installing)
-
-After you make your changes, simply press the `reload` button to see them in effect.
-
-
-
-
-
-
 ### Testing
 TBD
 
--- a/src/js/check-theme.js
+++ b/src/js/check-theme.js
@@ -21,6 +21,7 @@ function getStyle(themeInfo) {
     document.getElementById('sort-containers-link').style.color=themeInfo.colors.tab_background_text;
  
     
+    
   }   
 
 }
--- a/src/js/check-theme.js
+++ b/src/js/check-theme.js
@@ -1,0 +1,51 @@
+/* eslint-disable linebreak-style */
+/* eslint-disable no-redeclare */
+/* eslint-disable quotes */
+/* eslint-disable no-var */
+/* eslint-disable indent */
+
+function getStyle(themeInfo) {
+  
+  if (themeInfo.colors.toolbar) 
+  {
+  
+    document.body.style.backgroundColor= themeInfo.colors.toolbar;
+  
+    document.getElementById('current-tab').style.color=themeInfo.colors.tab_background_text;
+   
+  const headers = document.getElementsByTagName('h3');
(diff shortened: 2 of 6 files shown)


src/manifest.json
```diff
@@ -30,7 +30,8 @@
     "storage",
     "tabs",
     "webRequestBlocking",
-    "webRequest"
+    "webRequest",
+    "theme"
```
groovecoder: I'm asking our @mozilla/addons-editorial team if this will trigger a new permission prompt to users. That's usually a UX bump that we like to avoid if we can.

src/js/check-theme.js
```diff
@@ -0,0 +1,51 @@
...
+/* eslint-disable quotes */
+/* eslint-disable no-var */
+/* eslint-disable indent */
+
+function getStyle(themeInfo) {
+  
+  if (themeInfo.colors.toolbar) 
+  {
+  
+    document.body.style.backgroundColor= themeInfo.colors.toolbar;
```
groovecoder: ooooh, this is a neat approach.

src/js/check-theme.js
```diff
@@ -0,0 +1,51 @@
...
+  
+    document.getElementById('current-tab').style.color=themeInfo.colors.tab_background_text;
+   
+  const headers = document.getElementsByTagName('h3');
```
groovecoder: We should add classes to the elements so we can style them by class instead of tag name.

src/js/check-theme.js
```diff
@@ -0,0 +1,51 @@
...
+    } 
+   }
+   document.getElementById('sort-containers-link').style.color=themeInfo.colors.tab_background_text;
+
+   var legends1= document.getElementsByTagName('legend');
+  for (var i = 0; i < legends1.length; i++) {
+    legends1[i].style.color =themeInfo.colors.tab_background_text;
+    } 
+    
+    //this part of the code is for changing the popup's rows showing conatinerslist (it doesn't work)
```
groovecoder: @maxxcrawford - can you check this out?

src/manifest.json
groovecoder: Answer: no, [adding `theme` will not re-prompt the user for permission](https://searchfox.org/mozilla-central/source/toolkit/components/extensions/test/xpcshell/test_ext_permissions.js#562-585).

src/js/check-theme.js
silvererudite: @groovecoder and @maxxcrawford after much trials, I have figured out how to solve this. I will update the PR with new commits.

silvererudite:
@groovecoder and @maxxcrawford after the latest commit, the popups look like this
Screenshot (14)
Screenshot (15)
for the second window I deliberately kept the style like this because if the entire rows were colored then we'd most prolly need to change the icons as they wud look invisible in dark/light theme. 
Please let me know what you guys think

src/js/check-theme.js
groovecoder: Could we add a new class name to each of them - e.g., `tabBackgroundText`?

Then we can stye them all with a single for loop:
```js
for (const elToStye of document.querySelectorAll(".tabBackgroundText")) {
  elToStyle.style.color =themeInfo.colors.tab_background_text;
}
```

silvererudite: @groovecoder turns out I can be quite dim when it comes to optimizing code :D , this is definately better approach and reduces clutter. I'm updating the PR with these refactoring. Would love to improve with more suggestions like this.

README.md
```diff
@@ -21,6 +21,26 @@ For more info, see:
 1. `npm install`
 2. `./node_modules/.bin/web-ext run -s src/`
 
+or
+
```
silvererudite: @groovecoder and @maxxcrawford  just thought on providing some additional ways to develop. I found the second approach quite easy and maybe it would help someone. however I thought it would open to a separate PR.

silvererudite: I intended to send the modifications of the README as a separate PR but for some reason it automatically gets merged with this PR. Is it not possible to send a  new  PR with different changes until this PR is closed?

silvererudite: removed the changes in README for now.

silvererudite: removed the changes in README for now.

silvererudite:
this PR closes #1511

groovecoder:
Looking good. For the `.edit-container.pop-button` and `.remove-container.pop-button` - could you apply a [cascading svg fill color](https://css-tricks.com/cascading-svg-fill-color/) to make the svg icons match the color of the text?

silvererudite:
@groovecoder turns out the svgs don't look that invisible if theme changing or dark mode is applied. but there's still an unresponsive part which I don't really know why. 
Need some help fixing that . I'm sure I'm making some stupid mistake. Here's the current ss for ref
Screenshot (18)

silvererudite:
@groovecoder I managed to fix the white space around the svg in previous commit. However I had to disable a CSS rule for that. Even though it fixes my problem I'm not sure if doing so breaks sth else. Here's the ss for the current appearance
Screenshot (28)
Also I hate the fact that it reduced space between the icons

kendallcorner:
Thanks for this contribution. However, we don’t plan to merge this, and we're about to update the UI, so we’re closing some open PRs to make room for a couple more bug-fixes and updates.

Issue #1361, opened 2019-03-19
Opening in assigned container url-decodes parameters


Rob--W:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1535643 for the full report and STR.

I think that the issue is caused by the `decodeURIComponent` call here:
https://github.com/mozilla/multi-account-containers/blob/99db19279225664f943df7c873e01c355c875b26/src/js/confirm-page.js#L3

The parameter is set at:
https://github.com/mozilla/multi-account-containers/blob/c2f2d69ba1685a74335837ec86a0896779729a2f/src/js/background/assignManager.js#L426

The URLSearchParams API (`new URL(...).searchParams`) already decodes the parameter, so the extra `decodeURIComponent` call results in double-decoding.

kbenson:
As a very simple way to test this bug, if you have a google container that requests you switch when accessing google.com URLs, and to a Ctrl-k search for "at&t", after authorizing the switch to the google container you'll find your search is for "at".

The full chain looks like this:
1) Ctrl-k search for at&t and redirected to container chainge confirm page with URL moz-extension://bd774d98-6895-4a38-a343-f40686fefbf4/confirm-page.html?url=https%3A//www.google.com/search%3Fclient%3Dfirefox-b-1-d%26q%3Dat%2526t&cookieStoreId=firefox-container-1

2) On authorizing the container switch, you are sent to the following URL https://www.google.com/search?client=firefox-b-1-d&q=at&t

Note that q=at&t is not the same as q=at%26t (which is what google shows when you manually search there)

Rob--W likely identified the location this is happening, but I figured a fairly easy way to test it would be useful.

Rob--W:
@stoically This bug has steps to reproduce (last comment) and a description that is a hint towards the simple fix (original report).

Could you slap the good-first-bug label on this issue?

kendallcorner:
@maxxcrawford I'm working on this one.  Currently testing.

@Rob--W it works for this one example. Any pointers on testing this? Just find a site that uses a lot of %s and other characters in their urls?

maxxcrawford:
@kendallcorner Assigned. Thanks!

Rob--W:
@kendallcorner I wanted to reply to https://github.com/mozilla/multi-account-containers/issues/1361#issuecomment-542365390 , but noticed that you've already found a test case, created a patch and got it merged.

Thanks for the patch :)

Rob--W:
[...]

PR #1535, opened 2019-10-15
Fix for double decoding on confirm page issue #1361

kendallcorner:
Fix for #1361 

# Steps to reproduce bug:
from: https://bugzilla.mozilla.org/show_bug.cgi?id=1535643
1. Go to https://www.example.com/path?param=some%2Bvalue in a container
2. Mark it to "Always open in <container>"
3. Open a new tab outside of the container
4. Go to https://www.example.com/path?param=some%2Bvalue
5. Look at the "Open this site in your assigned container?" page

Actual results:
Redirect url is show as "https://www.example.com/path?param=some+value" - notice that "some%2Bvalue" is now "some+value".  Open in <container> button will also open that url and not the original.

Expected results:
Redirect url should show as "https://www.example.com/path?param=some%2Bvalue". Open in <container> button should also open that url.

# Test fix
Repeat above steps. Other urls with special characters still work.

Some other urls to test:
https://www.target.com/s?searchTerm=spiderman+costume
https://github.com/search?q=unicorn+horns
kendallcorner.com/test%25file.html (not a valid url, but it will show a change on the "Open this site in your assigned container?" page in the unfixed version.)

### DIFF
--- a/src/js/confirm-page.js
+++ b/src/js/confirm-page.js
@@ -1,6 +1,6 @@
 async function load() {
   const searchParams = new URL(window.location).searchParams;
-  const redirectUrl = decodeURIComponent(searchParams.get("url"));
+  const redirectUrl = searchParams.get("url");
   const cookieStoreId = searchParams.get("cookieStoreId");
   const currentCookieStoreId = searchParams.get("currentCookieStoreId");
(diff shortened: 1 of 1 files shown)


kendallcorner:
@maxxcrawford

Issue #323, opened 2017-03-03
Open bookmark into Container


gwynnarth:
Currently it is possible to open a link that's on a page into a specific container by right-clicking and choosing appropriate "context" from dropdown menu, but you cannot do that with a bookmark, neither from Bookmark bar nor from Bookmarks menu.
It would be useful to be able to right-click a bookmark and open it into a container. Or might be even more useful if user could associate a particular bookmark with a particular Container so that it automatically opens in the right Container. So when I left-click "Facebook" bookmark it opens in "Facebook" container, when I left-click my Online banking website it automatically launches into "Finances" container.

aleth:
Similarly, in the Bookmarks menu, "Open All in Tabs" could easily be extended to "Open All in... >" with a submenu of containers.

jonathanKingston:
There is a platform bug for this as I'm not sure we can add this to the experiment but certainly can add it to the Nightly changes we will be doing: https://bugzilla.mozilla.org/show_bug.cgi?id=1334696

aperkins81:
Until I saw this, I was nearly going to create an issue that adding a matching Tag to a bookmark does not auto-container bookmarks.  Glad to hear it's being looked at, would be a useful tool.

biva:
In addition, it would be great that the behavior takes the keyboard shortcuts into consideration, as suggested in https://github.com/mozilla/testpilot-containers/issues/119

klint:
This is definitely what I miss the most today with containers.
With the following default behaviour, IMHO: assuming you have a tab already opened in a certain container, left clicking from that tab on a bookmark that has no default container assigned (assuming that feature exists, of course ;)) should open the bookmark in the same container, either in the same tab or in a new tab, depending on the default setting for new tabs in Firefox pertaining to when new tabs are opened (see Tab Mix Plus settings).

MurzNN:
@klint Can you describe how to configure open new tab in same container (in Tab Mix Plus settings)? For me all bookmarks opens in new tab with no container, even if current tab have a container. But if I open bookmark in private tab, it opens new tab in private mode too.

klint:
@MurzNN Yes, that's the same for me. Sorry, my explanations were confusing. I have mentionned Tab Mix Plus just to say that my Firefox is configured to open any link or bookmark in a new tab, that's all.
But container-wise, it behaves exactly like you say. Which does not please me :)

MurzNN:
[...]

PR #1537, opened 2019-10-17
Adds a bookmark context menu for opening bookmarks in container tabs #323

kendallcorner:
Fix for #323 

# Steps to test
1. open bookmarks menu, sidebar, or library.
2. right click on a bookmark and select 'Open in container tab'

You can also right click a folder and open all the bookmarks in the same container tab from the folder.

One note: The context menu api does not allow a color to be set for the icon, so the only way I can see to colorize the icons in the context menu would be to make individual icons for each color, or I have seen an svg color filter, but we would still need one for each icon (and a filter for each color in use). I can do that if wanted.

# Screen capture
Peek 2019-10-17 15-12


kendallcorner:
@maxxcrawford

kendallcorner:
Thanks for the comments @Rob--W. I did see your addon, but I tried not to steal from it outright! But it sounds like you are ok with that, so I'll take a closer look at it.

I'll also look at optional_permissions.

src/js/background/assignManager.js
```diff
@@ -275,6 +282,19 @@ const assignManager = {
     }
   },
 
+  async _onClickedBookmark(info) {
+    let bookmarks = await browser.bookmarks.get(info.bookmarkId);
```
groovecoder: Could we do this with `const`?
```js
    async _getBookmarksFromInfo(info) {
      const bookmarkTreeNode = await browser.bookmarks.get(info.bookmarkId);
      const firstBookmark = bookmarkTreeNode[0];
      if (firstBookmark.type !== "folder") {
        return [firstBookmark];
      }
      return await browser.bookmarks.getChildren(firstBookmark.id);
    }

    const bookmarks = await _getBookmarksFromInfo(info);
    for (const bookmark of bookmarks) {
      browser.tabs.create({
        cookieStoreId: info.menuItemId,
        url: bookmark.url,
      });
    }
```

src/js/background/assignManager.js
```diff
@@ -221,7 +222,11 @@ const assignManager = {
 
   init() {
     browser.contextMenus.onClicked.addListener((info, tab) => {
-      this._onClickedHandler(info, tab);
+      if (tab) {
```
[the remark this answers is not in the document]
groovecoder: Could keep it 1 line like this?
```js
(info.bookmarkId) ? this._onClickedBookmark(info) : this._onClickedHandler(info, tab);
```

src/js/background/assignManager.js
```diff
@@ -442,6 +462,24 @@ const assignManager = {
...
+
+  async initBookmarksMenu() {
+    browser.contextMenus.create({
+      id: this.OPEN_IN_CONTAINER,
+      title: "Open Link in Container Tab",
+      contexts: ["bookmark"],
+    });
+
+    const identities = await browser.contextualIdentities.query({});
+    for (let identity of identities) { // eslint-disable-line prefer-const
```
groovecoder: I think we still prefer `const` in `for` loop variable declarations.

src/manifest.json
```diff
@@ -30,7 +30,8 @@
     "storage",
     "tabs",
     "webRequestBlocking",
-    "webRequest"
+    "webRequest",
+    "bookmarks"
```
groovecoder: Let's try to put this under `optional_permissions` and trigger a request for the permission at the right time. @Rob--W based on your experience with users of your add-on, can you think of a good time to prompt for this permission?

Rob--W: Usually, an [options page](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/user_interface/Options_pages) would be a good place for it. But this extension doesn't have one. The only UI surface that the add-on has is the browserAction popup, which doesn't look like a good place for this functionality.

Given the required effort (and cost of commitment to maintenance), I wonder whether it might be a better idea to not add the feature for now. Users can just install the other add-on if they want bookmark menu items.

kendallcorner: Rob, I think I'll still finish it, so that it's an applicable Outreachy application (assuming that's how this works). I guess after that it'll be up to the MAC team whether they want to fully implement it or not.

I was thinking I had two options here:

I can either add a button on the first page of the popup (like opposite the sort tabs button) to enable the bookmark context menu, which I'm not sure is a logical place for this.

Or

I can add an options page, which seems easy enough to do. After that if there's time, I could add a new intro panel to the popup that asks the user if they want to enable the feature (and submits the request).

Right now I have it working while checking if the permissions are enabled, I think I'm nearly done adding an options page, and that matches Rob's suggestion, so I think I'll keep working that way unless @groovecoder has a different suggestion.

Rob--W: The feature is not prominent enough to warrant an extra button in the popup (which currently looks relatively minimal & clean).

If you still want to add this feature somewhere, then a new options page would probably be the most fitting place.

kendallcorner: Newest commit adds the options page.

kendallcorner:
Latest commit shows the solution to the permissions problem with `optional_permissions` and an options page.

Second to latest commit updates the context menu when containers are updated, added, or removed.

Please let me know what you think! 

I still haven't tackled a few of @Rob--W 's comments including:

> * Some URLs cannot be opened by extensions, and would result in error messages in the console.
>
> * Bookmark folders can contain folders, and I guess that your implementation would result in the creation of tabs with the url "moz-extension://.../undefined" because those aren't filtered.
> * There are a couple of other minor edge cases that I didn't mention; I suggest to take a look at the implementation of my add-on that already implements this feature

maxxcrawford:
@kendallcorner Fixed your conflict on `src/manifest.json`!

kendallcorner:
New Options page and permissions request.

Peek 2019-10-24 14-13

kendallcorner:
Context menu updates when containers update: 
Peek 2019-10-24 15-14

src/js/background/assignManager.js
```diff
@@ -245,9 +240,48 @@ const assignManager = {
...
+  },
+
+  makeBookmarksMenu() {
     this.initBookmarksMenu();
+    browser.contextualIdentities.onCreated.addListener(this.contextualIdentCreated);
+    browser.contextualIdentities.onUpdated.addListener(this.contextualIdentUpdated);
+    browser.contextualIdentities.onRemoved.addListener(this.contextualIdentRemoved);
+  },
 
+  contextualIdentCreated(changeInfo) {
```
groovecoder: `contextualIdentityCreated`

src/js/background/assignManager.js
```diff
@@ -245,9 +240,48 @@ const assignManager = {
...
+    browser.contextMenus.create({
+      parentId: assignManager.OPEN_IN_CONTAINER,
+      id: changeInfo.contextualIdentity.cookieStoreId,
+      title: changeInfo.contextualIdentity.name,
+      icons: { "16": `img/usercontext.svg#${changeInfo.contextualIdentity.icon}` }
+    });
+  },
+
+  contextualIdentUpdated(changeInfo) {
```
groovecoder: `contextualIdentityUpdated`

src/js/background/assignManager.js
```diff
@@ -245,9 +240,48 @@ const assignManager = {
...
+    browser.contextMenus.update(changeInfo.contextualIdentity.cookieStoreId, {
+      title: changeInfo.contextualIdentity.name,
+      icons: { "16": `img/usercontext.svg#${changeInfo.contextualIdentity.icon}` }
+    });
+  },
+
+  contextualIdentRemoved(changeInfo) {
```
groovecoder: `contextualIdentityRemoved`

src/js/background/assignManager.js
```diff
@@ -282,11 +316,19 @@ const assignManager = {
...
-    let bookmarks = await browser.bookmarks.get(info.bookmarkId);
-    if (bookmarks[0].type === "folder") { 
-      bookmarks = await browser.bookmarks.getChildren(bookmarks[0].id);
+
+    async function _getBookmarksFromInfo(info) {
+      const bookmarkTreeNode = await browser.bookmarks.get(info.bookmarkId);
+      const firstBookmark = bookmarkTreeNode[0];
+      if (firstBookmark.type === "folder") {
+        return await browser.bookmarks.getChildren(firstBookmark.id);
+      } else {
```
groovecoder: You can drop the `else` logic branch here, because the preceding line will return out of the function early. So you can put the `return bookmarkTreeNode` after the `if (firstBookmark.type === "folder") {}` block.

groovecoder: I find that using `const` variables and keeping my logic branches to a minimum helps keep the code straight-forward.

src/options.html
```diff
@@ -0,0 +1,16 @@
...
+
+<html>
+  <head>
+    <meta charset="utf-8"/>
+  </head>
+
+  <body>
+    <form>
+      <input type="checkbox" id="bookmarksPermissions"/>
+      <label>Enable Bookmark Permissions</label>
```
groovecoder: Let's call this "Enable Bookmark Menus"? Since "permissions" is just the technical thing we need, and "menus" is the user-facing feature.

src/js/background/assignManager.js
```diff
@@ -245,9 +240,48 @@ const assignManager = {
         delete this.canceledRequests[options.tabId];
       }
     },{urls: ["<all_urls>"], types: ["main_frame"]});
+    this.getPermissions();
+  },
 
+  getPermissions() {
+    browser.permissions.getAll().
```
groovecoder: Can we use `async`/`await` here instead of `then()`?
```js
async getPermissions() {
  const {permissions} = await browser.permissions.getAll();
  permissions.includes("bookmarks") ? this.makeBookmarksMenu() : browser.contextMenus.remove(this.OPEN_IN_CONTAINER);
}
```

kendallcorner:
@groovecoder  I think I have everything now. The tests even pass "cleanly" now. The failed test with Travis CI is the timeout test. I'm not sure what the deal is with that. It was passing before I accidentally closed this issue above, and then failed when I reopened it.

src/js/background/assignManager.js
```diff
@@ -274,6 +306,38 @@ const assignManager = {
...
+
+    const bookmarks = await _getBookmarksFromInfo(info);
+    for (const bookmark of bookmarks) {
+      // Some checks on the urls from https://github.com/Rob--W/bookmark-container-tab/ thanks!
+      if ( !/^(javascript|place):/i.test(bookmark.url) && bookmark.type !== "folder") {
+        const openInReaderMode = bookmark.url.startsWith("about:reader");
+        if(openInReaderMode) {
+          try {
+            const parsed = new URL(bookmark.url);
+            bookmark.url = parsed.searchParams.get("url") + parsed.hash; // can't believe const lets me do this ...
```
groovecoder: image

src/js/background/assignManager.js
```diff
@@ -274,6 +306,38 @@ const assignManager = {
...
+      if (bookmarkTreeNode.type === "folder") {
+        return await browser.bookmarks.getChildren(bookmarkTreeNode.id);
+      }
+      return [bookmarkTreeNode];
+    }
+
+    const bookmarks = await _getBookmarksFromInfo(info);
+    for (const bookmark of bookmarks) {
+      // Some checks on the urls from https://github.com/Rob--W/bookmark-container-tab/ thanks!
+      if ( !/^(javascript|place):/i.test(bookmark.url) && bookmark.type !== "folder") {
```
groovecoder: Oh dang. This means nothing will happen when a user tries to open a `javascript:` or `place:` bookmark into a container, with no error message nor alt text telling them why they can't.

If we get bug reports about it, we could try moving this logic "up" into `init` before we add the listener to the context menu item.

src/js/background/assignManager.js
kendallcorner: haha, I did not mean to leave that in ... sorry!

src/options.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+
+<html>
+  <head>
+    <meta charset="utf-8"/>
```
Rob--W: Remove `/` in `/>` (also in the rest of this file). Self-closing tags are only used in XML and XHTML, but this is HTML5, where it does not serve any purpose.

src/options.html
```diff
@@ -0,0 +1,16 @@
...
+  </head>
+
+  <body>
+    <form>
+      <input type="checkbox" id="bookmarksPermissions"/>
+      <label>Enable Bookmark Menus</label>
```
Rob--W: For usability, it is generally recommended to either put the `<input>` inside the `<label>`, or to use `<label for="bookmarksPermission">` . Doing so enables the user to toggle the checkbox more easily, as they can then click on the label insted of trying to aim at the checkbox.

src/options.html
kendallcorner: Didn't know that! Thanks!

src/js/options.js
```diff
@@ -0,0 +1,41 @@
...
+    browser.permissions.request({permissions: ["bookmarks"]}).
+    then((response) => {
+      if (response) {
+        browser.runtime.sendMessage({ method: "resetBookmarksContext" });
+      } else {
+        checkbox.checked = false;
+      }
+    }).
+    catch((err) => {
+      return err.message;
```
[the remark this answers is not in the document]
kendallcorner: hah, this is only here because the linter was requiring it. Switching to async/await doesn't have this issue with the linter, and it seems like everyone agrees it is more straightforward, so it seems like that is the way to go.

kendallcorner:
@Rob--W thanks for the review! I'll try to take a look this week.

src/js/background/assignManager.js
```diff
@@ -274,6 +306,38 @@ const assignManager = {
...
+        const openInReaderMode = bookmark.url.startsWith("about:reader");
+        if(openInReaderMode) {
+          try {
+            const parsed = new URL(bookmark.url);
+            bookmark.url = parsed.searchParams.get("url") + parsed.hash; // can't believe const lets me do this ...
+          } catch (err) {
+            return err.message;
+          }
+        }
+        browser.tabs.create({
```
yellowwoods12: > Oh dang. This means nothing will happen when a user tries to open a `javascript:` or `place:` bookmark into a container, with no error message nor alt text telling them why they can't.
> 
> If we get bug reports about it, we could try moving this logic "up" into `init` before we add the listener to the context menu item.

For this we can just add a callback here. If the returned promise throws an error and a new tab cannot be created then generate an alert message for the same.

Issue #950, opened 2017-11-10
Undo close tab after contained tab, "duplicates" the tab


yosubis:
If I open a "sleeping" tab (from previous session) that is in a container and then undo close tab shortcut (ctrl+shift+T), a duplicate of this contained tab it opened. If I continue undo close tab, the same tab is duplicated again.

SoftVision-EmilPasca:
I have managed to reproduce the issue on Windows 10 x64 and Mac 10.12 with the latest Firefox release (56.0.2) and the latest Nightly (59.0a1-20171113100232) using the latest Multi-Account Containers (4.0.3).

Attached a screen recording of the issue:
![containers_duplicate_tab](https://user-images.githubusercontent.com/20083658/32728577-814c26e4-c889-11e7-94e6-f2cf41138238.gif)

kris7453:
Same problem. Now i have duplicated tabs in container and I can't show hidden duplicated tabs nor show visible tabs. When I have separated showed tabs by another container group, then when I click show, all duplicated tabs will open and when i close browser and open again then i have tripled tabs.
The problem appear  for newest firefox quantum 57.0(64bit)
![image](https://user-images.githubusercontent.com/20442936/32858268-4b024f6e-ca4b-11e7-8cd5-283a6df6a509.png)

TanviHacks:
Is wikipedia assigned to a container?

oksmelnik:
I'd like to work on this issue (outreachy applicant)

oksmelnik:
the problem is that opening recently closed container's tab triggers unhiding of all the other tabs of this container, including this particular tab again. Is containers tabs re-opening suppose to happen in this case, @groovecoder? If so, I'll try to find the way to not duplicate the tab.

oksmelnik:
please, check my PR here 
https://github.com/mozilla/multi-account-containers/pull/1540

groovecoder:
Fixed in #1540

PR #1540, opened 2019-10-19
#950: don't unhide duplicates of already opened tab

oksmelnik:
fix for https://github.com/mozilla/multi-account-containers/issues/950
Issue: If you open a tab, which was assigned to some containers and you have some hidden tabs from this container at the moment, all the hidden tabs will be opened again. 

Solution: check if the url of the hidden tab was already opened and if so skip it from re-opening to prevent duplicates.

### DIFF
--- a/src/js/background/backgroundLogic.js
+++ b/src/js/background/backgroundLogic.js
@@ -113,12 +113,12 @@ const backgroundLogic = {
     return list.concat(containerState.hiddenTabs);
   },
 
-  async unhideContainer(cookieStoreId, toHide) {
+  async unhideContainer(cookieStoreId, alreadyShowingUrl) {
     if (!this.unhideQueue.includes(cookieStoreId)) {
       this.unhideQueue.push(cookieStoreId);
       await this.showTabs({
         cookieStoreId,
-        toHide
+        alreadyShowingUrl
       });
       this.unhideQueue.splice(this.unhideQueue.indexOf(cookieStoreId), 1);
     }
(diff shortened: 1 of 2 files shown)


src/js/background/messageHandler.js
```diff
@@ -155,7 +155,24 @@ const messageHandler = {
...
+              throw e;
+            });
+
+            browser.tabs.onUpdated.removeListener(this.tabUpdateHandler);
+          }
+        };
+
+        // if it's a container tab wait for it to complete and
+        // unhide other tabs from this container
+        if (tab.cookieStoreId.startsWith("firefox-container")) {
```
groovecoder: Do we need this `if` check, or could we move this `addListener` into the above `if` that checks the `tab` is not `firefox-default` or `firefox-private` or a `moz-extension` url?

src/js/background/backgroundLogic.js
```diff
@@ -113,11 +113,12 @@ const backgroundLogic = {
     return list.concat(containerState.hiddenTabs);
   },
 
-  async unhideContainer(cookieStoreId) {
+  async unhideContainer(cookieStoreId, toHide) {
```
groovecoder: Can we change this variable name from `toHide` to `alreadyShowingUrl` to make it more explicit?

src/js/background/messageHandler.js
oksmelnik: yes, moving it makes perfect sense.  I'm not sure if 
`  firefox-default, firefox-private, nor our own confirm page loads `
would filter out all the non-container tabs, so added the additional check. Let me it you don't think it's necessary .

groovecoder:
Resolved a conflict but this looks good now. Thanks!

PR #1541, opened 2019-10-19
Fix #1041- [Feature request] Open previously hidden tabs in the background

### LINKED ISSUE #1041: [Feature request] Open previously hidden tabs in the background
When I open all tabs from a hidden container ("Show this container") all of the tabs are opened in the foreground. All tabs contents are loaded instead of loaded when I select them.

I would prefer if only the first selected tab is loaded and all the others are loaded in the background - i.e. not loaded. When I open a hidden container with e.g. 40 tabs  (previously hidden) Firefox gets really sluggish and even unresponsive.  
This behavior is even "worse" when I open a new tab which is automatically loaded in a specific currently hidden container (say Amazon for the shopping container). Next to this new tab I just started loading all of the hidden tabs - maybe this is the desired behavior, but I would like to be able to either not load them or load them in the background.

Maybe the new [`discard()`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/discard) feature maybe be leveraged to achieve this behavior?

Karskaya:
Add 'noload' property to openNewTab options, add 'discarded' parameter to browser.tabs.create. New tabs will be discarded in case they opened from showTabs

### DIFF
--- a/src/js/background/backgroundLogic.js
+++ b/src/js/background/backgroundLogic.js
@@ -55,6 +55,7 @@ const backgroundLogic = {
     let url = options.url || undefined;
     const userContextId = ("userContextId" in options) ? options.userContextId : 0;
     const active = ("nofocus" in options) ? options.nofocus : true;
+    const discarded = ("noload" in options) ? options.noload : false;
 
     const cookieStoreId = backgroundLogic.cookieStoreId(userContextId);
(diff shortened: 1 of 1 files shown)


Karskaya:
@maxxcrawford @groovecoder Could you please review this?

PR #1551, opened 2019-10-23
Ways to display complete name of the current tab

HagarMostafa:
This edit is for the popup title to appear in a tooltip as a full title.
Before the edit
image

After the edit
image


HagarMostafa:
@groovecoder what do you think?

Issue #1555, opened 2019-10-24
Make edit containers panel tabbable


dnahol:
### Actual behavior
Edit containers panel buttons not accessible via tab key
![EditContainersTabbing Issue](https://user-images.githubusercontent.com/2976514/67447445-b6c2e100-f5c8-11e9-944e-ae1edde1ea21.gif)

### Expected behavior
Edit containers panel buttons should be accessible via tab key

### Steps to reproduce
1. Open containers popup
2. Click Edit Containers button on bottom of main panel
3. Press tab to navigate through edit and delete buttons for every identity (It doesn't work)

### Notes
- Multi-Account Containers Version: Version 6.1.1, Last Updated: October 22, 2019
- Operating System + Version: Mac OS Sierra Version 10.13.6 
- Firefox Version: Firefox Quantum Version 69.0.3 (64-bit)

NoumanMasood:
Hi, I am outreachy intern. Can i work on this problem?

dnahol:
Hi @NoumanMasood I submitted a PR for this yesterday. I'd be happy to hear any comments you may have onboard!

NoumanMasood:
ok I need to do contribution for outreachy. Can you help me to guide which task should I do?

dnahol:
I hear you. I'm doing the same. As far as I can see, basically any open issue that isn't already assigned is fair game. There is a "Good First Bug" tag as well, but you don't have to limit yourself to that.

dnahol:
Oh and also some issues aren't assigned but they do have a PR on there already, like this one. You will see the PR mentioned like mine is above your comment.

NoumanMasood:
ok means that we can create issues ourselves too and then can work on it . right?

dnahol:
yep

grahamperrin:
`7.3.0`: **Manage Containers** appears to respond to intuitive use of arrow keys, and so on.

dannycolin:
Fixed in another bug.

PR #1556, opened 2019-10-24
Fixes Issue #1555 Make edit containers panel tabbable

dnahol:
Fixes Issue #1555 Make edit containers panel tabbable

## Before:
Screen recording of issue. Buttons not accessible via tabbing.
EditContainersTabbing Issue

## After:
Screen recording of fixed issue. Buttons now accessible via tabbing.
EditContainersTabbing PR

### DIFF
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -888,9 +888,12 @@ Logic.registerPanel(P_CONTAINERS_EDIT, {
           />
         </td>`;
       tr.querySelector(".container-name").textContent = identity.name;
-      tr.querySelector(".edit-container").setAttribute("title", `Edit ${identity.name} container`);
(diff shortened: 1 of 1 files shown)


maxxcrawford (changes requested):
@dnahol Thanks for working on this – tab support is really important. I found one issue – each "item" on the main panel has two buttons. 

1. The container name (e.g. "Work") which opens a new tab in that specific container
2. The arrow (`>`) which opens up that panel's settings/details. 

When tabbing through, it skips over this. Could you be sure that the tab index includes these secondary buttons? 

image

maxxcrawford:
Closing this PR as it applies to pre-7.0 [UI flow](https://github.com/mozilla/multi-account-containers/releases/tag/7.0.0).

PR #1565, opened 2019-10-30
fixed #1564: cleanup updates requested for bookmark context menu

### LINKED ISSUE #1564: Suggested updates from PR #1537 (bookmarks context menu)
Suggested edits from @Rob--W in [comment](https://github.com/mozilla/multi-account-containers/pull/1537#pullrequestreview-307579647) on PR #1537 

Related to #323

kendallcorner:
Suggested updates after merge from PR #1537 @Rob--W and @groovecoder 
Fix for #1564

### DIFF
--- a/src/js/background/assignManager.js
+++ b/src/js/background/assignManager.js
@@ -240,19 +240,26 @@ const assignManager = {
         delete this.canceledRequests[options.tabId];
       }
     },{urls: ["<all_urls>"], types: ["main_frame"]});
-    this.getPermissions();
+    this.resetBookmarksMenuItem();
   },
 
-  async getPermissions() {
-    const {permissions} = await browser.permissions.getAll();
-    permissions.includes("bookmarks") ? this.makeBookmarksMenu() : browser.contextMenus.remove(this.OPEN_IN_CONTAINER);
-  },
-
-  makeBookmarksMenu() {
-    this.initBookmarksMenu();
-    browser.contextualIdentities.onCreated.addListener(this.contextualIdentityCreated);
-    browser.contextualIdentities.onUpdated.addListener(this.contextualIdentityUpdated);
-    browser.contextualIdentities.onRemoved.addListener(this.contextualIdentityRemoved);
+  async resetBookmarksMenuItem() {
+    const hasPermission = await browser.permissions.contains({permissions: ["bookmarks"]});
+    if (this.hadBookmark === hasPermission) {
+      return;
+    }
+    this.hadBookmark = hasPermission;
+    if (hasPermission) {
+      this.initBookmarksMenu();
+      browser.contextualIdentities.onCreated.addListener(this.contextualIdentityCreated);
+      browser.contextualIdentities.onUpdated.addListener(this.contextualIdentityUpdated);
+      browser.contextualIdentities.onRemoved.addListener(this.contextualIdentityRemoved);
+    } else {
+      this.removeBookmarksMenu();
+      browser.contextualIdentities.onCreated.removeListener(this.contextualIdentityCreated);
+      browser.contextualIdentities.onUpdated.removeListener(this.contextualIdentityUpdated);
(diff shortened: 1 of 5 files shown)


src/js/options.js
```diff
@@ -1,39 +1,20 @@
...
-        browser.runtime.sendMessage({ method: "resetBookmarksContext" });
-      } else {
-        checkbox.checked = false;
-      }
-    }).
-    catch((err) => {
-      return err.message;
-    });
+    const response = await browser.permissions.request({permissions: ["bookmarks"]});
+    if (!response) { checkbox.checked = false; }
```
Rob--W: Rename `response` to `granted` to make it more obvious what it means, or use `if (!await ...) {` if you prefer to have smaller code.

Also add `return;`, to avoid sending a `resetBookmarksContext` event when nothing has changed.

src/js/background/assignManager.js
```diff
@@ -244,15 +244,18 @@ const assignManager = {
   },
 
   async getPermissions() {
```
Rob--W: Rename `getPermissions` so that its purpose is more obvious. E.g. to `resetBookmarksMenuItem`.

src/js/background/assignManager.js
```diff
@@ -244,15 +244,18 @@ const assignManager = {
...
-    permissions.includes("bookmarks") ? this.makeBookmarksMenu() : browser.contextMenus.remove(this.OPEN_IN_CONTAINER);
-  },
-
-  makeBookmarksMenu() {
-    this.initBookmarksMenu();
-    browser.contextualIdentities.onCreated.addListener(this.contextualIdentityCreated);
-    browser.contextualIdentities.onUpdated.addListener(this.contextualIdentityUpdated);
-    browser.contextualIdentities.onRemoved.addListener(this.contextualIdentityRemoved);
+    const hasPermission = await browser.permissions.contains({permissions: ["bookmarks"]});
+    if (hasPermission) {
```
Rob--W: To avoid the issue of attempting to add an already-added menu twice, remember whether you've already shown the menu, e.g. like this:

```javascript
if (this.hadBookmark === hasPermission) {
  return;
}
this.hadBookmark = hasPermission;
```

src/js/options.js
kendallcorner: fixed this and squashed.

src/js/background/assignManager.js
kendallcorner: and this

src/js/background/assignManager.js
kendallcorner: @Rob--W I'm not sure I see how the menu could be added twice. `getPermissions()`, or `resetBookmarksMenuItem()` now, only runs once when the containers extension loads and then when the checkbox is toggled for permissions in the options menu. When the containers are added/removed/updated, it only changes the single item that changed in the menu. Do you have an example of when this could happen?

Rob--W: This could have happened in your previous patch, where a `return` was missing.

This can still happen when a user opens `about:addons` plus the options page twice. The checkbox states are not updated when the permission is changed elsewhere, so when you select a checkbox in one options page, the event will fire in the background page. When you check the checkbox in the second options page, the event will fire once more.

src/js/options.js
```diff
@@ -1,39 +1,20 @@
...
-        browser.runtime.sendMessage({ method: "resetBookmarksContext" });
-      } else {
-        checkbox.checked = false;
-      }
-    }).
-    catch((err) => {
-      return err.message;
-    });
+    const granted = await browser.permissions.request({permissions: ["bookmarks"]});
+    if (!granted) { checkbox.checked = false; }
```
Rob--W: Not sure what happened, but the `return;` disappeared here.

It should be

```javascript
if (!granted) {
  checkbox.checked = false;
  return;
}
```

kendallcorner: There you go!

Issue #512, opened 2017-05-16
Implement full keyboard functionality


jonathanKingston:
#509 will add the ability for users to open the browser action however it doesn't add functionality to all screens.

@groovecoder agreed this can be a followup as it's likely to be a big one we perhaps could do in chunks.

The notable things that don't work are:

- [ ] Down, up arrows not on edit screen
- [ ] Tab key doesn't work in the panel for editing containers and making new containers?
- [ ] Right, Left doesn't work on entering the container right arrow, leaving it
- [ ] Down, up arrows on selecting existing tabs in the container view

jk3us:
I don't know if this is the right place, but I'd like to see keyboard shortcuts for things like cycling through containers with open tabs (and hiding all the others), and toggling between viewing "only container of current tab" and everything. This would give it some of the usefulness of the old tab groups.  Both of these are extensions of the existing show/hide container functionality.

jonathanKingston:
@ShivangiKakkar Yeah totally, smaller PRs are always easier to merge!

yellowwoods12:
Hi @groovecoder, @maxxcrawford  and @jonathanKingston, I would like to work on this one. Can you please assign it to me?

HagarMostafa:
Hello @groovecoder, @maxxcrawford, can you assign this issue to me if it is not already assigned?

PR #1575, opened 2019-11-03
Made Container-Info panel and Container-Edit panel tabbable

yellowwoods12:
In reference to issue: #512


kendallcorner:
Thanks for this contribution. However, we don’t plan to merge this, and we're about to update the UX, so we’re closing some open PRs.

Issue #1027, opened 2017-12-04
Confirming confirm page too fast results in null URL


jomo:
Steps to reproduce:

0. Configure a site to 'Always open in $container'
0. Open new tab with that site
0. Press enter while or immediately after confirm-page.html opens

confirm-page.html is now shown again, but with `null` URL and null Container:

![screenshot](https://user-images.githubusercontent.com/3585743/33577335-eeedd8ae-d942-11e7-8c07-53453d4d1328.png)

I didn't look at the code, but one possible cause might be that pressing the button too fast (before the actual URL is filled in?) redirects to empty string (or the like), thus redirecting to the confirm-page with no URL parameters.

sakshi97:
Hi @groovecoder. I would like to take up this issue if it's free.

bini11:
Hey @groovecoder I am an outreachy applicant and would like to work on this bug if it's free.

oksmelnik:
I'd like to work on this issue (outreachy applicant)

oksmelnik:
actually, can't reproduce it. I'm redirected from the confirm page to the page url when I press enter. Do you still have this bug, @jomo?

jomo:
Yes, it's still happening when you press enter fast enough. Using Version 6.1.1.

I typed the URL into the URL bar and kept pressing Enter really fast. I was able to reproduce it a couple times.

oksmelnik:
Thanks! See it now as well

grahamperrin:
> `null` URL 

Maybe fixed in a recent release of Firefox. Please review.

PR #1576, opened 2019-11-03
save confirm page url to the local storage as a backup #1027

oksmelnik:
fix for #1027 

for the cases when confirm page loads without search parameters and redirect url and container Id are lost, save confirmUrl to the local storage as a backup source of data.


jonathanKingston:
I don't think this is the correct solution, I think the issue is actually that the submit button is clickable before the data has loaded. Rather than persist the data I think we should delay any submission until it has been populated.

kendallcorner:
Thanks for this contribution. We're about to update the UX, so we’re closing some open PRs, and we don’t plan to merge this. Something along the lines of what @jonathanKingston described would be more likely to get merged. Thanks!

PR #1598, opened 2019-12-08
Add focus mode

samflores:
This adds a "focus mode" where a single container is shown at a time.

It includes a tabs counter and a new button on each container row to focus that container.
Unfocused

Once a container is focused it greys out other containers with open tabs.
Focused


kendallcorner:
Thanks for this contribution. However, we don’t plan to merge this, and we're about to update the UX, so we’re closing some open PRs. This might be good as a separate add-on.

Issue #1312, opened 2018-11-30
"Open link in new container tab -> No container" doesn't work if the link is to the current site which has been set "always open in" current container


garywill:
- Is "Firefox will: Never remember history" in the Firefox Preferences/Options under "Privacy & Security > History" selected?   no
- Are you using Firefox in a Private Window?  no
- Can you see a grayed out but ticked Checkbox with the description "Enable Container Tabs" in the Firefox Preferences/Options under "Tabs"?  yes
- Multi-Account Containers Version: 6.0.0
- Operating System + Version: Linux x64
- Firefox Version: 60
- Other installed Add-ons + Version + Enabled/Disabled-Status:

### Actual behavior

"Open link in new container tab -> No container" doesn't work if the link is to the current site which has been set "always open in" current container. Link will be opened in current container.

### Expected behavior

Link is opened in a no-container tab.

### Steps to reproduce

1. Create a container C, and set `www.site.com` always open in my container C.
2. In container C tab `www.site.com`, there's a link to `www.site.com/blah`. Right click that link and choose "Open link in new container tab -> No container"
3. A new tab is created, asking if I want to open in `Current container` or `container C`. From the color I see this asking tab is already in `container C`. So whatever I click, it open in `container C`

### Notes

If the link is to other site or to `other.site.com` or to `site.com`, no problem happens.

stoically:
This is working as expected. If you assign a website to "always open in" and "remember my choice", it should always open the site in that container, regardless of the "open in new container tab" selection.

cubetastic33:
Wouldn't it be helpful if it were possible to just open it in some other container for once, even though the _default_ option is a different container? That is, by default it opens in a specific container, but it can be opened in something else if needed.

garywill:
I found a workaround. Right-click the link and click "open in new window".

PR #1599, opened 2019-12-09
fix currentCookieStoreId left undefined #1312

garywill:
Background is https://github.com/mozilla/multi-account-containers/issues/1312

After debugging I found `browser.tabs.onCreated` is called twice.
The first call is ok. `tab.cookieStoreId = firefox-default`.
But the second call `tab.cookieStoreId = firefox-container-x`. The container id is not desired.

Then after digging the code, found in `reloadPageInContainer()` variable `let currentCookieStoreId;` is left undefined if `currentUserContextId` is false. That caused unwanted behavior.


jonathanKingston:
@garywill is this causing the code to do anything different or console errors?

My understanding is properties being undefined should use the `firefox-default` as that is how the extension APIs should work. (I would prefer we didn't pass undefined for the record, however I'm not sure if I prefer explicitly doing it either as this code might not account for future changes in Firefox either for example private browsing may eventually be supported).

src/js/background/assignManager.js
```diff
@@ -490,6 +490,9 @@ const assignManager = {
         currentCookieStoreId = backgroundLogic.cookieStoreId(currentUserContextId);
         confirmUrl += `&currentCookieStoreId=${currentCookieStoreId}`;
       }
+      else {
+        currentCookieStoreId = "firefox-default";
```
jonathanKingston: See my other comments however if we need this I would prefer all non test usage has a global.

garywill:
> is this causing the code to do anything different or console errors

Yes it is causing browser doing different. No console warning or error is showed. And there's some odd behavior still confusing me.

If leave the variable undefined :
- A tab with the comfirming page url is opened with `tab.cookieStoreId = firefox-default`. This seems ok.
- After 0.5 sec, the tab closes automatically. Then immediatly another tab with the comfirming page url is opened with `tab.cookieStoreId = firefox-container-x`. (as I described above)

It's the second one that is undesired and giving user unwanted container. I have no idea why this happend. These behaviors can be observed by adding `console.log` in `browser.tabs.onCreated.addListener` (I did it on Firefox 68).

garywill:
> private browsing may eventually be supported

Some time ago this addon could be set as allowed in private window. But in private window user can't open any container tab, browser console shows `Error: Illegal to set non-private cookieStoreId in a private window`.

After `"incognito": "not_allowed"` was added in `manifest.json`, no longer this addon ran in private window. I don't see any sense to use this addon in a private window. Is there any plan to do that? Or is there any thing I miss?

garywill:
According to Firefox source code, default container has userContextId=0

https://searchfox.org/mozilla-central/source/browser/base/content/utilityOverlay.js#776 in  function `createUserContextMenu()`
```js
 // If we are excluding a userContextId, we want to add a 'no-container' item.
  if (excludeUserContextId || showDefaultTab) {
    let menuitem = document.createXULElement("menuitem");
    menuitem.setAttribute("data-usercontextid", "0");
   .......
```

The background logic didn't associate 0 with default container. `userContextId` should return 0 when `cookieStoreId=firefox-default`
Incognito mode don't have `cookieStoreId`, `userContextId` can be `false`

After changing, it seems no reopening happends, so test fails. (I really don't understand why need reopening yet)

PR #1611, opened 2020-01-10
Sync Feature

kendallcorner:
To run tests, you may load it into a browser and run `browser.tests.runAll()` in the extension console.

That doesn't show you much, though. Best way to see it working is to open up two instances of nightly with the same fxa account, add the addon to each, mess around a bit, and then force your browsers to sync, which is an option under your fxa account icon.


test/features/sync.test.js
```diff
@@ -55,12 +54,55 @@ describe("Sync", () => {
...
     await nextTick();
 
     const sync = await background.browser.storage.sync.get();
+    console.log(await background.browser.storage.local.get());
 
     expect(sync.identities.length).to.equal(5);
+    console.log("!!!b");
+  });
+
+  it("should sync for the first time", async () => {
```
stoically: Just curious, I see you started working with the tests, but seem to have dropped it again and reverted back to testing directly in the browser - was there something not working as expected or had it a specific reason?

kendallcorner: Yes, I was having a really hard time with the test suite. I've never used Mocha, jsdom, etc before. I'm sure I could figure it out, but I didn't want to sink all of my time into it. There are tests in here, and I'll discuss with @groovecoder how to move forward from here.

kendallcorner: I think I can do most of what I'm doing in `test.js` with `background.browser.runtime.onMessage` (your suggestion) to actually get it into the test suite, but that seemed a little 'round-about'. That's probably the minimum I would do. I bet there's better ways to handle this, though. 

Do you know if we can do unit tests on a given function with mocha if this is not a node application? Everything I've seen makes it look pretty easy if the files are modules, but obviously that breaks the add on (oh, I tried it!). We discussed browserifying this next if that makes it easier/possible to do unit tests in Mocha.

stoically: Ah, yeah, testing with new tooling can be tricky and time-consuming indeed. The tests file for sync looks good tho, could possibly get merged into the actual tests afterwards.

> Do you know if we can do unit tests on a given function with mocha if this is not a node application?

The classes/functions could be exposed on `window` and then accessed with `background.window` in the tests. Not exactly pretty, but possibly easier than introducing bundling with webpack or similar.

If you're going to give testing with mocha another try, I'd be happy to help if I can, you could reach me on matrix.org (the chat platform Mozilla is moving to) at https://matrix.to/#/@stoically:matrix.org

groovecoder:
Whew, it works! I thought maybe I hit a/the bug where going from 73 to 74 deleted custom containers from my profile - but I don't think it was related to this change.

Continuing testing ...

src/js/background/sync.js
```diff
@@ -0,0 +1,483 @@
...
+      //   console.log("inLocal:", localStorage);
+      //   console.log("idents: ", await browser.contextualIdentities.query({}));
+      // }
+      console.log("Backed up!");
+      await sync.checkForListenersMaybeAdd();
+
+      async function updateSyncIdentities() {
+        const identities = await browser.contextualIdentities.query({});
+        await sync.storageArea.set({ identities });
+      }
```
stoically: The sync storage seems to get populated by directly asking the contextualIdentities API, this might lead to a case where containers are synchronized that didn't get created by MAC, but by other Container Add-ons - which could lead to unwanted and undefined behavior. E.g. my Temporary Containers Add-ons in Automatic Mode creates a lot of containers for its users, and those should never get synced. Would it maybe be possible to only sync containers created by MAC?

src/js/background/sync.js
```diff
@@ -0,0 +1,483 @@
...
+    await sync.storageArea.getIdentities();
+  const cookieStoreIDmap = 
+    await sync.storageArea.getCookieStoreIDMap();
+  // now compare all containers for matching names.
+  for (const syncIdentity of syncIdentities) {
+    syncIdentity.macAddonUUID = cookieStoreIDmap[syncIdentity.cookieStoreId];
+    if (syncIdentity.macAddonUUID){
+      const localMatch = localIdentities.find(
+        localIdentity => localIdentity.name === syncIdentity.name
+      );
```
stoically: This seems to cover my question already, but just to be sure: How will it be handled if the MAC sync update gets pushed out to users that have multiple devices? Does that effectively duplicate containers when the update arrives before the sync storage is updated or does this specific name check prevent that?

src/js/background/sync.js
kendallcorner: As I've seen, MAC doesn't care where the container was made. It'll still show up in the menu.  

I'm playing with your addon now, and I see how that could be messy. As soon as the containers are deleted, it will reset Sync (removing them), but it seems like it would be easy to get one stuck in one sync round, and then be syncing a temporary container around. It would also fill up the deleted uuid storage _fast_, which would waste our already pretty small sync storage.

src/js/background/sync.js
kendallcorner: This name check should prevent that. It will dupe them if the names are not identical, however, and it will just pick one for the site assignment, with preference to whichever container got into sync first.

stoically: Sounds good, thanks!

src/js/background/sync.js
stoically: Thanks for looking into it! I'm actually not sure how to sanely tackle that and still allow users to create containers outside of MAC (e.g. by using about:preferences#containers) :-/

stoically: Something that came to mind: Maybe a possible way would be if MAC provided an [API endpoint](https://github.com/mozilla/multi-account-containers/wiki/API) for excluding certain cookieStoreIds from sync?

src/js/background/identityState.js
```diff
@@ -26,16 +37,57 @@ const identityState = {
       });
     },
 
-    remove(cookieStoreId) {
+    async remove(cookieStoreId) {
       const storeKey = this.getContainerStoreKey(cookieStoreId);
-      return this.area.remove([storeKey]);
+      return await this.area.remove([storeKey]);
```
jonathanKingston: Just for reference, this has no impact. async functions just wrap the result in a promise. The return value is already a promise and is already awaited at the callsites.

src/js/background/identityState.js
```diff
@@ -54,9 +106,57 @@ const identityState = {
...
+  },
+
+  async lookupMACaddonUUID(cookieStoreId) {
+    // This stays a lookup, because if the cookieStoreId doesn't 
+    // exist, this.get() will create it, which is not what we want.
+    const cookieStoreIdKey = cookieStoreId.includes("firefox-container-") ? 
+      cookieStoreId : "firefox-container-" + cookieStoreId;
+    const macConfigs = await this.storageArea.area.get();
+    for(const configKey of Object.keys(macConfigs)) {
+      if (configKey === "identitiesState@@_" + cookieStoreIdKey) {
```
jonathanKingston: Can all the references to this key call into a reference?

But this one should be:
```suggestion
      if (configKey === this.getContainerStoreKey(cookieStoreIdKey)) {
```

src/js/background/sync.js
```diff
@@ -0,0 +1,461 @@
...
+      return await this.area.get();
+    },
+
+    async set(options) {
+      return await this.area.set(options);
+    },
+
+    async getDeletedIdentityList() {
+      const storedArray = await this.getStoredItem("deletedIdentityList");
+      return (storedArray) ?  storedArray : [];
```
jonathanKingston:
```suggestion
      return storedArray || [];
```

src/js/background/sync.js
jonathanKingston: I think just handling cleaning up of data is key here. I think if a container has been deleted it's historically been a very destructive action so it should likely delete all synced data across multiple machines etc.

src/js/background/sync.js
```diff
@@ -0,0 +1,461 @@
...
+
+    async dataIsReliable() {
+      const cookieStoreIDmap = await this.getCookieStoreIDMap();
+      const identities = await this.getIdentities();
+      for (const cookieStoreId of Object.keys(cookieStoreIDmap)) {
+        const match = identities.find(identity => 
+          identity.cookieStoreId === cookieStoreId
+        );
+        // if one has no match, this is bad data.
+        if (!match) return false;
```
jonathanKingston:
```suggestion
        if (!match) {
          return false;
        }
```
Shorthand if statements aren't used elsewhere (or at least weren't last time I checked)

src/js/background/sync.js
```diff
@@ -0,0 +1,461 @@
...
+      }
+      return !(Object.entries(cookieStoreIDmap).length === 0);
```
jonathanKingston: Was this supposed to be a `Array.reduce` or something? This has the impact of returning after the first iteration, that doesn't seem like the intent of the code. I suspect you meant to perhaps `return true;` at the end of the function and return false here if a certain condition is met perhaps to short circuit in the failure case.

src/js/background/sync.js
```diff
@@ -0,0 +1,461 @@
...
+      console.log("Initial State:", {syncInfo, localInfo, idents});
+    }
+    await sync.checkForListenersMaybeRemove();
+    console.log("runSync");
+
+    await identityState.storageArea.upgradeData();
+    await assignManager.storageArea.upgradeData();
+
+    const hasSyncStorage = await sync.storageArea.hasSyncStorage();
+    const dataIsReliable = await sync.storageArea.dataIsReliable();
```
jonathanKingston:
```suggestion
    if (hasSyncStorage && await sync.storageArea.dataIsReliable()) {
      await restore();
    }
```

src/js/background/sync.js
```diff
@@ -0,0 +1,461 @@
...
+      await browser.contextualIdentities.onUpdated.hasListener(listenerList[2])
+    );
+  }
+
+};
+
+sync.init();
+
+async function restore() {
+  console.log("restore");
```
jonathanKingston: Remove logs or make a debug function that is easy to disable etc.

src/js/background/sync.js
```diff
@@ -0,0 +1,461 @@
...
+      await assignManager.storageArea.set(siteKey, sites[siteKey]);
+    }
+  }
+}
+
+async function ifNamesMatch(syncIdentity, localMatch) {
+  // Sync is truth. if there is a match, compare data and update as needed
+  if (syncIdentity.color !== localMatch.color 
+      || syncIdentity.icon !== localMatch.icon) {
+    await browser.contextualIdentities.update(
```
jonathanKingston: The name of this method should match it's functionality as this is a destructive action.

jonathanKingston: I just realised the naming here, they should be named as if it could be used outside from the current callsite.

src/js/background/sync.js
stoically: If that works it'd be great, but @kendallcorner mentioned that there's a list of "deleted uuids" in the sync storage - which I guess makes sense since sync is eventually consistent and afaik there's no reliable way to determine whether all user devices synced the latest changes? And even if there were a way to do that, a user device could stay switched off for a long while, and Temporary Containers users tend to accumulate a lot of cookieStoreIds real quick and most likely exceeding the 100KB sync storage limit.

Other escape-hatches for TC might be: a way to completely disable the sync feature in MAC - preferably also with an API endpoint, so it could be done automatically.

src/js/background/identityState.js
```diff
@@ -54,9 +106,57 @@ const identityState = {
...
   },
 
+  async updateUUID(cookieStoreId, uuid) {
+    if (cookieStoreId && uuid) {
+      const containerState = await this.storageArea.get(cookieStoreId);
+      containerState.macAddonUUID = uuid;
+      await this.storageArea.set(cookieStoreId, containerState);
+      return uuid;
+    } 
+    throw new Error ("cookieStoreId or uuid missing");
```
jonathanKingston: I personally would make this the other way around:
```
if (!cookieStoreId || !uuid) {
  throw new Error ("cookieStoreId or uuid missing");
}
```
It has the benefit of reducing nesting in functions returning earlier like that.

src/js/background/sync.js
```diff
@@ -0,0 +1,461 @@
...
+        continue;
+      }
+    } catch (error) {
+      // this is probably old or incorrect site info in Sync
+      // skip and move on.
+    }
+  }
+}
+
+async function setAssignmentWithUUID (assignedSite, urlKey) {
```
jonathanKingston:
```suggestion
async function setAssignmentWithUUID(assignedSite, urlKey) {
```

src/js/popup.js
```diff
@@ -352,10 +352,13 @@ const Logic = {
   },
 
   getAssignmentObjectByContainer(userContextId) {
-    return browser.runtime.sendMessage({
-      method: "getAssignmentObjectByContainer",
-      message: { userContextId }
-    });
+    if (userContextId) {
```
jonathanKingston:
```suggestion
    if (!userContextId) {
      return {};
    }
```
Returning early reduces nesting and also in this case makes the 'diff noise' less too :)

src/js/background/assignManager.js
```diff
@@ -54,51 +59,99 @@ const assignManager = {
...
         });
       }
-      return this.area.set({
+      // eslint-disable-next-line require-atomic-updates
+      data.identityMacAddonUUID =
+        await identityState.lookupMACaddonUUID(data.userContextId);
+      await this.area.set({
         [siteStoreKey]: data
       });
+      if (backup) await sync.storageArea.backup({undelete: siteStoreKey});
```
jonathanKingston:
```suggestion
      if (backup) {
        await sync.storageArea.backup({undelete: siteStoreKey});
      }
```

src/js/background/sync.js
```diff
@@ -0,0 +1,461 @@
...
+        await removeFromDeletedSitesList(options.undelete);
+      // if (SYNC_DEBUG) {
+      //   const storage = await sync.storageArea.get();
+      //   console.log("inSync: ", storage);
+      //   const localStorage = await browser.storage.local.get();
+      //   console.log("inLocal:", localStorage);
+      //   console.log("idents: ", await browser.contextualIdentities.query({}));
+      // }
+      console.log("Backed up!");
+      await sync.checkForListenersMaybeAdd();
```
jonathanKingston: Do all of these await's need to be done in order? If not push them onto an array and `await Promise.all(arrayname);` whenever you need to ensure it's happened.

src/js/background/index.html
```diff
@@ -14,9 +14,11 @@
     ]
 -->
     <script type="text/javascript" src="backgroundLogic.js"></script>
+    <script type="text/javascript" src="sync.js"></script>
     <script type="text/javascript" src="assignManager.js"></script>
     <script type="text/javascript" src="badge.js"></script>
     <script type="text/javascript" src="identityState.js"></script>
     <script type="text/javascript" src="messageHandler.js"></script>
+    <script type="text/javascript" src="test.js"></script>
```
jonathanKingston: Ideally test files shouldn't ever be included in the extension output.

src/js/background/sync.js
```diff
@@ -0,0 +1,461 @@
...
+
+async function setAssignmentWithUUID (assignedSite, urlKey) {
+  const uuid = assignedSite.identityMacAddonUUID;
+  const cookieStoreId = await identityState.lookupCookieStoreId(uuid);
+  if (cookieStoreId) {
+    // eslint-disable-next-line require-atomic-updates
+    assignedSite.userContextId = cookieStoreId
+      .replace(/^firefox-container-/, "");
+    await assignManager.storageArea.set(
+      urlKey.replace(/^siteContainerMap@@_/, "https://"),
```
jonathanKingston: This replacement to `https://` is done in numerous places, I still am not sure if it's correct however it should probably be moved into it's own function.

blackjackshellac: The regular expressions could be stored as constants, but otherwise I don't see any real use for a separate function call here.

src/js/background/sync.js
jonathanKingston: I would suggest we give each MAC install it's own UUID then store that in the sync data as the creating install id rather than having a UUID for each container.
This would radically reduce storage cost also.

So a containers unique ID would become `${creatingInstallId}_${userContextId}` where `creatingInstallId` would be an array lookup of UUID's. So: `MacUUIDs: ['uuid1', 'uuid2']` and then indexes 1 and 2 would be used in the `creatingInstallId` rather than the much larger UUID.

That way we can also store in sync indexes of who has had the data synced and mark if it's needing deletion, once all Id's are removed from the 'seen' list we can delete it entirely from the syncing data.

src/js/background/sync.js
jonathanKingston: My rationales:
- Use of a well worded function name can help explain what the code is doing
- Reduce the copy and paste issues
- Simplify the review process if said code ever needs to account for different keys etc
- The current `storageArea` architecture is supposed to be isolated to code using it, even within it's own component, how it gets stored should only be understood by the storageArea functions such that there is separation of concerns.

Other parts of this code certainly don't abide to the last principal.

src/js/background/sync.js
stoically: That's a neat idea. I think that could work indeed.

kendallcorner: On install of MAC add uuid for install
Container and SA has been synced with install uuid

src/js/background/identityState.js
kendallcorner: Thanks. Didn't know this!

src/js/background/sync.js
kendallcorner: Data restructure removed `dataIsReliable`

src/js/background/sync.js
kendallcorner: Removed when data was restructured

src/js/background/index.html
kendallcorner: Working on that after/during QA

src/js/background/sync.js
kendallcorner: Halfway through I chanced `assignManager.getSiteStoreKey()` to take the url or the original siteStoreKey, so probably not needed anymore.

src/js/background/sync.js
kendallcorner: @jonathanKingston and I discussed, and I've added tracking of identities and site assignments on each install, so when something has been removed (or never existed) on every install, it is removed from the deleted list. This should work for this case.

src/js/background/identityState.js
```diff
@@ -11,12 +11,23 @@ const identityState = {
       const storeKey = this.getContainerStoreKey(cookieStoreId);
       const storageResponse = await this.area.get([storeKey]);
       if (storageResponse && storeKey in storageResponse) {
+        if (!storageResponse[storeKey].macAddonUUID){
+          storageResponse[storeKey].macAddonUUID = uuidv4();
+          await this.set(cookieStoreId, storageResponse[cookieStoreId]);
```
stoically: This tries to access `cookieStoreId` in `storageResponse`, but the response contains the `storeKey` at this point.

src/js/background/index.html
stoically: Opened https://github.com/kendallcorner/multi-account-containers/pull/1 to help with migrating the tests into the suite.

kendallcorner: Thanks!

src/js/background/identityState.js
kendallcorner: Good catch.

README.md
```diff
@@ -19,7 +19,7 @@ For more info, see:
 ## Development
 
 1. `npm install`
-2. `./node_modules/.bin/web-ext run -s src/`
+2. `./node_modules/web-ext/bin/web-ext run -s src/`
```
stoically: Could be written as `npx web-ext run -s src` as well, or moved into a package script. Directly linking to the script inside the module could fail in the future, since its location might change.

test/features/sync.test.js
```diff
@@ -0,0 +1,465 @@
...
+
+    const identities = await this.webExt.browser.contextualIdentities.query({});
+
+    identities.should.have.lengthOf(7, "There should be 7 identity entries");
+
+    Object.keys(getAssignedSites).should.have.lengthOf(5, "There should be 5 identity entries");
+
+    const personalContainer = 
+      this.syncHelper.lookupIdentityBy(identities, {name: "Personal"});
+    (personalContainer.color === "red").should.be.true;
```
stoically: `personalContainer.color.should.equal("red")` would work too.

groovecoder:
We published the announcement blog post so the link from the release notes wouldn't 404. So, I'm going to merge this and prepare the release. Let's fix up @stoically's comments in a follow-up.

d98ama:
Ok, so this is probably not the right place to ask this, but I couldn't figure out where, so I'll give it a shot...

I'm running my own Sync Server by pointing the identity.sync.tokenserver.uri to my own server (so, only the sync part, I'm using Mozillas account servers). 
Will the settings for multi-account-containers be stored on my sync server, together with synced tabs, history, bookmarks etc? Or will it be stored on servers belonging to Mozilla?

kendallcorner:
It uses what Firefox uses for Sync, so assuming your server up to date and allows addons to sync data, it should work fine and that data will end up on your server. Otherwise, I think it just won't sync.

hub2git:
Dear Firefox devs, 

Thank you very much for implementing syncing for such a useful extension.

Could we see other extensions also being given the ability to sync between computers?

cadeyrn:
> Could we see other extensions also being given the ability/authorization to sync between computers?

(I am not a Mozilla developer)

Other developers of Firefox extensions _have_ the ability to sync data, they just have to implement such a feature.

See https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/sync.

hub2git:
> 
> 
> > Could we see other extensions also being given the ability/authorization to sync between computers?
> 
> (I am not a Mozilla developer)
> 
> Other developers of Firefox extensions _have_ the ability to sync data, they just have to implement such a feature.
> 
> See https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/sync.
cadeyrn,

thanks so much for your reply. I shouldn't have addressed the Fox devs only. 
I plan to contact all the extension developers to add sync to their addons.

PR #1613, opened 2020-01-14
Recording - automatically add sites to containers as you browse

mckenfra:
Currently it's impossible to add an assignment if the URL immediately redirects. E.g. if you enter `google.com` in the address bar, it will immediately redirect to `www.google.com`. If you then open the MAC popup and click on `Always open in ...`, the assignment for `www.google.com` will be created, but not for `google.com`.

This pull request adds the ability to record assignments automatically as you browse. So in the example above, it will automatically add both `google.com` and `www.google.com`. See screencast below.

This pull request also improves the MAC messages that popup over your browser window, e.g. when you click `Always open in ...`. Previously the messages sometimes were mis-positioned or did not appear. Now the messages appear correctly, and multiple simultaneous messages stack with a smooth animation. See screencast below.

Unit tests included.

MAC

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "stylelint-config-standard": "^16.0.0",
     "stylelint-order": "^0.3.0",
     "web-ext": "^2.9.3",
-    "webextensions-jsdom": "^1.1.0"
+    "webextensions-jsdom": "^1.1.2"
   },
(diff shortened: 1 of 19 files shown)


jonathanKingston:
@TanviHacks I really think this is a pretty cool feature. How are we handling new features like this now? Do they have to go through design?
I'm happy to code review if we want to merge this.

PR #1627, opened 2020-02-01
Add feature to edit websites that open automatically in a certain container

anirudhb:
## Changes
* "Sites assigned to open in this container" is always visible, even if there are no sites assigned. However, this does not show when creating a new container.
* Added "+" icon on the side of the title, which creates a new input to add a website.

## Behavior
If the user types the URL "cnn.com", `https://` is added, but not `www`
If the user types the URL "http://cnn.com", nothing is added
If the user types the URL "www.cnn.com", `https://` is added, but not `www`
If the user types the URL "https://cnn.com", nothing is added

This was created out of personal need, but I feel that it could be useful to everyone to have. This is especially useful when the URL you want to add instantly redirects to another URL when it is typed in the omnibox (which you cannot do with the checkbox currently.)

I notice that #1613 also implements similar features, but takes a different approach. This is easier for power users because you do not have to _open_ the website in the container.

### DIFF
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -975,6 +975,30 @@ Logic.registerPanel(P_CONTAINER_EDIT, {
       Logic.showPreviousPanel();
     });
 
+    Logic.addEnterHandler(document.querySelector("#website-add-link"), () => {
+      const sitesList = document.querySelector(".assigned-sites-list");
+      const trElement = document.createElement("div");
+      trElement.innerHTML = escaped`
+      <div class="truncated-text hostname">
+        <input id="site-editor" />
+      </div>`;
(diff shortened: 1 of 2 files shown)


dannycolin:
First of all, thank you for your contribution. Unfortunately, I'll have to close this issue as it needs to be thoughtfully planned before working on the code. By this, I mean that there's UX/design decisions that needs to be discussed and made.

On top of that, there's some early discussion to move the site assignments feature in Multi-Account Containers to a separated WebExtension API. Depending on the outcome, it could drastically change our approach on the implementation of this feature. When I have more information on this, I'll make an announcement on your Matrix room at https://matrix.to/#/#containers:mozilla.org.

PR #1686, opened 2020-03-15
Fixes #1682 'open in current container button' not working

### LINKED ISSUE #1682: _Open in Current Container_ button stopped working
The "_Open in Current Container_" button stopped working for me, where I expect the tab to open without any special container, i.e. the "_Default_" container.

The same if I try the "Open link in Container..." context menu, which brings me to this page: 

image

"_Open in Shopping_" is a no-op. It always picks the assigned colored container for the domain.

Addon:
Version: 6.2.3
Last Updated: March 2, 2020

Firefox: 75.0b3 (64-bit) Windows 10

DeadlySurprise:
Fixes the 'open in current container button' on the confirmation page. It would always default to using the assigned container instead.  

I am not sure what exactly broke it, but [Event.explicitOriginalTarget](https://developer.mozilla.org/en-US/docs/Web/API/Event/explicitOriginalTarget) is non standard according to MDN. So assigning to separate handlers should be preferable anyway imo.  

I looked into writing test cases for this since according to #1682 this broke before but I could not get the confirmation page working in jsdom.

### DIFF
--- a/src/js/confirm-page.js
+++ b/src/js/confirm-page.js
@@ -17,25 +17,14 @@ async function load() {
     const currentContainer = await browser.contextualIdentities.get(currentCookieStoreId);
     document.getElementById("current-container-name").textContent = currentContainer.name;
   }
-
-  document.getElementById("redirect-form").addEventListener("submit", (e) => {
+  document.getElementById("deny").addEventListener("click", (e) => {
     e.preventDefault();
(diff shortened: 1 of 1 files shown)


groovecoder:
Thanks! This is a particularly workflow-intrusive bug. :( I'll plan to release this fix on Monday.

Issue #1607, opened 2019-12-19
Feature Request: Reorganize windows (not just tabs) by category


jeremymeyers:
I am a FF user who often has multiple windows with a large amount of tabs per-window. I will usually try to organize myself by keeping websites of the same category in the same window instance (all music sites together, all shopping, all social media, etc).

It would be useful to me to be able to use the categorizing tabs feature to sort and move tabs that are assigned to the same category into the same 'window' instance of FF (as FF defines windows)

Use Case:

BEFORE

Window 1
  - Discogs.com (music category)
  - allmusic.com (music category)
   - facebook.com (fb category)
  - amazon.com (shopping category)

Window 2:
  - amazon.com (shopping category)
  - amazon.com (shopping category)
  - amazon.com (shopping category)
  - ableton.com (music category)
  - Facebook.com (fb category)
  - github.com (no category assigned)

AFTER
Window 1 (all music tabs)
  - Discogs.com
  - allmusic.com
  - ableton.com

Window 2 (All shopping tabs)
  - amazon.com (shopping category)
  - amazon.com (shopping category)
  - amazon.com (shopping category)
  - amazon.com (shopping category)

Window 3 (all FB tabs)
  - Facebook.com (fb category)
  - Facebook.com (fb category)

Window 4 (uncategorized tabs)
  - github.com (no category assigned)
   - (in this arrangement, tabs from the same domain would be grouped together, and grouping priority would be given to tabs that were previously in the same window)

My suggestion would be to be able to do this on a per-window basis, and executing this action on multiple windows would sort things into the same window categories (rather than creating entirely new windows)

falolaf:
I believe that you can do this with Simple Tab Groups(STG) addon:
https://addons.mozilla.org/en-US/firefox/addon/simple-tab-groups/

It is possible to catch container tabs and put into any tab group. You can have a tab group per window if you wish. You can also catch more than one container type into same tab group.

I use STG but not in that way so I can't guarantee it works as you wish but it might be worth a try.

richardkmichael:
FWIW and related:

Multi-container tabs is able to move all container-specific tabs from _a single window_ to a new window.  Click the extension icon in the toolbar, expand a container's options via ">", then choose "Move tabs to a new window".

This feature might be accomplished by enhancing the current functionality to be multi-window aware and adding "Move all tabs to this window" (since the OP does not want a new window to be created).

QiiLin:
[...]

PR #1702, opened 2020-04-07
Issue 1607 Feature Request: Reorganize windows (not just tabs) by category

QiiLin:
Fixes the issues #1607 Feature Request: Reorganize windows (not just tabs) by category

**Implemented** 
- Added "Sort Tabs by Windows" button
- Added background logic for the Sort Tabs by Windows button
- Manually Test the features include edge case like (One window with different containers), (Multiple windows and different containers), (One window with different containers), (One window and one container).

**This is how the feature looks like now:**

### DIFF
--- a/src/css/popup.css
+++ b/src/css/popup.css
@@ -486,28 +486,28 @@ manage things like container crud */
 
 .container-panel-controls {
   display: flex;
-  justify-content: flex-end;
   margin-block-end: var(--block-line-space-size);
   margin-block-start: var(--block-line-space-size);
-  margin-inline-end: var(--inline-item-space-size);
-  margin-inline-start: var(--inline-item-space-size);
 }
 
-#container-panel #sort-containers-link {
+#container-panel .sort-link {
   align-items: center;
   block-size: var(--block-url-label-size);
   border: 1px solid #d8d8d8;
   border-radius: var(--small-radius);
   color: var(--title-text-color);
   display: flex;
+  flex-grow: 1;
   font-size: var(--small-text-size);
   inline-size: var(--inline-button-size);
   justify-content: center;
+  margin-left: var(--block-line-space-size);
+  margin-right: var(--block-line-space-size);
   text-decoration: none;
 }
 
-#container-panel #sort-containers-link:hover,
-#container-panel #sort-containers-link:focus {
+#container-panel .sort-link:hover,
+#container-panel .sort-link:focus {
   background: #f2f2f2;
 }
 
--- a/src/js/background/backgroundLogic.js
+++ b/src/js/background/backgroundLogic.js
@@ -238,7 +238,7 @@ const backgroundLogic = {
   },
 
   async sortTabsByWindow() {
-    let windows = await browser.windows.getAll();
+    const windows = await browser.windows.getAll();
     let containers = new Set();
 
     // loop through each tab to find all active containers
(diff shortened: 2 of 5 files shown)


jeremymeyers:
A few things

1) I feel like having one click make this big of a change to the users organization will necessitate an 'undo' feature
2) I've been trying to figure out a succinct label for what function the button provides and having trouble. Maybe something like 'One Window Per Category View'? or maybe its a toggle between that language and 'multiple categories per window'?  Do users generally understand a 'window' to be what devs understand it to be? (i know its in the menu language)

dannycolin:
> I feel like having one click make this big of a change to the users organization will necessitate an 'undo' feature

An undo feature might complexify a lot more the code and there's little bandwidth to review patches. We could use a confirmation dialog as an alternative. That's what currently used when you delete a container.

> I've been trying to figure out a succinct label for what function the button provides and having trouble.

I think "Sort Windows by Containers" should be good enough. We can add information in both the documentation and the confirmation dialog to let the user know about what's going to happen especially if there's new windows that are going to be created by this action.

@QiiLin is it still something you'd like to work on (rebase on top of the main branch)?

jeremymeyers:
I would just want to make sure to account for what would happen if the user takes an action while the windows are reorganizing. 

- is input disabled?
- what if they quit the app mid process (through the windows taskbar for example) or start closing tabs? 
- if a tab has a JavaScript that somehow interrupts the process, will anything be lost?


Basically for me the worst case is
- click ok to reorganize
- ???
- multiple windows and/or open tabs are unrecoverablly lost mid process with no "revert" functionally

dannycolin:
> is input disabled?

What do you mean?

> what if they quit the app mid process (through the windows taskbar for example) or start closing tabs?
> if a tab has a JavaScript that somehow interrupts the process, will anything be lost?

Basically the same as if you use do:
1. Open context menu by right-cliking on a tab
2. Move Tabs
3. Move to a new window

Firefox will try to restore all tabs and windows unless you have explicitly disabled the restore session feature. In very rare cases, a session can be corrupted and tabs/windows can be lost but that'd be an issue with the browser itself and not the addon.

dannycolin:
Closing since I didn't get an answer from the PR author.

The original bug report will stay open because this is still something we'd be open to discuss and implement.

jeremymeyers:
> > is input disabled?
> 
> What do you mean?
While the app is in the process of reorganizing tabs into various windows, is it technically possible for a user to enter a keystroke or command that could adversely affect that process? What if they were holding control+f4, for example?
> 
> > what if they quit the app mid process (through the windows taskbar for example) or start closing tabs?
> > if a tab has a JavaScript that somehow interrupts the process, will anything be lost?
> 
> Basically the same as if you use do:
> 
> 1. Shift + Left click to select multiple tabs
> 2. Open context menu by right-cliking on a tab
> 3. Move Tabs
> 4. Move to a new window
> 
> Firefox will try to restore all tabs and windows unless you have explicitly disabled the restore session feature. In very rare cases, a session can be corrupted and tabs/windows can be lost but that'd be an issue within the browser itself and not the addon.

I guess i'm just thinking in terms of the cycles it might take for this to happen and what a user could do to mess things up.  I will often be in a situation where i have 10-11 open windows with 15-20 tabs open per-window.  I would imagine "reorganize into windows" might take upwards of a few seconds to execute and my concern would be around something happening in those few seconds to either leave the process half-completed (with no way to revert back to the original configuration) or to have tabs or windows lost.

PR #1728, opened 2020-05-01
Changed IRC to Matrix

Undoundoundo:
replaced reference to IRC with reference to Matrix chat
not sure if that should be offered as a link via Riot for convenience, just making sure this PR exists so it gets updated either way :)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 Everyone is welcome to contribute to containers. Reach out to team members if you have questions:
 
-- IRC: #containers on irc.mozilla.org
+- Matrix chat: #containers:mozilla.org
 - Email: containers@mozilla.com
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -2,7 +2,7 @@
 
 Everyone is welcome to contribute to containers. Reach out to team members if you have questions:
 
-- IRC: #containers on irc.mozilla.org
+- Matrix chat: #containers:mozilla.org
```
maxxcrawford, line 5: Realized we can/should hyperlink to this better. Could you include this link? 

```suggestion
- Matrix chat: [#containers:mozilla.org](https://matrix.to/#/#containers:mozilla.org)
```

maxxcrawford (changes requested):

PR #1756, opened 2020-05-28
Show fence icon in popup

nathanhinchey:
Fixes issue where Facebook container has no icon in popup. Also allows user to select fence icon for other containers.

Fixes https://github.com/mozilla/contain-facebook/issues/491
Relates to https://github.com/mozilla/contain-facebook/issues/205


groovecoder:
Yeah, that's fine. We landed the fence icon into central so it makes sense for the add-on to surface it up as an icon option. Thanks!.

Issue #1427, opened 2019-05-26
Ability to backup config to a file


skank:
Recent events has caused users to lose their configs.
The config is rather cumbersome with a lot of clicking etc.

Ideally we should be able to export settings to a file.
As an example uBlock & uMatrix has this very essential feature.

julien-garrigue:
Ah yes, that would be very very useful to save the config somewhere. Thank you

cdev45:
Hi Team. I LOVE this Firefox Add-on, however we need the ability to backup our settings during device formats, device transfers, etc. Starting from scratch is a nightmare. Thank you!

kendallcorner:
Linking to the original #1282 , but I'm leaving this open since a PR is submitted on it.

dannycolin:
Duplicate of #1282

Issue #1282, opened 2018-10-04
Feature request: Export and Import containers & settings


rm-root:
Hi, will you develop a way to export and import containers and settings? At the moment I must setup a lot of containers every time I use conteiners on a new device. Thank you.

IOAyman:
That would be great!

alex3kov:
Please do this. Preferably export everything (containers and URLs assigned to them) to one file - json or whatever.

pierlauro:
Waiting for an eventual official support, I created a script to export and import containers among different machines.

https://github.com/pierlauro/ffcontainers

MathieuDR:
> 
> 
> Waiting for an eventual official support, I created a script to export and import containers among different machines.
> 
> https://github.com/pierlauro/ffcontainers

Sadly doesn't work for me. But I used it to manually export it, Lets see if I can import it as well.
Thanks!

pierlauro:
> 
> 
> > Waiting for an eventual official support, I created a script to export and import containers among different machines.
> > https://github.com/pierlauro/ffcontainers
> 
> Sadly doesn't work for me. But I used it to manually export it, Lets see if I can import it as well.
> Thanks!

Please, open an issue on [ffcontainers' issue tracker](https://github.com/pierlauro/ffcontainers/issues) detailing the problem so I will be able to fix it.

throwaway167:
Adding an import & export capability for Containers, Settings, & Container's URL lists would be a TREMENDOUS asset. — As someone who uses profiles across multiple PCs, and occasionally has to delete & recreate FF profiles. . . needing to **manually recreate all your containers and their associated URLs** can be _extremely time-consuming_ and a complete pain. 

Would be MUCH appreciated, thank you!

avliet:
Import/Export would make life a LOT easier when upgrading Firefox on my Debian hosts.  Right now every time I upgrade Firefox, I lose all my containers, sessions, everything. Not this this project's fault, but REALLY annoying.

drod3763:
Same right here especially now that sync isn't working for some people. - myself included. This would be a good workaround.

moocan:
Hello,

I upvote for this too.
It would be very useful to have an `export/import settings` feature using a file in `Multi-Account Containers` like some add-ons allow you to do it already.

[...]

PR #1767, opened 2020-06-15
Support managed storage

komachi:
Allows to use [managed storage](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/managed) to pre-define hostnames. For example, you can save this file as `~/.mozilla/managed-storage/@testpilot-containers.json`

```
{
  "name": "@testpilot-containers",
  "description": "ignored",
  "type": "storage",
  "data": {
    "siteContainerMap@@_en.wikipedia.org": {
      "neverAsk": true,
      "userContextId": "2"
    }
  }
}
```

And then en.wikipedia.org would be opened in container with id "2".

Not a fix, but related to: #1282 #1427

### DIFF
--- a/src/js/background/assignManager.js
+++ b/src/js/background/assignManager.js
@@ -10,14 +10,15 @@ window.assignManager = {
     exemptedTabs: {},
 
     getSiteStoreKey(pageUrlorUrlKey) {
-      if (pageUrlorUrlKey.includes("siteContainerMap@@_")) return pageUrlorUrlKey;
-      const url = new window.URL(pageUrlorUrlKey);
       const storagePrefix = "siteContainerMap@@_";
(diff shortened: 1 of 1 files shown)


dannycolin:
First of all, thanks for contributing. I'm sorry you didn't get an answer earlier! I'm also a volunteer like you and got only recently more involved on Multi-Account Container. 

I can't give a timeline nor promise anything but I'm currently working on an API proposal to move the website assignments to Firefox itself. That means supporting the managed storage isn't something we're planning to support directly in Multi-Account Containers.

I will then close this PR but if you have any other feature/bug you'd like to work on, please feel free to get in touch with me directly. You can contact me on our chat platform at https://chat.mozilla.org. There are instructions on how to connect to it on https://wiki.mozilla.org/Matrix.

PR #1815, opened 2020-07-15
Site isolation saved changes do not affect other containers

### LINKED ISSUE #1781: "Limit to designated sites" option not saving changes correctly
- Multi-Account Containers Version: 7.0.0
- Operating System + Version: Arch Linux
- Firefox Version: 79.0b5

### Actual behavior
1) Go to  "Manage containers"
2) For each container, check "Limit to designated sites" to ON.
3) Close popup and reopen "Manage containers" to check settings.
4) "Limit to designated sites" is (seemingly) randomly set to ON or OFF for each container.

### Expected behavior
All containers have "Limit to designated sites" set to on. 

### Notes
I debugged this and it looks like issue happens because [document.querySelector("#site-isolation")](https://github.com/mozilla/multi-account-containers/blob/52749ee1f0169706c424d8d0bc0b7846201e9046/src/js/popup.js#L1311) is returning all the containers that have been opened till that point. So on opening 1->2->3 containers in succession, that query will return [1,2,3] and will toggle the state for all 3, hence giving a random state at the end for each container.

Didn't understand the logic for loading or unloading panels, but the solution seems to be properly unloading the panel for a particular container on pressing the "back" button.

The workaround is of course to close the popup after changing each container and reopening it.

Thanks!

maxxcrawford:
Fixes #1781 

To verify:

1. Open panel, click "manage containers" 
1. Select first container, check box for Site isolation
1. Click < back button, select other container
1. check box for Site isolation in that container
1. Go back, select first container: 

Expected: Container checkbox is still checked. 

Run through the same test, except instead of going back, click out of panel and reopen/click "manage containers"

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "testpilot-containers",
   "title": "Multi-Account Containers",
   "description": "Containers helps you keep all the parts of your online life contained in different tabs. Custom labels and color-coded tabs help keep different activities — like online shopping, travel planning, or checking work email — separate.",
-  "version": "7.0.1",
+  "version": "7.0.2",
   "author": "Andrea Marchesini, Luke Crouch and Jonathan Kingston",
   "bugs": {
     "url": "https://github.com/mozilla/multi-account-containers/issues"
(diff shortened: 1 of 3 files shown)


eberhardweber:
I'm still getting this bug on Bazzite 43 and Firefox 147.0.1 flatpak, exactly as advertised in #1781 and #1815. 

Operating System: Bazzite 43
KDE Plasma Version: 6.5.5
KDE Frameworks Version: 6.22.0
Qt Version: 6.10.1
Kernel Version: 6.17.7-ba25.fc43.x86_64 (64-bit)
Graphics Platform: Wayland

Issue #1765, opened 2020-06-11
[Quality of Life request] Replace/close tab after clicking on a URL mapped to a Container while in another container.


4aHxKzD:
What the title says.

Consider for example having configured Multi-Account Containers to always open Facebook.com in its own container, and do the following steps:

1. You are currently using a search engine (e.g. Google, DuckDuckGo) while on the `Default Container`
2. You click a hyperlink to Facebook

Now you have two Tabs, one for your search engine and one for Facebook.

Clicking the Facebook hyperlink should create a new tab in its container, but also close the previous Tab that you were using to browse the internet.

xdhmoore:
I ran across this use case while using Tabliss (an extension for configuring your New Tab page). If I create a "Quicklink" to Facebook messenger on my Tabliss New Tab, clicking the link opens another tab with my configured MAC Facebook container, but leaves the old "New Tab" tab open (and changes it to a blank white page for some reason). Ideally, I think opening the link would replace the existing tab. I don't know how that would work with existing functionality. I suppose there are security rules why existing link target="" behavior can't be respected...?

Also, after digging through the issues I think this is probably a duplicate of #1590 .

Solid-Ice8:
For those that may not be aware, middle clicking any link will open the link in its own container.

4aHxKzD:
There is an option for this in Preferences now.

Issue #1704, opened 2020-04-10
Always open in specific container leaves deprecated tab


a-wigand:
- Multi-Account Containers Version: 6.2.3
- Operating System + Version: Windows 10 (1909)
- Firefox Version: 75.0 (64-bit)
- Other installed Add-ons + Version: LeechBlock NG (1.0.5)

### Actual behavior
When opening a site that is set to always open in a specific container, that site will be opened in a new tab. The deprecated tab has to be closed manually. So far so good and not a big problem.
With LeechBlock enabled and set to forward to a specific site, however, this can create an issue.
I set LeechBlock to open a site as a blocking site, which is set to open in a specific container. When triggered, this opens 'infinite' tabs of the same site as the original tab, that triggers the LeechBlock intervention never closes.

### Expected behavior
If opening a site that belongs to a different container, the current tab should be exchanged for the new tab (closing the current tab and opening the site in a new tab or switching the tab's container in one step)

### Steps to reproduce
1. Set LeechBlock to show a site that belongs to a specific container as a blocking site
2. Open a new tab
2. Enter a site that would be blocked by LeechBlock
3. The blocking site would usually be opened in the tab of the blocked site but as the container site opens in a new tab and the triggering tab never closes, this will create 'infinite' new tabs.

### Notes
This might be an issue with LeechBlock itself but even without LeechBlock I would prefer the container to open in the same tab or for the old tab to be closed automatically.

xdhmoore:
Related to #1765 and #1590

dannycolin:
Can you still reproduce this issue since #1835 has been merged?

Issue #1590, opened 2019-12-01
[Feature request] reuse current tab for "reopen in container" 


stinos:
When using the 'reopen in container' functionality I have personally not once wanted it to open a new tab and currently I always find myself using 'reopen in container' and then closing the original tab manually. I wouldn't be surprised if others have the same problem as well. So I'd like a way to have 'reopen in container' open the site in the current tab, not a new one. Or alternatively close the current one then open a new one.  This behavior could be behind a configuration option to enable/disable it, or a seperate menu entry 'reopen in container (current tab)' or so.

My usecase is that I work as contractor for different companies, and some of those require me to have different accounts for github/bitbucket and so on, and I'm logged in to those accounts in different containers. So often I end up on e.g. github (like now) but if I have to make changes under a non-standard account I have to reopen the same tab in the correct container first, and there's no reason to keep the tab open in the original container.

rhazor:
Any update on this, real struggle closing old tabs all the time after the page was opened in a container in a new tab by default.

PR #1835, opened 2020-07-27
Replace tab instead of creating a new one (#1765 #1704 #1590  #1837)

Moonloard:
Adds an option in "Preferences" to replace the current tab if a page which is assigned to another container is opened instead of keeping the current tab open. Middle clicking is unaffected and always opens a new tab as before.

Related issues which should be solved by this: #1765 #1704 #1590 #1837

### DIFF
--- a/src/js/background/assignManager.js
+++ b/src/js/background/assignManager.js
@@ -51,6 +51,11 @@ window.assignManager = {
       return !!syncEnabled;
     },
 
+    async getReplaceTabEnabled() {
+      const { replaceTabEnabled } = await browser.storage.local.get("replaceTabEnabled");
+      return !!replaceTabEnabled;
+    },
+
     getByUrlKey(siteStoreKey) {
       return new Promise((resolve, reject) => {
         this.area.get([siteStoreKey]).then((storageResponse) => {
@@ -233,9 +238,11 @@ window.assignManager = {
         return {};
       }
     }
+    const replaceTabEnabled = await this.storageArea.getReplaceTabEnabled();
     const removeTab = backgroundLogic.NEW_TAB_PAGES.has(tab.url)
       || (messageHandler.lastCreatedTab
-        && messageHandler.lastCreatedTab.id === tab.id);
+        && messageHandler.lastCreatedTab.id === tab.id)
+      || replaceTabEnabled;
     const openTabId = removeTab ? tab.openerTabId : tab.id;
 
     if (!this.canceledRequests[tab.id]) {
--- a/src/js/options.js
+++ b/src/js/options.js
@@ -28,14 +28,12 @@ async function enableDisableReplaceTab() {
 async function setupOptions() {
   const hasPermission = await browser.permissions.contains({permissions: ["bookmarks"]});
   const { syncEnabled } = await browser.storage.local.get("syncEnabled");
-  const { replaceTabEnabled } =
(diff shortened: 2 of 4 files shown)


src/js/background/assignManager.js
```diff
@@ -233,9 +238,11 @@ window.assignManager = {
         return {};
       }
     }
+    const replaceTabEnabled = await this.storageArea.getReplaceTabEnabled();
     const removeTab = backgroundLogic.NEW_TAB_PAGES.has(tab.url)
       || (messageHandler.lastCreatedTab
-        && messageHandler.lastCreatedTab.id === tab.id);
+        && messageHandler.lastCreatedTab.id === tab.id)
+      || replaceTabEnabled;
```
Moonloard, line 245: Essentially this one line is the only functionality change. The other code of this PR is there solely to add the option to turn this functionality on or off. The removal of the current tab was already implemented but was only used for blank tabs.

src/js/background/assignManager.js
```diff
@@ -51,6 +51,11 @@ window.assignManager = {
       return !!syncEnabled;
     },
 
+    async getReplaceTabEnabled() {
+      const { replaceTabEnabled } = await browser.storage.local.get("replaceTabEnabled");
+      return !!replaceTabEnabled;
```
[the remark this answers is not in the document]
groovecoder: Assigning the value with `!!checkbox.checked` means this isn't necessary, but I've been bitten by enough boolean type bugs to want to keep this here too. Belt *AND* suspenders. 👍

src/js/options.js
```diff
@@ -16,25 +16,24 @@ async function requestPermissions() {
...
-  if (checkbox.checked) {
-    await browser.storage.local.set({syncEnabled: true});
-  } else {
-    await browser.storage.local.set({syncEnabled: false});
-  }
+  await browser.storage.local.set({syncEnabled: !!checkbox.checked});
   browser.runtime.sendMessage({ method: "resetSync" });
 }
 
+async function enableDisableReplaceTab() {
```
groovecoder, line 23: funky way to name this, but I see it matches `enableDisableSync` so better to be consistently funky than pedantically un-funky.

src/js/options.js
```diff
@@ -16,25 +16,24 @@ async function requestPermissions() {
...
+  const { replaceTabEnabled } = await browser.storage.local.get("replaceTabEnabled");
   if (hasPermission) {
     document.querySelector("#bookmarksPermissions").checked = true;
   }
-  if (syncEnabled) {
-    document.querySelector("#syncCheck").checked = true;
-  } else {
-    document.querySelector("#syncCheck").checked = false;
-  }
+  document.querySelector("#syncCheck").checked = !!syncEnabled;
```
groovecoder, line 35: nice bit of clean-up. thanks.

src/js/background/assignManager.js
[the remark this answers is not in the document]
Moonloard: I also thought about removing it. Essentially what @groovecoder said is why I left it there. And for consistency with `getSyncEnabled` above.

src/js/options.js
Moonloard: Exactly what I thought. Maybe refactoring to `toggleSync` and `toggleReplaceTab` would be a good idea but I wasn't sure enough.

PR #1891, opened 2020-10-15
Bottom button shouldn't obstruct the list

yshui:
Having .bottom-btn { position: fixed; } cause the button to obstruct the
last few items in the list.


yshui:
also you really should've had the `overflow-x: hidden`, currently the list scroll left-right as well.

maxxcrawford (changes requested):
Could you please include some screenshots outlining this issue and include some steps to reproduce? We need some more info to review this properly. If there is an issue logged with this info, you can just include it. Thanks!

yshui:
1611343622x

You can clearly see the button obstructing the tab, and there is in fact one more tab below that one. Just open a bunch of tabs and you will see it.

yshui:
@maxxcrawford Also if you request change please tell me what I should change.

kendallcorner (approved):
Tested. Works great. Nice catch. Thanks!

Issue #1817, opened 2020-07-15
Adding Local PDFs to a container "breaks" the container


dylmye:
- Multi-Account Containers Version: 7.0.1
- Operating System + Version: Windows 10 Latest x64
- Firefox Version: Beta 79.0b7 (64-bit)
- Other installed Add-ons + Version + Enabled/Disabled-Status:

<details>
<summary>Extensions</summary>

|Name|Type|Version|Enabled|ID|
|--- |--- |--- |--- |--- |
|1Password extension (desktop app required)|extension|4.7.5.90|true|onepassword4@agilebits.com|
|Amazon.co.uk|extension|1.1|true|amazon@search.mozilla.org|
|BetterTTV|extension|7.2.83|true|firefox@betterttv.net|
|Bing|extension|1.1|true|bing@search.mozilla.org|
|bootstrap-responsiveHelper|extension|1.2.6|true|{76f8f20b-e3bc-44ea-a243-152f71953407}|
|Bypass Paywalls|extension|1.7.4|true|bypasspaywalls@bypasspaywalls.weebly.com|
|Chambers (UK)|extension|1.0|true|chambers-en-GB@search.mozilla.org|
|Cookie Quick Manager|extension|0.5rc2|true|{60f82f00-9ad5-4de5-b31c-b16a47c51558}|
|DuckDuckGo|extension|1.0|true|ddg@search.mozilla.org|
|eBay|extension|1.1|true|ebay@search.mozilla.org|
|Firefox Color|extension|2.1.5|true|FirefoxColor@mozilla.com|
|Firefox Multi-Account Containers|extension|7.0.1|true|@testpilot-containers|
|FrankerFaceZ|extension|4.0|true|frankerfacez@frankerfacez.com|
|Google|extension|1.0|true|google@search.mozilla.org|
|Honey|extension|12.1.1|true|jid1-93CWPmRbVPjRQA@jetpack|
|HTTPS Everywhere|extension|2020.5.20|true|https-everywhere@eff.org|
|JSON Lite|extension|19.6.0|true|{16a49f65-1369-4839-a5ef-db2581e08b16}|
|Markdown Viewer Webext|extension|1.5.0|true|{943b8007-a895-44af-a672-4f4ea548c95f}|
|MetaMask|extension|8.0.4|true|webextension@metamask.io|
|NightTimeFP theme|extension|1.0.1|true|{b3a962bb-594a-488e-ad44-d3e3e7673c67}|
[...]

Issue #1717, opened 2020-04-17
Hide/Show tabs in this container actually closes and reopens tabs instead of hiding/showing, meaning it cannot reopen (show) privileged tabs (file, chrome, about, etc.)


gijsk:
Original report is https://bugzilla.mozilla.org/show_bug.cgi?id=1620757 .

Quoting Tomislav:

> It looks like instead of showing/hiding tabs, it actually closes and opens them:
https://github.com/mozilla/multi-account-containers/blob/d654810d/src/js/background/backgroundLogic.js#L315
> 
> Unfortunately, we prevent extensions from opening file:// urls for security reasons, we only support limited interactions with local files that users explicitly open themselves.

This was previously covered in #46, but it doesn't look like it was obvious then that this means the functionality actually doesn't work correctly.

If we're concerned about resource usage, we should be able to use the tab discarding API to effectively remove the relevant docshells without closing the tab, before hiding the tabs. I'm hoping/assuming (but haven't tested) that discarding file:/chrome:/about: tabs is allowed.

schra:
This issue is a duplicate of #918, see https://github.com/mozilla/multi-account-containers/issues/918#issuecomment-360537785

> [Steps to reproduce]:
> 3. Open a "Work" container tab and drag a local file into it.

However I like like the wording of this issue better because it actually describes the root of the problem.

dannycolin:
Duplicate of #802

Issue #1926, opened 2020-12-20
Instructions in readme section "Development" do not work. Cannot build the plugin


schra:
- Multi-Account Containers Version: master (https://github.com/mozilla/multi-account-containers/commit/f0afc0da36d8bd217ef5db89c39f04cb0e27379d)
- Operating System + Version: Arch Linux
- Firefox Version: 84.0
- Other installed Add-ons + Version + Enabled/Disabled-Status: I disabled all plugins but Multi-Account Containers

### Actual behavior

```
$ npm install
npm WARN ajv-keywords@3.5.2 requires a peer of ajv@^6.9.1 but none is installed. You must install peer dependencies yourself.
npm WARN webextensions-api-fake@1.3.0 requires a peer of sinon@^8.1.1 but none is installed. You must install peer dependencies yourself.
npm WARN webextensions-jsdom@1.4.1 requires a peer of sinon@>=8.1.1 but none is installed. You must install peer dependencies yourself.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.13 (node_modules/chokidar/node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.13: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@2.1.3 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@2.1.3: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@2.0.1 (node_modules/web-ext/node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@2.0.1: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

audited 1688 packages in 6.261s

92 packages are looking for funding
  run `npm fund` for details

found 14 vulnerabilities (9 low, 5 high)
  run `npm audit fix` to fix them, or `npm audit` for details

$ ./node_modules/web-ext/bin/web-ext run -s src/
(node:41062) Warning: Accessing non-existent property 'cat' of module exports inside circular dependency
[...]

PR #1927, opened 2020-12-20
Fix breaking containers when adding tab with local file

### LINKED ISSUE #918: Container broke: "old" and "new" tabs both listed, cannot show or hide
One of my containers randomly broke, from now on: Work tabs.
(If i recall correctly,) My work tabs randomly opened as another instance, i closed them back in to not have duplicates.
After probably days, I wanted to hide Work tabs, but there was only "Show tabs".
Restarting the browser doesn't help, "Show tabs" works once, opens the hidden work tabs, but stays as "Show tabs".
The ones "stuck" are the old open tabs, they do not update, and stay listed after i close them.
The ones I am using, cannot be hidden, as theres no Hide tabs option, but update and show up correctly.
When i click Show Tabs after a restart, the "stuck" tabs are still listed as not opened, and they open and get listed as opened aswell

The last work tab I opened was a local html file, which might cause it (?)

I have no idea how to reproduce the bug, or how to supply some data to help investigate.

Firefox 57.0b10
Containers 4.0.3

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/MAC-617)

schra:
See https://github.com/mozilla/multi-account-containers/issues/1717 for
a description of the underlying issue:

> Hide/Show tabs in this container actually closes and reopens tabs
> instead of hiding/showing, meaning it cannot reopen (show) privileged
> tabs (file, chrome, about, etc.)

This commit fixes #918 and #1817

### DIFF
--- a/src/js/background/backgroundLogic.js
+++ b/src/js/background/backgroundLogic.js
@@ -93,7 +93,8 @@ const backgroundLogic = {
     // We can't open these we just have to throw them away
     if (protocol === "about:"
         || protocol === "chrome:"
-        || protocol === "moz-extension:") {
+        || protocol === "moz-extension:"
+        || protocol === "file:") {
       return false;
     }
     return true;


src/js/background/backgroundLogic.js
```diff
@@ -93,7 +93,8 @@ const backgroundLogic = {
     // We can't open these we just have to throw them away
     if (protocol === "about:"
         || protocol === "chrome:"
-        || protocol === "moz-extension:") {
+        || protocol === "moz-extension:"
+        || protocol === "file:") {
```
schra, line 97: Are there other "privileged tabs" than `about`, `chrome`, `moz-extension` and `file`? See https://github.com/mozilla/multi-account-containers/issues/1717 for context.

Furthermore, https://github.com/mdn/webextensions-examples/blob/2184fee75afcde39328384b36adf9e52c5917949/apply-css/background.js only allows the `http` and `https` protocols. Should we maybe do the same and just whitelist all good protocols? If yes, which?

schra:
I btw couldn't test this diff since I can't build the code. See #1926

dylmye:
While this looks correct to me, you can see by re-opening `about:preferences` in a Container that you're not prevented from actually adding the tab. There may need to be conditional logic on whether to grey out the 'reopen in container' menu. I think that logic would be in assignManager?

schra:
@dylmye you are correct. I was now able to build the plugin and can confirm that this diff does not yet prevent you from opening a local file in a container. I converted this PR to a draft and will have a look at this soon.

schra:
I moved to Simple Tab Group which doesn't have this problem.

PR #1944, opened 2021-01-08
Add a dark mode to MAC

dannycolin:
Hi,

As mentioned in #containers:mozilla.org, here's my patch to add a dark mode to MAC. Only the "Limit to designated sites" checkboxes aren't styled since checkbox elements aren't easily styleable without using `appearance: none`. 

Would it be acceptable to merge the patch as is or would you prefer if I also style the checkboxes? If you prefer the latter, we'll need an SVG for the checkbox mark icon.

### DIFF
--- a/src/css/popup.css
+++ b/src/css/popup.css
@@ -64,7 +64,6 @@ body {
   --icon-fit: 8;
 }
 
-
 @media (min-resolution: 1dppx) {
   html {
     font-size: 14px;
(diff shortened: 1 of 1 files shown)


groovecoder:
Looks really nice! I found a bug in the "Manage Container" panel - looks like there's a simple color inverse that also inverts the favicon, and the gradient background? It makes the "X" illegible.

image

dannycolin:
> Looks really nice! I found a bug in the "Manage Container" panel - looks like there's a simple color inverse that also inverts the favicon, and the gradient background? It makes the "X" illegible.

It should be fixed now :).

maxxcrawford (approved):
Looks great – If you wanna do a follow-up PR for the check boxes, feel free. This is ready to ship as is.

PR #1964, opened 2021-02-02
fix #1948 Add option to select a light/dark theme

### LINKED ISSUE #1948: Select light/dark theme option in preferences panel
Currently, Dark Mode support automatically [matches](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) the dark/light theme of the OS. 

There should be an option to override this behavior in the preferences panel. 

Copy:
- Headline: Dark Mode Theme
- Radio Buttons: 
  - Default to system
  - Light color mode 
  - Dark color mode 
- Description: This setting manages how the add-on appears in your browser. 
 
Placement: 

image

dannycolin:
This patch add a new preference in the option page to let the user select a light/dark theme. It also auto-detect the OS theme if it's set to `auto`.

### DIFF
--- a/src/js/options.js
+++ b/src/js/options.js
@@ -53,11 +53,20 @@ async function enableDisableReplaceTab() {
   await browser.storage.local.set({replaceTabEnabled: !!checkbox.checked});
 }
 
+async function changeTheme(event) {
+  const theme = event.currentTarget;
+  await browser.storage.local.set({currentTheme: theme.value});
+  await browser.storage.local.set({currentThemeId: theme.selectedIndex});
+}
+
 async function setupOptions() {
   const { syncEnabled } = await browser.storage.local.get("syncEnabled");
   const { replaceTabEnabled } = await browser.storage.local.get("replaceTabEnabled");
+  const { currentThemeId } = await browser.storage.local.get("currentThemeId");
+
   document.querySelector("#syncCheck").checked = !!syncEnabled;
   document.querySelector("#replaceTabCheck").checked = !!replaceTabEnabled;
+  document.querySelector("#changeTheme").selectedIndex = currentThemeId;
   setupContainerShortcutSelects();
 }
 
@@ -114,6 +123,8 @@ browser.permissions.onRemoved.addListener(resetPermissionsUi);
 document.addEventListener("DOMContentLoaded", setupOptions);
 document.querySelector("#syncCheck").addEventListener( "change", enableDisableSync);
 document.querySelector("#replaceTabCheck").addEventListener( "change", enableDisableReplaceTab);
+document.querySelector("#changeTheme").addEventListener( "change", changeTheme);
+
 maybeShowPermissionsWarningIcon();
 for (let i=0; i < NUMBER_OF_KEYBOARD_SHORTCUTS; i++) {
   document.querySelector("#open_container_"+i)
--- a/src/js/pageAction.js
+++ b/src/js/pageAction.js
@@ -32,6 +32,9 @@ async function init() {
   list.appendChild(fragment);
 
   MozillaVPN.handleContainerList(identities);
+
+  // Set the theme
+  Utils.applyTheme();
 }
 
 init();
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -51,6 +51,7 @@ async function getExtensionInfo() {
   return extensionInfo;
 }
 
+
 // This object controls all the panels, identities and many other things.
 const Logic = {
   _identities: [],
@@ -64,6 +65,9 @@ const Logic = {
     browser.runtime.sendMessage({
       method: "MozillaVPN_attemptPort"
     }),
+
+    // Set the theme
+    Utils.applyTheme();
 
     // Remove browserAction "upgraded" badge when opening panel
     this.clearBrowserActionBadge();
@@ -1492,7 +1496,7 @@ Logic.registerPanel(P_CONTAINER_EDIT, {
             } else {
               MozillaVPN.handleMozillaCtaClick("mac-edit-container-panel-btn");
             }
-           
+
           });
 
           this.switch.addEventListener("click", async() => {
--- a/src/options.html
+++ b/src/options.html
@@ -59,6 +59,24 @@
         <p><em data-i18n-message-id="replaceTabDescription"></em></p>
       </div>
 
+      <!--
+        TODO
+        - Add data-i18n
+      -->
+      <h3 data-i18n-message-id="theme"></h3>
+
+      <p><label class="keyboard-shortcut">
+          <span data-i18n-message-id="chooseTheme"></span>
+          <select id="changeTheme" name="changeTheme">
+            <option value="auto" selected data-i18n-message-id="themeAuto">
+            </option>
+            <option value="light" data-i18n-message-id="themeLight">
+            </option>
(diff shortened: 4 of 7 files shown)


maxxcrawford (changes requested):
Travis CI is kicking out some [lint errors](https://travis-ci.org/github/mozilla/multi-account-containers/builds/757285700) on the CSS. Could you please resolve them? 

```
23:16    ✖  Expected "#ffffff" to color-hex-length be "#fff"
111:1    ✖  Expected no more than 1 max-empty-lines empty line
115:3    ✖  Expected box-shadow to order/properties-alphabetical-order come before box-sizing
136:1    ✖  Expected no more than 1 max-empty-lines empty line
657:3    ✖  Unexpected duplicate declaration-block-no-duplicate-properties "background-color"
893:3    ✖  Expected filter to come order/properties-alphabetical-order before padding-inline-start
938:3    ✖  Unexpected property property-blacklist "border-top"
1092:1   ✖  Expected no more than 1 max-empty-lines empty line
```

dannycolin:
> Travis CI is kicking out some [lint errors](https://travis-ci.org/github/mozilla/multi-account-containers/builds/757285700) on the CSS. Could you please resolve them?

Done. :)

src/js/options.js
```diff
@@ -53,11 +53,20 @@ async function enableDisableReplaceTab() {
   await browser.storage.local.set({replaceTabEnabled: !!checkbox.checked});
 }
 
+async function changeTheme() {
+  const theme = document.querySelector("#changeTheme");
+  await browser.storage.local.set({currentTheme: theme.value});
+  await browser.storage.local.set({currentThemeId: theme.selectedIndex});
+}
```
groovecoder, line 60: It looks like this function is only called via `document.querySelector("#changeTheme").addEventListener(...)`, so, instead of calling `document.querySelector("#changeTheme")` again, we can reference the element via either `this` or `currentTarget`?

```suggestion
async function changeTheme(event) {
  const theme = event.currentTarget;
  await browser.storage.local.set({currentTheme: theme.value});
  await browser.storage.local.set({currentThemeId: theme.selectedIndex});
}
```

src/js/pageAction.js
```diff
@@ -34,4 +34,20 @@ async function init() {
...
+      popup.setAttribute("data-theme", "dark");
+    } else {
+      popup.setAttribute("data-theme", "light");
+    }
+  } else {
+    popup.setAttribute("data-theme", currentTheme);
+  }
+}
+
+applyTheme();
```
groovecoder, line 52: Nested `if/else` blocks smell like too much logic in a single function. Can we refactor into a helper function here?

```suggestion
function getTheme(currentTheme, window) {
  if (typeof currentTheme !== "undefined" && currentTheme !== "auto") {
    return currentTheme;
  }
  if (window.matchMedia("(prefers-color-scheme: dark)").matches) {
    return "dark";
  }
  return "light";
}

async function applyTheme() {
  const { currentTheme } = await browser.storage.local.get("currentTheme");
  const popup = document.getElementsByTagName("html")[0];
  const theme = getTheme(currentTheme, window);
  popup.setAttribute("data-theme", theme);
}

applyTheme();
```

src/js/popup.js
```diff
@@ -52,6 +52,23 @@ async function getExtensionInfo() {
...
+    } else {
+      popup.setAttribute("data-theme", "light");
+    }
+  } else {
+    popup.setAttribute("data-theme", currentTheme);
+  }
+}
+
+applyTheme();
+
```
groovecoder, line 71: suggestion: make the same `getTheme` helper refactor here as in `pageAction.js`, or bonus points if we can move this into a single `async function applyTheme(document, window)` in `utils.js` which is included in both popups.

src/options.html
```diff
@@ -59,6 +59,24 @@ <h3 data-i18n-message-id="tabBehavior"></h3>
         <p><em data-i18n-message-id="replaceTabDescription"></em></p>
       </div>
 
+      <!--
+        TODO
+        - Add data-i18n
+      -->
+      <h3 data-i18n-message-id="theme"></h3>
```
groovecoder, line 66: Looks like we still need to add all these messages to the https://github.com/mozilla-l10n/multi-account-containers-l10n/search?q=theme repo? Are you comfortable doing that? Or maybe @maxxcrawford can do that part?

groovecoder (changes requested):
Good stuff. Couple change requests but overall looking good.

src/js/pageAction.js
dannycolin: Fixed via [a29285f](https://github.com/mozilla/multi-account-containers/pull/1964/commits/a29285f0fa730dd27e406cc17decbbac60213f4a)

src/js/popup.js
dannycolin: [a29285f](https://github.com/mozilla/multi-account-containers/pull/1964/commits/a29285f0fa730dd27e406cc17decbbac60213f4a)

src/options.html
dannycolin: See https://github.com/mozilla-l10n/multi-account-containers-l10n/pull/12

groovecoder (approved):
JS, HTML, and locales code and spot-checks look good. Thanks! @maxxcrawford - did you want to check the CSS changes?

lesleyjanenorton (changes requested):
Hi Danny, overall this looks good but have just a few initial nits around the Mozilla VPN UI bits. I was only able to do a quick initial pass here and will pick this back up later today.

- Radio buttons in the server lists are not visible (photo attached)
- Transparency issue with Mozilla VPN tout background (photo attached)
- :active states missing for a number of buttons

Screen Shot 2022-03-25 at 12 07 36 PM
Screen Shot 2022-03-25 at 11 52 30 AM

dannycolin:
> * Radio buttons in the server lists are not visible (photo attached)
> * Transparency issue with Mozilla VPN tout background (photo attached)

Both issues should be fixed.

> * :active states missing for a number of buttons

I've added a `:active` state for all the elements that have one in the production version. For other elements in production that currently don't follow the Proton Design System, I intend to make a follow-up patch.

Thanks for the review :)

lesleyjanenorton (commented):
This is looking great! I'm not 100% familiar with how the l10n pipeline works but I think we want to remove src/_locales from this PR before merging.

dannycolin:
> This is looking great! I'm not 100% familiar with how the l10n pipeline works but I think we want to remove src/_locales from this PR before merging.

Done. I had to refresh a bit my git-fu on that one ;p

lesleyjanenorton (approved):

lesleyjanenorton:
@dannycolin @groovecoder it looks like we've lost the ability to open new tabs in a specific container by clicking on the ">" for that container from the doorhanger main view. Was that intentional?

dannycolin:
> Container _name_ still opens new tabs. All good.

Yep. It was like that before this patch and also throws me off a little bit the first time I discovered there was 2 clickable area. I think we should find a better solution and I'd be willing to work on this. I'll file a bug report and keep you posted in Slack with a mockup I've been working on for this. :)

lesleyjanenorton:
100% agree. There is almost no way to discover this functionality except by accident.

maxxcrawford (approved):
Awesome work, @dannycolin! Let's go! 🎸

PR #2024, opened 2021-04-21
Add option to disable the page action

bepvte:
In the recent Photon UI update, the ability to hide the page actions in the URL bar or manage them was removed. Personally, I never use the page action button to make a site always open in a tab, and always do it from the extensions toolbar button. Because this button just gets in my way, and because it used to be able to be moved until recently, I think making it toggleable would be a good idea.


bepvte:
oops

bepvte:
If your wondering whats happening here, I was removing my name and email from all my repos. I messed up pretty bad and pushed without checking 😬 . Ill make a new one of this with a new copy in a bit.

Issue #2044, opened 2021-05-06
Poor contrast on  "Always Open this Site in…" confirmation page


htwyford:
- Multi-Account Containers Version: 7.3.0
- Firefox Version: 89

This is a dupe of [Bug 1709812](https://bugzilla.mozilla.org/show_bug.cgi?id=1709812):

User Agent: Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:90.0) Gecko/20100101 Firefox/90.0

Steps to reproduce:

(I) Click on the hamburger menu
(II) Click on More tools
(III) Click on Customize Toolbar…
(IV) Select the Dark theme
(V) Open a webpage
(VI) Click on the Always open this in a Container icon in the urlbar
(VII) Select any Container
(VIII) A page should pop up, asking for confirmation

Actual results:

The foreground color, used for the URL, is white, whereas the background color is a slightly darker white. I can't tell which URL I'm asking Firefox to assign to a Container.

Expected results:

Either the foreground or the background color should be black.
Considering how in the White theme the foreground color is black and the background color is white, in the Dark theme the foreground color should be white and the background color should be black.
This doesn't occur in Windows 10

htwyford:
This issue has two related causes:
1. Starting in Firefox 89, in-content pages follow the Firefox theme and not the system theme. This happens with the new `-moz-toolbar-prefers-color-scheme: dark` media query: https://searchfox.org/mozilla-central/search?q=-moz-toolbar-prefers-color-scheme%3A+dark&path=&case=false&regexp=false.
2. We use solid colours on #redirect-url: https://github.com/mozilla/multi-account-containers/blob/4b56a2f0bb91cd9a12b95121fcff0c8e96ff66bd/src/css/confirm-page.css#L48-L65.

What's happening in the bug is that the user has a Light system theme but a dark Firefox theme. That gives us `prefers-color-scheme: light` but `-moz-toolbar-prefers-color-scheme: dark`. The in-content page in Firefox will be dark, but the #redirect-url element defined in this add-on will be light.

There are a few things we could do to resolve this. Using a transparent background would work, or we could use system colors like `Field` and `FieldText`. We could also define `#redirect-url` rules with `-moz-toolbar-prefers-color-scheme: dark` and use `prefers-color-scheme: dark` rules as a Firefox <89 fallback. Here's a sketch:

```
@media (prefers-color-scheme: dark) {
  #redirect-url {
      background: #38383d; /* Grey 70 */
[...]

PR #2045, opened 2021-05-06
#2044 - Use color-mix for confirmation page background on Firefox 89+.

htwyford:
color-mix is still experimental. Technically the API is subject to change, it but was [used widely](https://searchfox.org/mozilla-central/search?q=color-mix) in mozilla-central for Proton. It's really useful for cases like this, where you want a transparent variable background with opaque text on top. I opted to use color-mix so we don't have to tweak these colors each time Firefox is redesigned (see #1375!). Naturally, multi-account-containers is owned by Mozilla so we can always just tweak the implementation of color-mix down the road if necessary.

### DIFF
--- a/src/css/confirm-page.css
+++ b/src/css/confirm-page.css
@@ -45,8 +45,27 @@ html {
   word-break: break-all;
 }
 
+/* Fallbacks for Firefox < 89. In Firefox 89+, content pages like the
+   confirmation page follow the Firefox theme instead of the system theme.
+   That requires the -moz-toolbar-prefers-color-scheme media query. Instead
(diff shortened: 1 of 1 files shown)


htwyford:
I should mention I tested this on Release 88 and Nightly 90.

kendallcorner:
This is what I get with a light theme on FF 90:
Screen Shot 2021-07-25 at 10 57 28 PM

dannycolin:
It works on Firefox 107+. @htwyford would you mind rebasing your PR and it should be good to get this merge after that.

dannycolin:
Closing as staled.

PR #2191, opened 2021-11-02
Remove nonexistent "applications" manifest key

lesleyjanenorton:
- Removes the `"applications"` key from manifest.json, which doesn't seem to be a valid key (https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json).
- Moves `strict_min_version` to `browser_specific_settings > gecko > strict_min_version` where the AMO linters can find it.

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -8,12 +8,6 @@
     "48": "img/container-site-d-48.png",
     "96": "img/container-site-d-96.png"
   },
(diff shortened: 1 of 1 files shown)


maxxcrawford (approved):
Verified that older versions (Tested `91.0`) will not open. Good to go!

PR #2224, opened 2021-11-29
Version bump: 8.0.3

bakulf:


maxxcrawford (changes requested):
The [package.json](https://github.com/mozilla/multi-account-containers/blob/version8.0.3/package.json) also needs to be updated.

maxxcrawford (approved):

lesleyjanenorton (approved):

Issue #2076, opened 2021-07-01
Deleted containers reappearing


sunsetbrioche:
Multi-Account Containers Version: 7.3.0
    Operating System + Version: Windows 10 Pro 19041.1052
    Firefox Version: 89.0.2
    Other installed Add-ons + Version + Enabled/Disabled-Status: None

    Actual behavior
    Deleted a container. After a few days, multiple of the deleted container reappeared in the list.

    Expected behavior
    After deleting the container, that's it.

    Steps to reproduce
    1. Delete the desired container
    2. Reopen Firefox
    3. Check multi-account containers

    Notes
![image](https://user-images.githubusercontent.com/86792284/124159661-fd6f1b00-dacd-11eb-9fc5-80ade0c43c50.png)

ecureuill:
I'm facing the same issue, but instead of days, deleted containers reapears in minutes. I disabled the sync feature to avoid this issue.

Operating System + Version: Debian Bullseye
Firefox Version: 90.0.2
Other installed Add-ons + Version + Enabled/Disabled-Status: Temporary Containers 1.9.2

benediktms:
I'm having the same issue, disabling sync will stop containers from re-appearing, but that does not seem like a fix. Seems like there should be functionality to choose if you want containers to sync by overriding from local or restore to local containers.

Yowlen:
This has been a bug in some form or another for quite a while, unfortunately, and a rather important one, at that. Why it hasn't been fixed quickly, I have no idea, but it can be traced back to at least March of 2020 with #1696 near as I can tell.

I haven't had sync enabled for quite some time because otherwise it'll just flood my list with a bunch of tmp containers that are created by the Temporary Containers addon that I use in combination with this one for extra security and protection against trackers.

benediktms:
The worst thing about this issue is that you cannot even delete the containers manually. It seems that the sync feature has trouble replacing remote content with the local content. At least for the JSON that contains the container metadata.

iverik-no:
> The worst thing about this issue is that you cannot even delete the containers manually. It seems that the sync feature has trouble replacing remote content with the local content. At least for the JSON that contains the container metadata.

Not only that, but in my experience, if you try to delete a container, the sync somehow duplicates the container, and you'll end up with a ton of duplicates. 

[...]

Issue #1912, opened 2020-11-26
Thousands of Duplicate Containers when Syncing


criscolord:
- Multi-Account Containers Version: 7.1.0
- Operating System + Version: Win10 20H2, Ubuntu 20.04
- Firefox Version: 83.0
- Other installed Add-ons + Version + Enabled/Disabled-Status: 
1Password X – Password Manager	extension	1.22.3	true	{d634138d-c276-4fc8-924b-40a0ea21d284}
Amazon.com	extension	1.3	true	amazondotcom@search.mozilla.org
Bing	extension	1.2	true	bing@search.mozilla.org
Disable Facebook News Feed	extension	2.1	true	{85cd2b5d-b3bd-4037-8335-ced996a95092}
DuckDuckGo	extension	1.1	true	ddg@search.mozilla.org
eBay	extension	1.2	true	ebay@search.mozilla.org
Enhancer for YouTube™	extension	2.0.101	true	enhancerforyoutube@maximerf.addons.mozilla.org
Firefox Multi-Account Containers	extension	7.1.0	true	@testpilot-containers
Ghostery – Privacy Ad Blocker	extension	8.5.3	true	firefox@ghostery.com
Google	extension	1.1	true	google@search.mozilla.org
Night Owl🔥	extension	0.3.5	true	{74997d71-ea10-4020-921b-47056dd7d28c}
Reddit Enhancement Suite	extension	5.20.9	true	jid1-xUfzOsOFlzSOXg@jetpack
uBlock Origin	extension	1.31.0	true	uBlock0@raymondhill.net
Wikipedia (en)	extension	1.1	true	wikipedia@search.mozilla.org
Cisco Webex Extension	extension	1.8.0	false	ciscowebexstart1@cisco.com
Cookiebro	extension	2.18.0	false	{15b1b2af-e84a-4c70-ac7c-5608b0eeed5a}
HTTPS Everywhere	extension	2020.11.17	false	https-everywhere@eff.org
Load Time	extension	0.2.2	false	{f5434366-ae48-4e6c-91b8-370b89487e8f}
Startpage.com — Private Search Engine	extension	1.3.0	false	{20fc2e06-e3e4-4b2b-812b-ab431220cada}

### Actual behavior
[...]

Issue #1838, opened 2020-07-30
Feature request: Exclusion rule for "Enable Sync" feature


barnscott:
**TLDR:**
Requesting a feature where I can add wildcard rules for names of containers that WONT get sync'd with "Enable Sync" feature. For example, I want to be able to exclude containers that match "tmp*" from syncing.

**Background / reason:**
I use both "Multi-Account-Containers" add-on and "Temporary Containers" add-on. 
When I open a new tab, by default, it opens as a Temporary Container and the cookies will be deleted when the tab is closed. This is true unless the tab is configured to open with a Multi-Account-Container (MAC).

Currently, if "Enable Sync" is turned-on, then all of these temporary containers are also backed up, and it fails to "forget" these temporary containers even after the tab is closed, because Firefox Sync recovers/re-syncs the temporary containers that were intended for termination.

Currently, I have just turned-off the Enable Sync feature, but it would be preferable to still sync the containers **_I had manually created_** with MAC. Since Temporary Containers start with "tmp*", I would like a feature in which I can exclude all containers that begin with "tmp*" from being sync'd.

danielfilho:
Related

- #1748

Issue #1675, opened 2020-03-06
Sync Containers with Temporary Account Containers


ghost:
### Information
- Multi-Account Containers Version: 6.2.3
- Firefox Version: 73.0.1
- Other installed Add-ons + Version + Enabled/Disabled-Status: 
Temporary Containers v1.8

### Actual behavior and Expected behavior
It is syncing all Container Labels and works fine

### Problem
It is also syncing all Containers from the Temporary Account Containers Addon.
Temporary Account Container is deleting the tmp containers after the Tab is closed for 2 minutes. So it is often synced to the cloud.
So i am getting on all other Firefox instances the "tmp**" labels from my other clients.
It would be good if it is possible for example to disable syncing all containers with Syntax "tmp**"

## Feature Request
Allow to disable/ignore syncing for cointainers with special characters or syntax

### Steps to reproduce
1. Use Multi Account Container with temporary Containers
2. Create a new TMP Container 
3. It will be Synced with Firefox Sync

tigerjack:
I have the same problem and it get so serious that I am unable to synchronize anything at all due to the huge number of temporary containers synchronized
```

1584168207630	Sync.Engine.Extension-Storage	ERROR	Syncing @testpilot-containers: request failed: Error: HTTP 507 Insufficient Storage: Resource access is forbidden for this user (Maximum bytes per object exceeded " "(39015 > 16384 Bytes.)(resource://services-common/kinto-http-client.js:2771:5) JS Stack trace: ServerResponse@kinto-http-client.js:2771:5
processResponse@kinto-http-client.js:2925:13
1584168207630	Sync.Engine.Extension-Storage	WARN	Syncing failed: Error: HTTP 507 Insufficient Storage: Resource access is forbidden for this user (Maximum bytes per object exceeded " "(39015 > 16384 Bytes.)(resource://services-common/kinto-http-client.js:2771:5) JS Stack trace: ServerResponse@kinto-http-client.js:2771:5
processResponse@kinto-http-client.js:2925:13
```

I have also tried to manually delete the temp containers from `~/.mozilla/firefox/PROFILENAME/containers.json`; at the startup the containers are not shown, but they gradually start to pop up again after a while.

alongfield:
[...]

PR #2231, opened 2021-12-05
Excluding containers from sync with RegExp

stoically:
Introduces an input box that allows to specify a RegExp pattern which filters containers by name before storing them to sync storage.

I'm not sure how exactly the new locale strings should get introduced in the submodule, this PR points to my fork of multi-account-containers-l10n, which surely isn't how it should be done. Would I open a PR in the multi-account-containers-l10n repo?

The default pattern `^tmp\d+$` is a really narrow and specific pattern that should not interfere with regular usage of MAC, but helps thousands of Temporary Containers users to be able to also use sync without having to manually adjust preferences. In the future more sophisticated solutions could be sought after, but this relieves the biggest pain point. Originally it was also planned to support the combination of MAC + TC when sync was introduced (https://github.com/mozilla/multi-account-containers/pull/1611#discussion_r370247876), but it unfortunately didn't work out.

Fixes https://github.com/mozilla/multi-account-containers/issues/1675
Fixes https://github.com/mozilla/multi-account-containers/issues/1623
Fixes https://github.com/mozilla/multi-account-containers/issues/1838
Fixes https://github.com/mozilla/multi-account-containers/issues/1912
Fixes https://github.com/mozilla/multi-account-containers/issues/2076
Fixes https://github.com/stoically/temporary-containers/issues/371

@bakulf @groovecoder Would be great if could give this a look, thanks!


FlorianWendelborn (approved):
LGTM, surprised how minimalistic the implementation can be for this feature

zwarag:
Hello @FlorianWendelborn. Just out of curiosity. You wrote LGTM but did not add a review. Since then the code changed, and now it's failing. Is this on purpose? I'm just curious and not trying to pressure :)

FlorianWendelborn:
> You wrote LGTM but did not add a review  
> — @zwarag

I did, I approved.

> Since then the code changed, and now it’s failing  
> — @zwarag

I neither changed the code, made it fail nor have the ability to do either of this

PixsaOJ:
This only fixes prefix "tmp" what if I use prefix "anon" ?

dannycolin:
Instead of filtering out any container that match a regex, why not implementing the opposite solution. What I mean is MAC could add a prefix to the containers it creates (e.g. Personal -> MAC_Personal). We could easily split the string to only show the part after the `_` in the UI. And, when we sync the containers, we only do so for those starting with the `MAC_` prefix. This way we don't need to add an input box that the user needs to deal.

PixsaOJ:
How long does this take to merge?

stoically:
@PixsaOJ There isn't any reaction from the maintainers - so likely never.

dannycolin:
> @PixsaOJ There isn't any reaction from the maintainers - so likely never.

As a contributors, I do understand your frustration that it takes more time than you'd expect to get this issue moving forward. However, I don't think comment like this will help us.

stoically:
> > @PixsaOJ There isn't any reaction from the maintainers - so likely never.
> 
> As a contributors, I do understand your frustration that it takes more time than you'd expect to get this issue moving forward. However, I don't think comment like this will help us.

Didn't want to sound frustrated, sorry. Last I checked the project was only in maintenance mode, new features were mostly related to VPN stuff - and my one line change from 2018 is still open as well: https://github.com/mozilla/multi-account-containers/pull/1141

PixsaOJ:
> > @PixsaOJ There isn't any reaction from the maintainers - so likely never.
> 
> As a contributors, I do understand your frustration that it takes more time than you'd expect to get this issue moving forward. However, I don't think comment like this will help us.

Can you share your opinion about this feature, and how important is it to merge?

dannycolin:
@stoically I pinged you on Matrix to see if I can help you with this.

PixsaOJ:
Btw, if anyone wants a solution: disable sync in extension settings and install separate Container sync extension.

src/js/background/backgroundLogic.js
```diff
@@ -362,6 +364,17 @@ const backgroundLogic = {
   cookieStoreId(userContextId) {
     if(userContextId === 0) return "firefox-default";
     return `firefox-container-${userContextId}`;
+  },
+
+  async setDefaultOptions() {
```
dannycolin, line 369: I think `setDefaultOptions` is a bit too generic for a name and we don't really have other defaults right now. IMO, something like `setSyncExclusion` or the like would be more appropriate.

dannycolin (changes requested):
Looks good to me :). But could you remove `src/_locales` from the commit and open a PR for it on the MAC l10n repo. See https://github.com/mozilla/multi-account-containers#addupdate-messages-for-translation=.

src/options.html
```diff
@@ -83,6 +83,9 @@ <h3 data-i18n-message-id="keyboardShortCuts"></h3>
       <h3 data-i18n-message-id="onboarding"></h3>
       <button data-i18n-message-id="resetOnboardingPanels"></button>
       <p><em data-i18n-message-id="onboardingToggle"></em></p>
+      <h3 data-i18n-message-id="syncExclude"></h3>
```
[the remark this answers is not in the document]
stoically: Makes sense, moved.

stoically:
@dannycolin Thanks. Made the changes and moved the locale change to https://github.com/mozilla-l10n/multi-account-containers-l10n/pull/16. The mentioned string is no longer required as it's now under the "firefox sync" heading.

src/options.html
```diff
@@ -47,6 +47,9 @@ <h3 data-i18n-message-id="firefoxAccountsSync"></h3>
           <span class="bold" data-i18n-message-id="enableSync"></span>
         </label>
         <p><em data-i18n-message-id="enableSyncDescription"></em></p>
+        <br>
+        <input id="syncExcludeRegExp">
```
dannycolin, line 51: I got a request to add a placeholder for the input element. I guess something like "Insert a regular expression" should be good enough.

dannycolin: I think we should also have a label for the input (e.g. "Container exclusion") just to make sure it's as accessible as possible.

stoically: Had to add some CSS to make it look OK with the combination of label and input box. Adjusted the i18n PR to include the label string. Also added the placeholder as plain sting for now.

dannycolin: I found how.
```
        <input id="syncExcludeRegExp"
               data-i18n-attribute="placeholder"
               data-i18n-attribute-message-id="syncInput">
```
It only supports one attribute tho but if you want to update it that's the way.

gpoole:
@dannycolin I can see you've rolled back the fix for this in https://github.com/mozilla/multi-account-containers/pull/2450. Will it be back? This is a long term frustration for users of this setup so it would be awesome to have a fix in place.

dannycolin:
Stoically is the author of the first patch. I'm still open to review a new patch for this issue but I personally won't work on it.

gpoole:
I'm happy to work on it; I found the patch as I was submitting a PR for my own version 🤦. Is there a reason it was rolled back that needs to be fixed before it can be released?

dannycolin:
The patch doesn't work as intented. See: https://github.com/mozilla/multi-account-containers/issues/2439

dwillis77:
Any update on when this fix might get re-implemented? This has been an issue for a long time, and it appears the reason for backing out the original fix ( #2439 ) has itself been fixed, so if there is any way the regex exclusion option could be added back that would be much appreciated!

(apologies if there is some other roadblock for getting this fixed that I'm missing, which is very possible - just wanted to check in and see what the status was)

Thanks!

lkraider:
Syncing temporary containers is still an issue.
@gpoole do you believe you can work on this issue?

MurzNN:
Is this feature removed in the current version of the extension? Can't find it anywhere in the settings and even in the source code: https://github.com/search?q=repo%3Astoically%2Fmulti-account-containers+syncExcludeRegExp&type=code

FlorianWendelborn:
> Is this feature removed in the current version of the extension?
> — @MurzNN

Yes, it sadly seems to have been removed. The original issue #1675 was re-opened and there is a bit more context there.

PR #2267, opened 2022-01-17
Remove delete container warning

lesleyjanenorton:
We missed a string in the delete container panel during internationalization. After reviewing this screen, I think we should remove this string altogether and possibly consider implementing something more helpful as a follow-up (perhaps showing the titles of open tabs, pulling relevant tabs into a new window so the user doesn't have to potentially go hunting through their open tabs to find them, something else...), or replacing this with a single generic "Tabs open in this Container will be closed" flag.

Fixes Jira ticket MAC-698

### DIFF
--- a/src/css/popup.css
+++ b/src/css/popup.css
@@ -1542,6 +1542,7 @@ manage things like container crud */
   font-size: 13px;
   inline-size: 100%;
   justify-content: space-between;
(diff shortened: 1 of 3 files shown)


maxxcrawford (approved):
This works as expected: 

image

I did see one other issue while reviewing this: 

The Proxy onboarding panel (with the two buttons) was missing the blue texts CTA). 

image

PR #2283, opened 2022-02-09
return single ProxyInfo object when advanced proxy

groovecoder:
To test:

0. Run Mozilla VPN
1. Change to this branch of code
2. Run the extension (`web-ext run -s src`)
3. Create a container and make it use a Mozilla VPN proxy
4. In the vpn container, go to https://mullvad.net/en/ and make sure the top banner shows the vpn location you expect
5. In the vpn container, go to https://www.dnsleaktest.com/ and make sure it shows the vpn location you expect
6. Create a container using an advanced proxy setting (e.g., `socks://au3-wg.socks5.mullvad.net:1080`)
4. In the advanced proxy container, go to https://mullvad.net/en/ and make sure the top banner shows the proxy location you expect
5. In the advanced proxy container, go to https://www.dnsleaktest.com/ and make sure it shows the proxy location you expect
4. In a NON-container tab, go to https://mullvad.net/en/ and make sure the top banner shows the proxy location you expect
5. In a NON-container tab, go to https://www.dnsleaktest.com/ and make sure it shows the proxy location you expect

### DIFF
--- a/src/js/background/assignManager.js
+++ b/src/js/background/assignManager.js
@@ -198,7 +198,7 @@ window.assignManager = {
     }
 
     if (!result.proxy.mozProxyEnabled) {
-      return [{ ...result.proxy, proxyDNS: true }];
+      return { ...result.proxy, proxyDNS: true };
     }
 
     // Let's add the isolation key.


src/js/background/assignManager.js
```diff
@@ -198,7 +198,7 @@ window.assignManager = {
     }
 
     if (!result.proxy.mozProxyEnabled) {
-      return [{ ...result.proxy, proxyDNS: true }];
+      return { ...result.proxy, proxyDNS: true };
```
bakulf, line 201: Do we have to fix line 191, 197, and 205 too?

bakulf (approved):

src/js/background/assignManager.js
groovecoder: Not that I could tell from the end-to-end testing.

lesleyjanenorton: These brackets, in particular!..(🌪)

So after updating to an 8.0.6 with this change, proxy settings configured in 8.0.5 or earlier should just work again?

groovecoder: Yes, it should just work.

groovecoder: It could be. TBH I looked at [the code changes that caused the bug](https://github.com/mozilla/multi-account-containers/pull/2258/files), and I noticed that I unneccessarily changed [this line](https://github.com/mozilla/multi-account-containers/pull/2258/files#diff-b9770584df24d2a25f9eb106e0dfc4d079bb6ce210ce18c933f7f53f5657789fR201) to return an array instead of an object like it was doing before. So I turned it back into an object and it fixed the bug. 🤷 

I'll show people from the add-ons team this issue and see if they think it's a bug in Firefox or maybe the docs aren't clear?

PR #2288, opened 2022-02-13
Honor the "don't ask again" checkbox when use of the assigned container is *denied*

### LINKED ISSUE #1730: "Remember my decision for this site" doesn't remember decision to override default container
- Multi-Account Containers Version: 6.2.5
- Operating System + Version: Windows 10 Education
- Firefox Version: 75.0
- Other installed Add-ons + Version + Enabled/Disabled-Status: Switch Container Plus v0.19

### Actual behavior
Every time I navigate to Google Drive from a "Personal" container when "School" is set as the default, firefox asks me if I *really* want to do that, even when I tell it to remember my decision.

### Expected behavior
After telling firefox to "remember my decision for this site", it will stop asking my permission to override the default container I set for that site.

Just to make the use case clear:

*Most* of the time I want to use my "School" account. So when I open up a new tab (which will start in the blank container) and go to Google Drive, it should switch to my "School" container. But there are still lots of times that I want to use my "Personal" account, so if I open up a "Personal" tab and go to Google Drive, it should just stay in my "Personal" container!

### Steps to reproduce
1. Navigate to a site in container A
2. In the add-on, check "Always open in container A"
3. Open up a new tab in container B
4. Navigate to the same site
5. When firefox asks, "Open this site in your assigned container?" check "Remember my decision for this site" and then click "Open in container B"
6. Open a new tab in container B and navigate to the same site
7. The same screen will appear asking for your permission

### Notes
I feel like this is almost worth a feature request for a less aggressive "Always open in" setting. There's also confusing terminology, as the base/blank container is titled "Default", but when I say default container I mean the container a site will default to. Maybe the current behavior works for some people, but as a user I expect the following behavior:

> I want this site to open in container A by default, but if I explicitly navigate there from container B, then open the site in container B!

It seems pretty hard to accidentally open a container tab, so I don't get why firefox is so insistent on double checking that I actually want to use that container.

D-side:
My rough attempt at remembering to stay in the original container when the checkbox "don't ask again" is **checked**, but use of the assigned container is **denied**. Fixes #1730

This is done by introducing a new site setting `forceContainer`, which indicates what decision was made by the user when they passed through the confirmation page with the "don't ask again" checkbox checked: `true` (default) if they confirmed the use of the assigned container and, conversely, `false` if they denied it.

> ⚠ One possibly undesirable side effect — it doesn't seem possible to open websites assigned to containers in the default container anymore. At all. I would love to differentiate tabs opened from the OS from the ones opened explicitly, but I don't know how hard that is right now.

> 🚧  I haven't wrapped my head around Mocha and the mocks in use here, so no tests yet, thus it's a draft. Any help with this or other implementation improvements are most welcome! 🙌

> ➕ And I also added `<!DOCTYPE html>`s all over the HTMLs to get rid of the Quirks Mode warnings. Open to extracting into another PR if there's a need for that!

### DIFF
--- a/src/confirm-page.html
+++ b/src/confirm-page.html
@@ -1,3 +1,4 @@
+<!DOCTYPE html>
 <html>
 <head>
   <meta http-equiv="content-type" content="text/html; charset=utf-8">
--- a/src/js/background/assignManager.js
+++ b/src/js/background/assignManager.js
@@ -169,6 +169,7 @@ window.assignManager = {
       this.storageArea.get(pageUrl).then((siteSettings) => {
         if (siteSettings) {
           siteSettings.neverAsk = true;
+          siteSettings.forceContainer = m.forceContainer;
           this.storageArea.set(pageUrl, siteSettings);
         }
       }).catch((e) => {
@@ -323,7 +324,8 @@ window.assignManager = {
         tab.index + 1,
         tab.active,
         siteSettings.neverAsk,
-        openTabId
+        openTabId,
+        siteSettings.forceContainer
       );
     }
     this.calculateContextMenu(tab);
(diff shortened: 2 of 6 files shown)


D-side:
Yeah, I clearly didn't study the mechanics in this addon very well, 'cause navigating to an address assigned to one container while in another causes the request to be **blocked**. I figure the cause is somewhere in the "on before request"…

dannycolin:
Thank you for your contribution. The confirmation flow definitely needs to be reworked. However, I don't think that simply accepting the "deny" choice is the solution that should be implemented here.

I think we should limit this to a more simpler solution where you confirm or remove the assignment. Something like:
image 

Feel free to join on your Matrix room at https://matrix.to/#/#containers:mozilla.org or open an issue, here on GitHub, to discuss the details before working on the code.

D-side:
Thanks! Sorry to leave this hanging for so long.
So long in fact that I lost the job where I've had this problem I was trying to fix here. 😅

Yeah, I suppose I was going with the assumption that my container settings should generally remain static — but that was probably because changing them was cumbersome. I've since gotten used to the button in the address bar that does the job in a couple clicks every time. Not sure it was there back then, but it's certainly there now.

So this really was just begging for closure. Sorry again about the distraction!

Issue #2295, opened 2022-02-21
Proxy settings dialog/popup truncates input to 50 characters


Caligatio:
- Multi-Account Containers Version: 8.0.6
- Operating System + Version: Windows 10
- Firefox Version: 97.0.1
- Other installed Add-ons + Version + Enabled/Disabled-Status: N/A

### Actual behavior
The containers proxy dialog/popup truncates long input to 50 characters which then mangles the proxy settings

### Expected behavior
To be able to enter proxy settings correctly

### Steps to reproduce
1. Open the dialog/popup to edit proxy settings for a container
2. Enter a 50+ character string, e.g. socks://username:password@test-socks5.relativelylongdomain.com:1111 and hit `Apply to Container`
3. Go back to the settings box and note that the parsed server is incorrect, e.g. socks://test-socks5.relativelylo:undefined
4. Attempting to use the container tab with the mangled proxy settings goes straight out using native connection

### Notes
The code in question is at https://github.com/mozilla/multi-account-containers/blob/cdd8ebc66a7101f33c62de7ab9d8d744f280c6d4/src/popup.html#L302

I'm happy to do a PR to eliminate the `maxlength` property but I'm not sure if there are side-effects to doing so.

dannycolin:
Closing since a patch has been merged in the main branch.

dabblejohns:
the 50 character limit on proxy still seems to be an issue is there a patch or way to fix this ?

PR #2296, opened 2022-02-21
Remove maxlength from proxy settings

Caligatio:
Per my #2295 issue, the UI limits the total proxy input length to 50 characters which is insufficient in the case of credentials or long domain names. This PR simply removes the `maxlength` property completely as there is no upper bound should someone have a very long username and/or password.

### DIFF
--- a/src/popup.html
+++ b/src/popup.html
@@ -299,7 +299,7 @@
           <legend class="form-header" data-i18n-message-id="icon"></legend>
         </fieldset>
         <fieldset class="proxies"> <!---- PROXIES -->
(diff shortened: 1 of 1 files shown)


bakulf (approved):

sunknudsen:
@bakulf I am experiencing same issue… has PR been released?

thansk:
@sunknudsen it's been fixed for me for a while, does it work for you?

dabblejohns:
how do you update this to get the length on proxy changed i am having isseus when uploading proxy on mine

dabblejohns:
> Per my #2295 issue, the UI limits the total proxy input length to 50 characters which is insufficient in the case of credentials or long domain names. This PR simply removes the `maxlength` property completely as there is no upper bound should someone have a very long username and/or password.

anyway you could show me how to fix this?

PR #2333, opened 2022-04-19
enable proxyDNS only for supported proxies (SOCKS4/5)

### LINKED ISSUE #2278: Advanced Proxy Settings PROBLEM!
The "Advanced Proxy Settings" option does not change my proxy on any of my containers. Yesterday it worked. I need help please.

eladyn:
From https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/proxy/ProxyInfo:

> `proxyDNS`
> `boolean`. If true, the proxy server is used to resolve certain DNS queries (only usable with `"socks4"` and `"socks"`). Defaults to `false`.

This only enables `proxyDNS`, if the proxy type is a `socks` or `socks4` proxy, because setting it on other types caused the whole proxy setting to be ignored.

Fixes #2278 (at least for me, confirmations appreciated)


src/js/background/assignManager.js
```diff
@@ -197,12 +197,17 @@ window.assignManager = {
       return {};
     }
 
+    // proxyDNS only works for SOCKS proxies
+    if (result.proxy.type === "socks" || result.proxy.type === "socks4" ) {
```
nekno, line 201:
```suggestion
    if (["socks", "socks4"].includes(result.proxy.type)) {
```

Just a thought on the approach. Thanks for fixing this!

nekno (approved):
Works for me with an HTTP proxy over the LAN, with or without my additional edit.

src/js/background/assignManager.js
eladyn: Thanks for the suggestion, that is indeed a good idea!

bakulf (approved):

sysarcher:
Hi Folks... any idea when this will land on the firefox addons page?

PR #2383, opened 2022-06-24
Fix empty regexp and incoming sync exclude behavior

stoically:
- [x] Leaving the regexp empty would no longer sync any outgoing container
- [x] Incoming containers should get excluded as well
- [x] Cover the scenarios with tests

Follow-up from https://github.com/mozilla/multi-account-containers/pull/2231.

### DIFF
--- a/src/js/background/sync.js
+++ b/src/js/background/sync.js
@@ -128,13 +128,12 @@ const sync = {
       await sync.checkForListenersMaybeAdd();
 
       async function updateSyncIdentities() {
-        const { syncExcludeRegExp } = await browser.storage.local.get("syncExcludeRegExp");
-        const excludeRegExp = new RegExp(syncExcludeRegExp, "i");
+        const excludeRegExp = await getExcludeRegExp();
         const identities = await browser.contextualIdentities.query({});
 
         for (const identity of identities) {
           // skip excluded identities
-          if (identity.name.match(excludeRegExp)) {
+          if (excludeRegExpMatchesIdentity(excludeRegExp, identity)) {
             continue;
           }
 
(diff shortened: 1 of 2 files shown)


src/js/background/sync.js
```diff
@@ -343,6 +343,14 @@ async function reconcileIdentities(){
     const deletedCookieStoreId = 
       await identityState.lookupCookieStoreId(deletedUUID);
     if (deletedCookieStoreId){
+      if (excludeRegExp) {
+        const deletedIdentity = await identityState.get(deletedCookieStoreId);
+        // skip excluded identities
+        if (excludeRegExpMatchesIdentity(excludeRegExp, deletedIdentity)) {
+          continue;
+        }      
+      }
```
dannycolin, line 352: Any chance we can unnest this a bit and assign the last if condition to a variable. Should be easier to read and refactor in the future if we need to check for other things.

src/js/background/sync.js
```diff
@@ -585,3 +603,21 @@ async function setAssignmentWithUUID(assignedSite, urlKey) {
...
+}
+
+function excludeRegExpMatchesIdentity(excludeRegExp, identity) {
+  if (excludeRegExp && identity.name.match(excludeRegExp)) {
+    if (SYNC_DEBUG) console.log(`Exclude regexp matches identity '${identity.name}'`);
+    return true;
+  } else {
+    return false;
+  }
+}
```
dannycolin, line 623: I know we aren't doing it for the rest of the code put I think it'd be great to have a short description and what's the variable passed and spitted. It could save someone else (maybe future you or me) time and maybe headaches :).

stoically: Comments added. Never too late to start documenting things. :+1:

dannycolin (approved):

PR #2390, opened 2022-07-19
Add ability to filter the container list

### LINKED ISSUE #1103: UI Enhancement: Scrollable container list with search bar
After about 20 containers it becomes unwieldy. It would be good if the container list scrolled so that it had a maximum height and above the container box was a search box that you can type and it would filter the displayed list based on what was typed.

It would make utilizing many containers much more user friendly.

dannycolin:
Resolves #1103 

Preview:

mac-filter-container-list

### DIFF
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -51,7 +51,6 @@ async function getExtensionInfo() {
   return extensionInfo;
 }
 
-
 // This object controls all the panels, identities and many other things.
 const Logic = {
   _identities: [],
@@ -465,6 +464,23 @@ const Logic = {
     default:
       break;
     }
+  },
+
+  filterContainerList() {
+    const pattern = /^\s+|\s+$/g;
+    const list = Array.from(document.querySelectorAll("#identities-list tr"));
+    const search = document.querySelector("#search-terms").value.replace(pattern, "").toLowerCase();
+
+    for (const i in list) {
+      const text = list[i].querySelector("td div span");
+
+      if (text.innerText.replace(pattern, "").toLowerCase().includes(search) ||
+          !search) {
+        list[i].style.display = "block";
+      } else {
+        list[i].style.display = "none";
+      }
+    }
   }
 };
 
@@ -847,6 +863,7 @@ Logic.registerPanel(P_CONTAINERS_LIST, {
 
     document.addEventListener("keydown", Logic.keyboardNavListener);
     document.addEventListener("keydown", Logic.shortcutListener);
+    document.addEventListener("input", Logic.filterContainerList);
 
     MozillaVPN.handleContainerList(identities);
 
--- a/src/popup.html
+++ b/src/popup.html
@@ -158,6 +158,18 @@
       </h4>
     </div>
     <div class="scrollable identities-list">
+      <div class="searchbar">
+        <label for="search-terms"
+               class="hide-label"
+               data-i18n-message-id="filterInputLabel">
+        </label>
+        <input type="text"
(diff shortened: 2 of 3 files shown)


src/js/popup.js
```diff
@@ -465,6 +464,24 @@ const Logic = {
...
+    const pattern = /^\s+|\s+$/g
+    const list = Array.from(document.querySelectorAll("#identities-list tr"));
+    const search = document.querySelector("#search-terms").value.replace(pattern, '').toLowerCase();
+
+    for (let i in list) {
+      let text = list[i].querySelector("td div span");
+
+      if (text.innerText.replace(pattern, '').toLowerCase().includes(search) ||
+          search == "" ||
+          search == null) {
```
stoically, line 479:
```suggestion
      if (text.innerText.replace(pattern, '').toLowerCase().includes(search) || !search) {
```

stoically (approved):

src/popup.html
```diff
@@ -158,6 +158,18 @@ <h4 class="moz-vpn-logotype vpn-status-container-list display-none">Mozilla VPN
       </h4>
     </div>
     <div class="scrollable identities-list">
+      <div class="searchbar">
+        <label for="search-terms"
+               class="hide-label"
+               data-i18n-message-id="filterInputLabel">
```
flodolo, line 164: I don't see this (`Filter container list by name:`) in the GIF. How is this used?

dannycolin: The label is for screen reader only.

groovecoder (approved):
Code changes look good.

What if we make the pop-up auto-focus into this new container filter list? Or maybe so we don't break existing habits where people might immediately type "1" or "2" to open a container tab in the 1st or 2nd container ... can we make the filter box completely invisible? So, if I open the pop-up and start typing "pers" or "shop" (basically any character that's NOT a number), it will call `filterContainerList` with what I've typed so far? And then, for mega bonus cherries, if the filtered list contains only 1 item, we auto-focus it, so pressing "enter/return" will open it.

I guess I'm thinking of the auto-filter/search experience in other apps:

1. Press Command + K
2. Start typing anything
3. The list of matching "things" gets smaller and smaller until there's just 1 item
4. Press "Return/Enter" to use the 1 item

dannycolin:
> @groovecoder What if we make the pop-up auto-focus into this new container filter list? Or maybe so we don't break existing habits where people might immediately type "1" or "2" to open a container tab in the 1st or 2nd container

I decided not to autofocus exactly because of the hidden functionality to open a container. 

> Can we make the filter box completely invisible? So, if I open the pop-up and start typing "pers" or "shop" (basically any character that's NOT a number), it will call `filterContainerList` with what I've typed so far? 

On a UX point of view, I'd be against hiding the input. I'm already not a fan of the hidden "press a number" function.

> And then, for mega bonus cherries, if the filtered list contains only 1 item, we auto-focus it, so pressing "enter/return" will open it.

Actually, I want to implement something similar in a follow-up PR. Pressing Enter could open whatever the first element in the filtered list is. That first element should always be hightlighted and the starting point if you tab/arrow down so that would be easy to select other items in the list.

However, all of this definitely needs UX work first so I'll merge this PR as-is and we can work on more advanced features in a follow up bug.

groovecoder:
Sounds good, thanks @dannycolin !

groovecoder:
Last idea-bomb: Maybe if someone types "f" for "filter" in the pop-up, that will auto-focus into the container filter input?

PR #2391, opened 2022-07-21
Fix #2168 Use a different l10n string for sites opened in no container

### LINKED ISSUE #2168: [l10n] String openInContainer is reused for no container
cc @lesleyjanenorton @bakulf 

```
  "openInContainer": {
    "message": "Open in $containerName$ Container",
    "placeholders": {
      "containerName": {
        "content": "$1"
      }
    }
  },
```

This string is reused when is empty. We need a different string for that.

Schermata 2021-10-26 alle 14 21 35

dannycolin:
If you assign a website to the Personal container and open the same website in the Work container, it will show you a confirmation page where the left button shows "Open in Work Container". 

However, if the current tab in which you open the website isn't a container, it will show "Open in Container" because the code returns a null value for the container name.

This patch fixes this by showing a different message for this situation by adding a new l10n string (openInNoContainer).

### DIFF
--- a/src/confirm-page.html
+++ b/src/confirm-page.html
@@ -23,8 +23,21 @@
       </label>
       <br />
       <div class="button-container">
-        <button id="deny" class="button" data-message-id="openInContainer" data-message-arg="current-container-name"></button>
-        <button id="confirm" class="button primary" autofocus data-message-id="openInContainer" data-message-arg="container-name"></button>
+        <button id="deny"
+                class="button"
+                data-message-id="openInContainer"
+                data-message-arg="current-container-name">
+        </button>
+        <button id="deny-no-container"
+                class="button"
+                data-message-id="openInNoContainer">
+        </button>
+        <button id="confirm"
+                class="button primary"
+                autofocus
+                data-message-id="openInContainer"
+                data-message-arg="container-name">
+        </button>
       </div>
     </form>
   </main>
--- a/src/js/confirm-page.js
+++ b/src/js/confirm-page.js
@@ -12,20 +12,37 @@ async function load() {
     denySubmit(redirectUrl);
   });
 
+  document.getElementById("deny-no-container").addEventListener("click", (e) => {
+    e.preventDefault();
+    denySubmit(redirectUrl);
+  });
+
   const container = await browser.contextualIdentities.get(cookieStoreId);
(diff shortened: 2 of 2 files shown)


src/js/confirm-page.js
```diff
@@ -12,22 +12,40 @@ async function load() {
...
+  const buttonDeny = document.getElementById("deny");
+  const buttonDenyNoContainer = document.getElementById("deny-no-container");
+
+  if (currentContainerName) {
+    buttonDenyNoContainer.style.display = "none";
+    return currentContainerName;
+  } else {
+    buttonDeny.style.display = "none";
+    return;
+  }
```
groovecoder, line 46: nit: since the `if` block has a `return` in it, we don't need this `else` block. no logical difference, but I find that code with less "noisy" syntax in it is easier to read.

```suggestion
  }
  buttonDeny.style.display = "none";
  return;
```

src/confirm-page.html
```diff
@@ -23,8 +23,21 @@ <h1 class="title-text" data-i18n-message-id="openThisSiteConfirmation"></h1>
...
-        <button id="deny" class="button" data-message-id="openInContainer" data-message-arg="current-container-name"></button>
-        <button id="confirm" class="button primary" autofocus data-message-id="openInContainer" data-message-arg="container-name"></button>
+        <button id="deny"
+                class="button"
+                data-message-id="openInContainer"
+                data-message-arg="current-container-name">
+        </button>
+        <button id="deny-no-container"
+                class="button"
+                data-message-id="openInNoContainer">
```
groovecoder, line 33: I checked out this branch and did a `git submodule update` but this message ID still isn't in the `src/_locales/en/messages.json` file. Does this branch need another submodule udpate?

groovecoder (changes requested):

src/confirm-page.html
dannycolin: Something's wrong with `git submodule update`. It always checkout to an old commit (`Submodule path 'src/_locales': checked out '26e17db70ce276ab5ccb5ae582e5c142f12ba655'`)

groovecoder: Looks like this is fixed now!

groovecoder (approved):

Issue #2394, opened 2022-07-23
Update devDependencies that have known vulnerabilities


marperia:
### Before submitting a bug report

- [X] I updated to the latest version of Multi-Account Container and tested if I can reproduce the issue
- [X] I searched for existing reports to see if it hasn't already been reported

### Step to reproduce

1. I cloned a repo
2. I ran `npm install --legacy-peer-deps`

### Actual behavior

While the initial set up run the following vulnerabilities are present in node

```
26 vulnerabilities (15 moderate, 8 high, 3 critical)

To address all issues (including breaking changes), run:
  npm audit fix --force
```

### Expected behavior

No vulnerabilities should be present

### Additional informations

Looks like it is a copy of #1261. Should I run `npm audit fix --force` and submit PR?

### Provide a copy of Troubleshooting Information page

```plain text
I suppose you don't need any Information pages
```

marperia:
One important detail: I had to run it three times before there were three "moderate" messages left

dannycolin:
Thanks for the report. I'll look at it this week.

PR #2395, opened 2022-07-23
fix npm vulnerabilities #2394

marperia:
Updated package.json a bit for #2394 issue


ProKashif (approved):

dannycolin:
I'll merge a quick patch that'll bump the web-ext version. This will remove all the critical and high vulnerability alerts. For the moderate vulnerabilities, I need to test a replacement for htmllint that's a bit more maintained so I'll do it in a follow-up PR.

PR #2422, opened 2022-10-06
Fix #2421 - Truncate text to ensure the delete icon is visible

### LINKED ISSUE #2421: Unable to manage site list a URL is too long

### Step to reproduce

I accidentally added a site to my Work container.

I went to remove it by navigating through Manage Containers > Work > Manage Site List...

However, I cannot remove the site as the trash icon is not visible. I beileve this is because another site added previously has an unusually long URL and the icons are in the overflow to the right.
Screenshot (1)

### Actual behavior

Other containers are working normally and the trash icon to remove sites are visible at the right edge of the dropdown, eg
Screenshot (4)

### Expected behavior

If the icon appeared to the left of the URLs, it would be possible to delete them no matter how long the URLs are.

### Additional informations

_No response_

_No response_

dannycolin:
This fix an issue where the user is unable to delete an assignment because the text is overflowing and the delete icon is drawn offcanvas.

### DIFF
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -1427,7 +1427,7 @@ Logic.registerPanel(P_CONTAINER_ASSIGNMENTS, {
         trElement.innerHTML = Utils.escaped`
         <td>
           <div class="favicon"></div>
-          <span title="${site.hostname}" class="menu-text">${site.hostname}</span>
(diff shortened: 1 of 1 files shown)


maxxcrawford (approved):
Nice code-reuse/fix! LGTM.

PR #2428, opened 2022-10-20
Fix broken radio focus styling in light mode

dannycolin:
It was impossible to know what color and icon was selected in manage container because the :focus and :active color weren't correctly applied.


maxxcrawford (approved):
Verified (in both modes)

image

image

Issue #2439, opened 2022-11-02
Sync – Deleting a container in one browser does not remove the container from the other browser


acornestean:
### Before submitting a bug report

- [X] I updated to the latest version of Multi-Account Container and tested if I can reproduce the issue
- [X] I searched for existing reports to see if it hasn't already been reported

### Step to reproduce

1. Install the 8.1.0 version of Multi-Account Containers on 2 browsers and sign in with the same Firefox account on both. Make sure sync is enabled for both instances of the extension. 
2. On browser 1 (B1) create a container
3. On browser 2 (B2) wait for the sync or manually sync and check that the created container appears in the MAC container list
4. On B1, delete the container created at Step 2
5. On B2 wait for the sync or manually sync and check for the presence of the previously deleted container
6. Observe the container is still present on B2 after sync

### Actual behavior

The container which was deleted on B1 still persists on B2 and is not removed i.e. sync does not work with deleted/removed containers.

### Expected behavior

The container which was deleted on B1 should also be removed from B2 after syncing.

### Additional informations

Sync was performed on 2 Nightlies (108.0a1/20221101213659)

acornestean:
@stoically

I tested the currently available public version 8.0.7 and indeed, containers are no longer present on the second Firefox instance after syncing.

However, I’m testing the pre-release version 8.1.0 (https://github.com/mozilla/multi-account-containers/releases/tag/8.1.0) and it seems with this one, containers are still present in the second instance of Firefox after they are deleted from the first one. I followed the instructions you provided in #2240 but to no avail.

I observed that the MAC developer console correctly logs the deleted container (I named it DELETE and I mention this so it can be more easily followed in the console logs).

So as you can see in the image below, the `deletedIdentityList` array on the left (first FX instance) which has data from deleting the “DELETE” container matches what should be on the second instance of FX (on the right). The list of available containers also matches and the “DELETE” container is nowhere to be seen there. 

![2022-11-03_12h39_41](https://user-images.githubusercontent.com/50236075/199721965-b7b2245e-78d6-47a0-9355-7f320aab5490.png)

So far so good. Only issue is that the container is still present in the second instance of FX (see screenshots).

First FX instance:
![2022-11-03_12h40_15](https://user-images.githubusercontent.com/50236075/199722069-1eae11d0-da98-427e-ac42-4412b27f4794.png)

Second FX instance:
[...]

PR #2444, opened 2022-11-11
Backout #2231 from 8.1.0

dannycolin:
We won't be able to have a fix for #2439 before 8.1.0 is released so we need to backout #2231 and #2383.


dannycolin:
Closing in favor of #2450

PR #2458, opened 2022-11-29
Add extension shortcut logic for tab sorting

xenoruon:
Added implementation for using an extension shortcut to trigger the _Sort Tabs by Container_ command (with a bit of refactoring/optimization for the other shortcut extensions).


src/manifest.json
```diff
@@ -43,6 +43,12 @@
       },
       "description": "Open containers panel"
     },
+    "sort_tabs_by_container": {
+      "suggested_key": {
+        "default": "Ctrl+Comma"
```
dannycolin, line 48: We should also add:
```
"mac": "MacCtrl+Comma"
```

src/js/background/backgroundLogic.js
```diff
@@ -15,13 +15,25 @@ const backgroundLogic = {
   unhideQueue: [],
   init() {
     browser.commands.onCommand.addListener(function (command) {
-      for (let i=0; i < backgroundLogic.NUMBER_OF_KEYBOARD_SHORTCUTS; i++) {
```
dannycolin, line 18: If `NUMBER_OF_KEYBOARD_SHORTCUTS` isn't needed anymore, we should remove all references to it in `backgroundLogic.js`.

dannycolin (changes requested):
There's only a few minor changes but other than that it looks good. :)

src/manifest.json
xenoruon: Added the suggested keyboard shortcut for mac.

src/js/background/backgroundLogic.js
xenoruon: Removed `NUMBER_OF_KEYBOARD_SHORTCUTS` from `backgroundLogic.js`; however, since there was a reference to it in `identityState.js` I moved it there.

xenoruon:
Thanks for the feedback!

dannycolin:
Closing in favor of #2492

Issue #2066, opened 2021-06-07
How to remove the "Always open this in a container" icon from the Address/URL bar?


jimaldon:
- Multi-Account Containers Version:
- Operating System + Version:
- Firefox Version:
- Other installed Add-ons + Version + Enabled/Disabled-Status:

### Steps to reproduce
1.
2.
3.

unnamed777:
I'd like to do it too.

whenov:
If you want to remove this button right now, add following code to your userChrome.css:
```css
#pageAction-urlbar-_testpilot-containers {
    display: none !important;
}
```

dougwl:
There is this pull request #2042 waiting for review that address this problem, but the repository seems to be missing some activity lately.

dannycolin:
Closing since pageAction will be removed in MV3

cunlem:
The button is still present in version 8.1.2 but `userChrome.css` method works like a charm

draeath:
Is this going anywhere? It's been over a year now since #2477 was brought up as a resolution for this, and that's still open.

draeath:
I stand corrected, that is now closed.

PR #2477, opened 2022-12-20
MV3: Remove pageAction

dannycolin:
This PR remove pageAction that is deprecated in Manifest V3.


draeath:
Is this going anywhere? It's been over a year now since it's existence was used to close #2066

Can we have that re-opened if this isn't going to be happening any time soon?

PR #2501, opened 2023-02-08
Add new shortcut navigation in the browser panel

dannycolin:
- Add Slash shortcut to focus the filter input
- Use event.code instead of event.keyCode
- Add modifier key detection

### DIFF
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -396,7 +396,11 @@ const Logic = {
   },
 
   shortcutListener(e){
-    function openNewContainerTab(identity) {
+    function openTopContainers() {
+      const identities = Logic.identities();
+      const key = e.code.substring(5);
+      const identity = e.code === "Digit0" ? identities[9] : identities[key - 1];
+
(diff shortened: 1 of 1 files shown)


maxxcrawford (approved):
Works as expected except for the modifier keys logic. 

How can I test this part?

dannycolin:
> Works as expected except for the modifier keys logic.

WFM. Could you pull the newest commit and retry?

> How can I test this part?

For the modifier detection:

1. Open MAC popup
2. Press `[mod] + [0-9]` where `[mod] can be altKey, ctrlKey, shiftKey, metaKey or a combinaison of it.

Expected: Nothing. It should prevent a container to be opened. Only pressing `[0-9]` should still work.

For the search input focus detection:

1. Open MAC popup
2. Press `[0-9]`

Expected: It should prevent a container to be opened and the `[0-9]` character should be inserted in the search input.

maxxcrawford (approved):

Issue #2134, opened 2021-09-28
[Suggestion] Open specific link WITH special schema corresponding to a certain container.


momoblydblk:
### Expected behavior (Examples)
Open link such as `moz-extension://[whatever id is assigned to this extension(the container)]/[container_name]/twitter.com` will open twitter in the specific container.
Example:
Link: `moz-extension://1bbd76ed-631f-4901-ac00-54c1ca5ba151/Personal/twitter.com` will open `twitter.com` in the *Personal* container (if there is a container named that)

> I'm not sure if this is the correct id assigned to this extension by AMO, but this is what I get from my browser, so I decide to stick with it.

### Notes
Will be handy if you have multiple google accounts with their own calendar. One for work and one for personal. You can add this as bookmarks to fastly open the corresponding calendar page.

momoblydblk:
> 
> 
> This would be really awesome to have. Any external tool/app could open urls in a specific container. I was searching for some way to do this. This extension is the main reason why I use firefox.
> 
> This would save a lot o manual work.

The ideal situation I expected for this is to use this with a bookmark. Having a work folder all containing schema to open page in the container and having a personal folder doing the same thing but in a different container. I don't need to mannually open tab in specific containers before clicking the bookmark.

momoblydblk:
So no one from the official noticed this thing?

toupeira:
I currently use https://github.com/honsiorovskyi/open-url-in-container which does something like this, but it would definitely be great to have this in the official addon.

I mainly use this for bookmarks where I use multiple accounts in separate containers, so I can't just set those sites to always open in a specific container since they share the same domain. My own use-case could also be solved by more granular URL patterns (with an optional path, and wildcards), or container settings on bookmarks.

But a custom URL scheme would still be useful for other use-cases like scripting, or shortcuts outside of Firefox.

momoblydblk:
> I currently use https://github.com/honsiorovskyi/open-url-in-container which does something like this, but it would definitely be great to have this in the official addon.
> 
> I mainly use this for bookmarks where I use multiple accounts in separate containers, so I can't just set those sites to always open in a specific container since they share the same domain. My own use-case could also be solved by more granular URL patterns (with an optional path, and wildcards), or container settings on bookmarks.
> 
> But a custom URL scheme would still be useful for other use-cases like scripting, or shortcuts outside of Firefox.

Thanks, it helps a lot

LuisAPI:
[...]

PR #2530, opened 2023-05-11
Add option to ask which container a page should be open in when it has no site assigned to it

BradSharp:
Introduces the select page which prompts the user to select the container they want to use when they open a site without a container already assigned to it. This encourages users to separate out their websites into containers if they wish, a common feature that I want myself and have also seen others ask for.

By default this setting is disabled but it can be enabled by the user if they go into the options menu.

An example of what the select page looks like with default containers:

image

I am not familiar with the localization process but I have edited the json files in the locales submodule. I would appreciate it if someone could explain to me what the intended way to include those is.

### DIFF
--- a/src/css/select-page.css
+++ b/src/css/select-page.css
@@ -1,0 +1,107 @@
+/* General Rules and Resets */
+.title {
+  background-image: none;
+}
+
+main {
+  background: url(/img/onboarding-4.png) no-repeat;
+  background-position: 200px 0;
+  background-size: 120px;
+  margin-inline-start: -350px;
+  padding-inline-start: 350px;
+}
+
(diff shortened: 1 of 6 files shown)


KittyKatt:
Would this possibly fulfill the AC required to satisfy #2535 ? I've been keeping an eye on this issue for months and it feels like if I was asked to select a container for every tab that is not currently assigned to be auto-opened in a container, then the Ctrl-T new tab behavior would work as expected.

unrealsolver:
Would it be possible to have manual rules, as it described here: https://github.com/mozilla/multi-account-containers/issues/2134
The use case is mainly 3rd party apps that does not care about containers when opening a link.

PR #2543, opened 2023-06-18
Fix missing i18n strings in the UI

dannycolin:
This patch fix all the missing i18n strings in the UI. It also standardize the main title to "Firefox Multi-Account Containers" everywhere we use it in the addon.

In `src/js/background/backgroundLogic.js`, we update the i18n strings for the open_container_[0-9] commands on startup and on install. This is done to avoid using both an hardcoded value with a i18n strings (e.g `__MSG_myMessage__ my extra text`). See https://github.com/mozilla-l10n/multi-account-containers-l10n/pull/23

### DIFF
--- a/src/popup.html
+++ b/src/popup.html
@@ -1,7 +1,7 @@
 <html data-theme="auto">
 <head>
   <meta http-equiv="content-type" content="text/html; charset=utf-8">
-  <title>Multi-Account Containers</title>
+  <title>Firefox Multi-Account Containers</title>
   <script type="text/javascript" src="./js/i18n.js"></script>
   <link rel="stylesheet" href="./css/popup.css">
 </head>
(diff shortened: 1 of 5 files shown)


src/popup.html
```diff
@@ -1,7 +1,7 @@
 <html data-theme="auto">
 <head>
   <meta http-equiv="content-type" content="text/html; charset=utf-8">
-  <title>Multi-Account Containers</title>
+  <title>Firefox Multi-Account Containers</title>
```
groovecoder, line 4: question (non-blocking): Should these hard-coded English titles be replaced with translations? I don't see an attribute identifier for JS to do any replacement here.

groovecoder (changes requested):

src/popup.html
dannycolin: AFAIK, we don't translate the product name anywhere (e.g. SUMO KBs) and it's the same for other Mozilla products so that's why I didn't feel like we needed an i18n string here. However, if you think it'd be a good idea, I don't have any strong opinion about it.

groovecoder (approved):

PR #2546, opened 2023-06-20
Update iconography to match Firefox Design System

dannycolin:
This patch updates the iconography based on the "new" Firefox Design System guideline. It includes:

* Update default Multi-Account Containers icon
* Update "Always open in" icon (only for the page action)
* Remove unused icons


unrealsolver:
It would be great if you'll attach screenshots of before/after for all the changes.

PR #2554, opened 2023-08-01
fix: don't preventDefault on modifier keys

### LINKED ISSUE #2553: Can't imput any characters involving a modifier since commit ef1030789
I was testing another issue I will be opening soon, when I noticed I couldn't imput a ':' at all - in none of the text fields. In fact, nothing that needs a modifier was working, even doing Ctrl+v.

Bisecting from the version I'm actually using day-to-day (8.1.2) I reached the commit ef1030789, which changes shortcut handling.

leodag:
Fixes #2553

The preventDefault on modifier keys seemed fishy, removing it solved the problem immediately. The digit shortcuts seem to still be working fine.

### DIFF
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -426,7 +426,6 @@ const Logic = {
 
     if (e.altKey || e.shiftKey || e.ctrlKey || e.metaKey) {
       isModifierPressed = true;
-      e.preventDefault();
     }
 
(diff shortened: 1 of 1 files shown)


dannycolin:
@leodag Thanks again for the bug report. I want with a different solution of completely reverting the code listening to a modifier added in my previous patch (See #2587) so I'll close this one. However, don't hesitate to file a bug if needed.

PR #2555, opened 2023-08-01
fix: remove broken escape from issue template

leodag:
Currently, clicking on 'New Issue' in the issues tab doesn't show the option to actually create an issue because the template is broken (maybe this is a problem with github's parsing?) with the following error:

image

(screenshot from https://github.com/mozilla/multi-account-containers/blob/main/.github/ISSUE_TEMPLATE/bug.yml )

Removing the backslash before a single quote fixes the issue. Even if the issue is with GitHub's parsing, the backslash is redundant anyway.

You can test this on my fork: https://github.com/leodag/multi-account-containers/issues/new/choose


leodag:
Already fixed on 1142c738, closing this.

PR #2557, opened 2023-08-01
Fix proxy parsing

### LINKED ISSUE #2556: Proxy port gets saved as 'undefined' if not provided
If you try to set a proxy for a container, and don't provide a port, it will be saved as undefined, also showing in the UI as such. This also makes the proxy fail silently.

leodag:
Fixes #2556 

You'll probably want to apply #2554 too to test this, since otherwise you can't enter ':' (nor Ctrl+v on the field)

If a match for "port" is not found, it can be set as 1080 (the default SOCKS port).

This may have some implications for the following lines. If Mozilla VPN is set, does it matter whether port is set or not?


src/js/proxified-containers.js
```diff
@@ -50,6 +50,10 @@ proxifiedContainers = {
       return false;
     }
 
+    if (!matches.groups.port) {
+      matches.groups.port = 1080;
```
dannycolin, line 54: IIRC, HTTP(S) have a different default port. Since we're accepting both SOCKS and HTTPS, we should also consider this scenario.

dannycolin (changes requested):

PR #2559, opened 2023-08-03
Apply sorting in page action menu too

Cimbali:
Quite simple, the page action menu was not sorted. This PR applies the sort, to be coherent with other displays of the containers.

### DIFF
--- a/src/js/pageAction.js
+++ b/src/js/pageAction.js
@@ -1,6 +1,14 @@
 async function init() {
   const fragment = document.createDocumentFragment();
-  const identities = await browser.contextualIdentities.query({});
+  const [identities, containerOrderStorage] = await Promise.all([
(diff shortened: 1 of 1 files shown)


dannycolin:
Thanks a lot for your contribution. However, pageAction is deprecated in the new webextension API (Manifest v3). This means that browsers are going to remove the pageAction feature and we will need to remove it from Multi-Account Containers. See #2477 

If there's any other feature you'd like to work on, please feel free to get in touch with us. You can ping me on a specific issue or join our matrix room #containers:mozilla.org on chat.mozilla.org (See https://wiki.mozilla.org/Matrix for instructions).

Cimbali:
Yes actually it’s not even the list I was looking to sort, which is the one when long-pressing the “+” button for new tabs. But that one is not handled by MAC.

PR #2634, opened 2024-04-17
Missing translation for a contextmenu item

bakulf:
For reference: https://github.com/mozilla-l10n/multi-account-containers-l10n/pull/26

### DIFF
--- a/src/js/background/assignManager.js
+++ b/src/js/background/assignManager.js
@@ -764,7 +764,7 @@ window.assignManager = {
   async initBookmarksMenu() {
     browser.contextMenus.create({
       id: this.OPEN_IN_CONTAINER,
-      title: "Open Bookmark in Container Tab",
+      title: browser.i18n.getMessage("openBookmarkInContainerTab"),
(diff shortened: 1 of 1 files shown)


src/js/background/assignManager.js
```diff
@@ -764,7 +764,7 @@ window.assignManager = {
   async initBookmarksMenu() {
     browser.contextMenus.create({
       id: this.OPEN_IN_CONTAINER,
-      title: "Open Bookmark in Container Tab",
+      title: browser.i18n.getMessage("openBookmarkInContainerTab"),
```
flodolo, line 767: Where do I see this string? I have the add-on installed, but don't see it in the context menu for bookmarks.

dannycolin: 1. In MAC preferences, you need to tick "Enable bookmark menus"

image

2. Open Firefox sidebar and select the bookmarks view (Ctrl + B).
3. Right-Click on a bookmark

flodolo (approved):

PR #2674, opened 2024-09-29
Fix #2437: The current panel is not added to the history if it’s alre…

### LINKED ISSUE #2437: Cannot customize containers due to intermittent “Back” button issue

### Step to reproduce

1. Install the 8.1.0 version of Multi-Account Containers
2. Open the Multi-Account Containers pop-up select any available container via the “>” button beside the container
3. Click on “Manage this Container” button
4. Once on the container detail page, rename the container to something else, change the icon and color
5. Click the “Back” button to return to the previous panel
6. Notice that sometimes, the “back” button does not work immediately as it requires multiple clicks in order to return to the previous panel. It, however, instead highlights the container name field and resets it to it’s previous value thus not saving the customization of that container. Color and icon are also not taken into account and reset.

### Actual behavior

The “Back” button does not always work and causes changes to containers to be reset.

### Expected behavior

The “Back” button should return the user to the previous panel and not affect container customization saving.

### Additional informations

There might be a related issue regarding the “Back” button here: https://github.com/mozilla/multi-account-containers/issues/1857.

https://user-images.githubusercontent.com/50236075/199227776-807c0af6-9621-483f-bf80-efa5cffd9c78.mp4

_No response_

axshaykp:
…ady the last panel

# Description

Duplicate entries in the previous panel path were preventing the back button from navigating to the previous panel, which hindered container editing.

## Type of change

*Select all that apply.*

- [x] Bug fix
- [ ] New feature

*
*
*

### DIFF
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -277,7 +277,7 @@ const Logic = {
   },
 
   async showPanel(panel, currentIdentity = null, backwards = false, addToPreviousPanelPath = true) {
-    if ((!backwards && addToPreviousPanelPath) || !this._currentPanel) {
+    if ((!backwards && addToPreviousPanelPath) && this._currentPanel && this._currentPanel !== panel) {
       this._previousPanelPath.push(this._currentPanel);
     }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "testpilot-containers",
-  "version": "8.1.3",
+  "version": "8.2.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "testpilot-containers",
-      "version": "8.1.1",
+      "version": "8.2.0",
       "license": "MPL-2.0",
       "devDependencies": {
         "addons-linter": "^5.28.0",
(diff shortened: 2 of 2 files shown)


src/js/popup.js
```diff
@@ -277,7 +277,7 @@ const Logic = {
   },
 
   async showPanel(panel, currentIdentity = null, backwards = false, addToPreviousPanelPath = true) {
-    if ((!backwards && addToPreviousPanelPath) || !this._currentPanel) {
+    if ((!backwards && addToPreviousPanelPath) && this._currentPanel && this._currentPanel !== panel) {
```
dannycolin, line 280: While the change seems to work in this specific step-to-reproduce, it introduces an issue in the conditional logic in the following scenario:

1. Open popup to show containerList
1. Click on the right arrow of the Personal container to go to the containerInfo panel
1. In containerInfo panel, click on "manage this container" button
1. In containerEdit panel, click on the back button to go to the previous panel containerInfo
1. In containerInfo panel, click on the back button to go to the previous panel containerList

On the last step, `backwards` value is set to `false` but it should be set to `true` since we clicked on the back button.

dannycolin (changes requested):

PR #2683, opened 2024-10-24
Update webExt package

bakulf:


src/_locales
dannycolin, line 1: Can you remote _locales from the PR please?

bin/addons-linter.sh
```diff
@@ -9,27 +9,9 @@
...
-print G "done."
-
-printn Y "Copying data... "
-cp -r src $TMPDIR || die
-print G "done."
-
-printn Y "Removing the github folder... "
-rm -rf $TMPDIR/src/_locales/.github || die
-print G "done."
-
```
dannycolin, line 33: You originally added this to avoid the following linting failure:

```
Code                          Message                    Description                                        File            Line   Column
NO_MESSAGES_FILE_IN_LOCALES   Empty language directory   messages.json file missing in "_locales/.github"   manifest.json
```

Do you think it's fine to let the linter complain about it or would there be a cleaner solution to avoid this error?

bin/addons-linter.sh
```diff
@@ -9,27 +9,9 @@
...
 print Y "Running the test..."
-npx addons-linter $TMPDIR/src || die
+npx addons-linter ./src || die
```
dannycolin, line 17: Nitpicking but I'd prefer if we could run it from any subfolder instead of using `./`.

PR #2710, opened 2025-01-16
Fix #1670 - Add option to manually assign an URL to a container

### LINKED ISSUE #1670: Manually add URL's to a container
- Multi-Account Containers Version: 6.2.3
- Operating System + Version: Kubuntu 19.10
- Firefox Version: Nightly `Mozilla/5.0 (X11; Linux x86_64; rv:75.0) Gecko/20100101 Firefox/75.0 ID:20200229214828`
<details>
<summary>Other installed Add-ons + Version + Enabled/Disabled-Status (Click to expand/hide)</summary>

About Sync 0.0.51
Amazon.com 1.1
Authenticator 6.0.1
Awesome RSS 1.3.5
Bing 1.1
CanvasBlocker 1.1 [DISABLED]
Certainly Something (Certificate Viewer) 1.2.3
Compat Report 0.1.0
Containers theme 1.0.3 [DISABLED]
Context Plus 0.6.0
Debian queries 2.3
Decentraleyes 2.0.13
Disconnect 5.19.3 [DISABLED]
DuckDuckGo 1.0
DuckDuckGo Privacy Essentials 2019.12.12 [DISABLED]
Emoji Cheatsheet for GitHub, Basecamp etc. 2.0.1
Enhancer for YouTube™ 2.0.99
Enterprise Policy Generator 5.1.0 [DISABLED]
Feedbro 4.0.3
Firefox DevTools ADB Extension 0.0.5
Firefox Lightbeam 2.1.0
Firefox Multi-Account Containers 6.2.3
FoxyProxy Standard 7.4.3
Google 1.0
Greasemonkey 4.9
HTTP/2 Indicator 3.0
IPFS Companion 2.10.0
Laboratory 3.0.7 [DISABLED]
MathML Font Settings 0.3.2
Native MathML 1.9.4
Nightly Tester Tools 4.0
Notes by Firefox 4.3.5 [DISABLED]
Offline QR Code Generator 1.6
PassFF 1.9.1 [DISABLED]
Plasma Integration 1.7
ProtonDB for Steam 1.6.0
React Developer Tools 4.4.0 [DISABLED]
Session Sync 3.1.12
Snooze Tabs 1.1.1
Tab Session Manager 5.2.0 [DISABLED]
To Google Translate 4.0.6
Tree Style Tab 3.3.6
True Sight 1.7.1
Twitter 1.0
Vue.js devtools 5.3.3
Wikipedia (en) 1.0
YouTube Container 1.3.0
eBay 1.0
uMatrix 1.4.0

</details>

### Actual behavior
No way to add redirecting URL's.
- **If you DON'T FORGET TO DO THIS EVERY TIME:**
  - Only way at moment is to use right-click and choose a container.
  - Else you are screwed...

### Expected behavior
To automatically open known redirect URI's we need a EASY way to add them to a container.
Sometimes you want to catch redirect links, to force them to always open in a certain container **before** the browser navigates to the destination of the link.
For this purpose we **NEED** a way to manually add URL's to a container of choice !

### Steps to reproduce
1. Try to add a rule using the UI for links like `https://go.microsoft.com/fwlink/?LinkID=xxxxx&wt.mc_id=fuckmedead`
2. Stuck because there is no way.
3. Write bug report on github, like i do now :wink: 

### Notes
Related issues: #719, #837, #839, #1057, #1075, #1180, #1335, #1317 and maybe many more im tired search now...

dannycolin:

- [ ] I rebased my work on top of the main branch.
- [ ] I ran `npm test` and all tests passed.

# Description

Add an option in "manage site list" panel to manually assign an URL to a container

## Remaining tasks

- [ ] Add a message when the list is empty
- [ ] Move hardcoded style values in popup.js to popup.css
- [ ] Move hardcoded strings to /locales
- [ ] ???
- [ ] Profit

## Type of change

*Select all that apply.*

- [ ] Bug fix
- [x] New feature

* #1670

### DIFF
--- a/src/css/popup.css
+++ b/src/css/popup.css
@@ -89,9 +89,13 @@
   --input-border-color: #8f8f9d;
   --input-border-focus-color: #0060df;
   --input-border-active-color: #054096;
+  --input-destructive-border-color: rgba(226, 40, 80, 1);
   --text-color-primary: #15141a;
   --menu-bg-active-color: #cfcfd8;
   --menu-bg-hover-color: #e0e0e6;
+  --messagebar-destructive-bg-color: rgba(255, 232, 232, 1);
+  --messagebar-destructive-border-color: rgba(0, 0, 0, 0.08);
+  --messagebar-destructive-text-color: rgba(21, 20, 26, 1);
   --panel-bg-color: #fff;
   --panel-separator-color: rgba(240, 240, 244, 1);
   --link-color: #0060df;
@@ -154,11 +158,15 @@
   --input-border-color: #8f8f9d;
   --input-border-active-color: #aaf2ff;
   --input-border-focus-color: #0df;
+  --input-destructive-border-color: rgba(255, 132, 128, 1);
   --link-color: #0df;
   --link-hover-color: #80ebff;
   --text-color-primary: #fbfbfe;
   --menu-bg-active-color: #5b5b66;
   --menu-bg-hover-color: #52525e;
+  --messagebar-destructive-bg-color: rgba(105, 15, 34, 1);
+  --messagebar-destructive-border-color: rgba(255, 255, 255, 0.08);
+  --messagebar-destructive-text-color: rgba(251, 251, 254, 1);
   --panel-bg-color: #42414d;
   --panel-separator-color: rgba(82, 82, 94, 1);
 
@@ -447,6 +455,7 @@ table {
 
 .button {
   background-color: var(--button-bg-color-secondary);
+  border: none;
   color: var(--text-color-primary);
 }
 
@@ -2440,4 +2449,38 @@ tr:hover > td > .reset-button {
 
 .searchbar input {
   inline-size: 100%;
+}
+
+/* Assignment panel */
+#edit-sites-new-assignment {
+  display: flex;
+  flex-direction: column;
+  gap: 8px;
+  margin: 8px;
+}
+
+#edit-sites-add-assignment {
+  block-size: 32px;
+  border: none;
+  border-radius: 4px;
+  font-family: inherit;
+  font-weight: bold;
+  text-align: center;
+}
+
+#edit-sites-add-assignment-info {
+
+}
+
+#edit-sites-add-assignment-error {
+  background-color: var(--messagebar-destructive-bg-color);
+  border: 1px solid var(--messagebar-destructive-border-color);
+  border-radius: 4px;
+  color: var(--messagebar-destructive-text-color);
+  display: none;
+  padding: 8px;
+}
+
+#edit-sites-add-assignment:hover {
+  cursor: pointer;
 }
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -129,7 +129,7 @@ const Logic = {
     notificationCards.forEach(notificationCard => {
       notificationCard.textContent = text;
       notificationCard.classList.add("is-shown");
-    
+
       setTimeout(() => {
         notificationCard.classList.remove("is-shown");
       }, 2000);
@@ -1454,6 +1454,35 @@ Logic.registerPanel(P_CONTAINER_ASSIGNMENTS, {
     return Promise.resolve(null);
   },
 
+  async addAssignment() {
+    const errorMessage = document.getElementById("edit-sites-add-assignment-error");
+    const assignmentInput = document.getElementById("edit-sites-add-assignment-info");
+
+    let url = document.getElementById("edit-sites-add-assignment-info").value;
+
+    if (!URL.canParse(url)) {
+      url = "http://" + url;
+    }
+
+    if (new URL(url).protocol == "https:" || new URL(url).protocol == "http:") {
+      url = new URL(url);
+    } else {
+      errorMessage.style.display = "block";
+      assignmentInput.style.border = "2px solid var(--input-destructive-border-color)";
+    }
+
+    const userContextId = Logic.currentUserContextId();
+    await Utils.setOrRemoveAssignment(
+      false, url.origin, userContextId, false
+    );
+
+    const assignments = await Logic.getAssignmentObjectByContainer(userContextId);
+    this.showAssignedContainers(assignments);
+
+    errorMessage.style.display = "none";
+    assignmentInput.style.border = "1px solid var(--input-border-color)";
+  },
+
   showAssignedContainers(assignments) {
     const closeContEl = document.querySelector("#close-container-assignment-panel");
     Utils.addEnterHandler(closeContEl, () => {
@@ -1464,6 +1493,12 @@ Logic.registerPanel(P_CONTAINER_ASSIGNMENTS, {
     const assignmentPanel = document.getElementById("edit-sites-assigned");
     const assignmentKeys = Object.keys(assignments);
     assignmentPanel.hidden = !(assignmentKeys.length > 0);
+
+    const addAssignmentButton = document.getElementById("edit-sites-add-assignment");
+    Utils.addEnterHandler(addAssignmentButton, async () => {
+      this.addAssignment();
+    });
+
     if (assignments) {
       const tableElement = document.querySelector("#edit-sites-assigned");
       /* Remove previous assignment list,
@@ -1487,6 +1522,7 @@ Logic.registerPanel(P_CONTAINER_ASSIGNMENTS, {
           <img title="${deleteSiteInfo}" class="trash-button delete-assignment"  src="/img/container-delete.svg" />
         </td>`;
         trElement.getElementsByClassName("favicon")[0].appendChild(Utils.createFavIconElement(assumedUrl));
+
         const deleteButton = trElement.querySelector(".trash-button");
         Utils.addEnterHandler(deleteButton, async () => {
           const userContextId = Logic.currentUserContextId();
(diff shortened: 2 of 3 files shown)


src/js/popup.js
```diff
@@ -1454,6 +1454,35 @@ Logic.registerPanel(P_CONTAINER_ASSIGNMENTS, {
...
+    const userContextId = Logic.currentUserContextId();
+    await Utils.setOrRemoveAssignment(
+      false, url.origin, userContextId, false
+    );
+
+    const assignments = await Logic.getAssignmentObjectByContainer(userContextId);
+    this.showAssignedContainers(assignments);
+
+    errorMessage.style.display = "none";
+    assignmentInput.style.border = "1px solid var(--input-border-color)";
```
achernyakevich-sc, line 1483: Looks like this block should be inside ` if (new URL(url).protocol == "https:" || new URL(url).protocol == "http:") {` because if condition fail assignment is not required.

Though probably more correct would be to do not make any assumptions about protocol, but expect from user that if he manually add URL then it would be fully qualified URL containing protocol. So in general code flow could looks like (it is not a real code but pseudo-code):

```JavaScript
let url = document.getElementById("edit-sites-add-assignment-info").value;

try {
    let urlObj = new URL(url);

        if (urlObj.protocol == "https:" || urlObj.protocol == "http:") {
        const userContextId = Logic.currentUserContextId();
        await Utils.setOrRemoveAssignment(
            false, urlObj.origin, userContextId, false
        );

        const assignments = await Logic.getAssignmentObjectByContainer(userContextId);
        this.showAssignedContainers(assignments);

        errorMessage.style.display = "none";
        assignmentInput.style.border = "1px solid var(--input-border-color)";
    }
} catch {
    errorMessage.style.display = "block";
    assignmentInput.style.border = "2px solid var(--input-destructive-border-color)";
}
```

TriMoon: @achernyakevich-sc 
Why not use the below instead of your if clause? :wink: 
```javascript
if (/^http(s)?::/.test(url)) {
```

TriMoon:
Looking forward for this work guys :+1:

src/js/popup.js
achernyakevich-sc: > Why not use the below instead of your if clause? 😉

At first it is "pseudo-code" as I mentioned in the comment. :) I just used code that already exists and regrouped it to show more correct execution flow.

But I think RegExp is not necessary here and now. MAC works for HTTP and HTTPS protocols only (AFAIK). And now code is "easily readable". We don't need to spend effort to care about that will never happen. Use KISS approach - Keep It Simple Stupid.

P.S.> One of my England customers said years ago - if I have a car that can drive to China it does not mean that I need to travel to Beijing.

src/js/popup.js
```diff
@@ -1454,6 +1454,35 @@ Logic.registerPanel(P_CONTAINER_ASSIGNMENTS, {
...
+    const errorMessage = document.getElementById("edit-sites-add-assignment-error");
+    const assignmentInput = document.getElementById("edit-sites-add-assignment-info");
+
+    let url = document.getElementById("edit-sites-add-assignment-info").value;
+
+    if (!URL.canParse(url)) {
+      url = "http://" + url;
+    }
+
+    if (new URL(url).protocol == "https:" || new URL(url).protocol == "http:") {
```
TriMoon, line 1467:
```js
if (/^http(s)?::/.test(url)) {
```
Better choice...

src/js/popup.js
```diff
@@ -1454,6 +1454,35 @@ Logic.registerPanel(P_CONTAINER_ASSIGNMENTS, {
...
+      url = new URL(url);
+    } else {
+      errorMessage.style.display = "block";
+      assignmentInput.style.border = "2px solid var(--input-destructive-border-color)";
+    }
+
+    const userContextId = Logic.currentUserContextId();
+    await Utils.setOrRemoveAssignment(
+      false, url.origin, userContextId, false
```
TriMoon, line 1476: `url.origin` can be non-existant/`undefined` due to the check above...

TriMoon (changes requested):

src/js/popup.js
TriMoon: It must be personal preference then i guess, but i bet my code runs faster with less extra resource usage... :stuck_out_tongue:

dannycolin: @achernyakevich-sc At first, I used exactly what you're proposing with a `try...catch`. However, I wanted the user to have the ability to omit http(s) similar to the previous PR for this feature. In this case, I had to check the URL validity and force one if it's missing. I feel like it makes it more intuitive for users that don't know that valid URLs need a protocol.

@TriMoon It's more readable as @achernyakevich-sc said. Also, the performane is negligeable in this case.

TriMoon: @dannycolin Anything is negligible on PC's these days with their power, when you look at each one on it's own, think about the bigger total :wink: 

But anyhow did you notice https://github.com/mozilla/multi-account-containers/pull/2710#discussion_r1920469373 :thinking:

dannycolin: This very specific case is fine. I'll try my best to look at your other comment during the weekend and make sur to report back on GitHub ;).

achernyakevich-sc: > However, I wanted the user to have the ability to omit http(s) similar to the previous PR for this feature. In this case, I had to check the URL validity and force one if it's missing.

@dannycolin It radically changes how we need to implement it! Below I will try to summarize how it would be simpler (to my mind) for understanding by users and simpler for implementation. :)

I have made some research of MAC behaviour and didn't check the code. So I will make some assumptions that proposal below is based on (to lazy to check the code base). The main one - MAC records not a whole URL of site, not protocol+host name, but just site origin. It means that though MAC works for HTTP and HTTPS, but it will work the same for both URLs - http://example.com/ and https://example.com/. Conclusion - UI should not only provide possibility to omit protocol part, but should force to enter site origin without protocol and path name. Only site origin!

Currently saving is triggered by button click. As soon as user have made click then it should go the following way:
* take `url` string that leading `http://` or `https://` will be removed.
* get `urlObject` based on cleaned `url` and added explicit `https://`.
* as we explicitly added `https://` we don't need checking for protocol.
* get site origin from `urlObject` and expose it to UI in input as cleaned up.

So user will see on UI exactly that will be saved as configuration and used for triggering container switching.

In this case code could looks like:

```JavaScript
let urlInput = document.getElementById("edit-sites-add-assignment-info")
let url = urlInput.value.replace(/^http(s)?:\/\//, "");

try {
    let urlObject = new URL("https://" + url);
    urlInput.value = urlObject.origin;

    const userContextId = Logic.currentUserContextId();
    await Utils.setOrRemoveAssignment(
        false, urlObject.origin, userContextId, false
    );

    const assignments = await Logic.getAssignmentObjectByContainer(userContextId);
    this.showAssignedContainers(assignments);

    errorMessage.style.display = "none";
    assignmentInput.style.border = "1px solid var(--input-border-color)";
} catch {
    errorMessage.style.display = "block";
    assignmentInput.style.border = "2px solid var(--input-destructive-border-color)";
}
```

To make it even more explicit you could set placeholder text for the input as `site origin` and it will be visible before user enter anything there. :)

achernyakevich-sc: > Anything is negligible on PC's these days with their power, when you look at each one on it's own, think about the bigger total

@TriMoon It is not a problem if something takes more time almost uncatchable for user but it as well takes electricity. I think you will agree that:
* any operation on PC cause electricity costs.
* wasting of electricity increases carbon footprints and produces a damage to nature.

Problem is that small destructive things happen many-many times! Firefox and MAC is popular. :)

For us it costs almost nothing to make piece of code more electricity-effective. But resulting effect to nature could be very valuable.

dannycolin: @achernyakevich-sc I agree. It makes sense to drop the protocol detection since we don't store it anyway. I'll refactor the patch to match your proposal. Thanks!

src/js/popup.js
dannycolin: Closing since we won't need to detect the protocol after all.

src/js/popup.js
dannycolin: What "check" are you referring to?

TriMoon:
```javascript
   if (new URL(url).protocol == "https:" || new URL(url).protocol == "http:") {
      url = new URL(url);
   } else {
```
As you can see you change `url` here to be an `URL()`, but before this check `url` was a `string`

src/css/popup.css
```diff
@@ -2441,3 +2450,37 @@ tr:hover > td > .reset-button {
 .searchbar input {
   inline-size: 100%;
 }
+
+/* Assignment panel */
+#edit-sites-new-assignment {
+  display: flex;
+  flex-direction: column;
+  gap: 8px;
+  margin: 8px;
```
kurahaupo, line 2459: `margin` disallowed by property-disallowed-list rule

src/css/popup.css
```diff
@@ -2441,3 +2450,37 @@ tr:hover > td > .reset-button {
...
+
+}
+
+#edit-sites-add-assignment-error {
+  background-color: var(--messagebar-destructive-bg-color);
+  border: 1px solid var(--messagebar-destructive-border-color);
+  border-radius: 4px;
+  color: var(--messagebar-destructive-text-color);
+  display: none;
+  padding: 8px;
```
kurahaupo, line 2481: `padding` disallowed by property-disallowed-list rule

src/css/popup.css
```diff
@@ -2441,3 +2450,37 @@ tr:hover > td > .reset-button {
...
+  block-size: 32px;
+  border: none;
+  border-radius: 4px;
+  font-family: inherit;
+  font-weight: bold;
+  text-align: center;
+}
+
+#edit-sites-add-assignment-info {
+
```
kurahaupo, line 2472: empty CSS block disallowed by block-no-empty rule

Issue #1309, opened 2018-11-14
Undo "Sort Tabs"


theres-waldo:
I had a Firefox window with a lot of tabs (~70), and I had them neatly arranged into a hierarchy using Tree Style Tabs.

I accidentally pressed the "Sort Tabs" button in the popup dialog that comes up when pressing the Multi-Account Container extension's toolbar button, and now tab tree has been completely messed up. 

It would be nice to have a way to undo this operation.

sgautrin:
I just got the same thing. Now focus in tree style tab keep jumping everywhere because the order shown by tree style tab is completely different than the one in the background I guess. Painful with my nearly 900 tabs opened.

Disabling and re-enabling tree style tab might work, but then i'll probably lose the tabs hierarchy.

hiljusti:
I've also made this same mistake (also with tree-style-tabs and containers) and my mitigation now is to just never push that button. That said, an "undo" option if I hit it by mistake would be a life-saver

The ability to _**disable**_ the "Sort Tabs" button in some configuration file or menu would help me out. (This feels like an advanced option that general users do not need). Should I open that as a separate issue?

6a6kin:
This can be solved in a pretty simple way: just add one are-you-sure prompt with Yes/No before sort tabs run.

This won't require any heavy development like undo or new configuration options and save from occidental clicks (yes, I've done that with my ~100 tabs, too 😁 )

grahamperrin:
> … add one are-you-sure prompt …

With respect: let's not add such things. 

Instead, place the feature in an area where accidents will be less likely. 

Maybe adjacent to, or after, **Manage Containers**.

TheBrain0110:
Just had this exact same thing happen, and of course the first thing I find is an open issue about it that's been around for 4 years...

Is there even any ongoing development of this feature at all?

oshanz:
please add "disable" or "hide" option to this sort tabs button.
I have 100+ tabs and this button got clicked accidentally many times.  :tired_face:
if you can provide a hack that's also a great help. :weary:

zyansheep:
This just happened to me... all my tabs... all gone :sob:
Please add an option to hide the button. If you point me at the right files I'll even submit a PR myself!

dannycolin:
> This just happened to me... all my tabs... all gone 😭 Please add an option to hide the button. If you point me at the right files I'll even submit a PR myself!

[...]

PR #2714, opened 2025-01-26
#1309 add confirm sorting dialoge

Parvchannel:
Hello everyone!
This is my first pull request on GitHub, and I’d greatly appreciate your feedback.

Feature: Adds confirmation before sorting tabs (#1309).
Before: Clicking the sort button instantly sorts all open tabs.
After: A confirm() dialog is shown before sorting, and tabs are sorted only if "Yes" is selected.

How it works:

1)The user triggers the action in popup.js.
2)A message is sent to contentscript.js to display the dialog.
3)If the content script is not allowed on the page, it skips the dialog and directly sorts the tabs.
4)contentscript.js responds with true or false.
5)Based on the response, popup.js proceeds with sorting or stops.

Issue: I'm not receiving any response from contentscript.js, so I’m unable to proceed.

Looking forward to your feedback—thanks in advance!

### DIFF
--- a/src/js/content-script.js
+++ b/src/js/content-script.js
@@ -45,3 +45,10 @@ async function addMessage(message) {
 browser.runtime.onMessage.addListener((message) => {
   addMessage(message);
 });
+
+browser.runtime.onMessage.addListener(message => {
+  if (message.action === 'confirmSortDialogue') {
+    const result = confirm("confirm sorting tabs");
+    return Promise.resolve({result});
+  }
+});
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -756,16 +756,22 @@ Logic.registerPanel(P_CONTAINERS_LIST, {
     Utils.addEnterHandler(document.querySelector("#always-open-in"), () => {
       Logic.showPanel(ALWAYS_OPEN_IN_PICKER);
     });
+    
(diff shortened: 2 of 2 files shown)


dannycolin:
Hi @Parvchannel,

Thank you for your contribution! Some pages are considered priviliged (e.g about:preferences) and this is why you cannot inject your contentscript in them.

A solution that works on any page is to move the confirmation dialog in the addon popup itself. You can see an example of this behaviour when deleting a container. This would make sure the sort tabs button action is consistent at all time. Plus, it would follow a known pattern that the user is already accustomed when using this addon.

Here are the links to the P_CONTAINER_DELETE panel to give you an example of how to implement a solution of confirm dialog inside the popup:

- https://github.com/mozilla/multi-account-containers/blob/037a8047259d9c485b5e2150edfd142e2e6aab35/src/js/popup.js#L32
- https://github.com/mozilla/multi-account-containers/blob/main/src/js/popup.js#L1969
- https://github.com/mozilla/multi-account-containers/blob/main/src/js/popup.js#L2332

If you have any other question, feel free to ask.

Parvchannel:
Hey, I'm almost done with the new changes and have implemented the suggested updates. Before I proceed with translating it into all 50 languages, I’d like to confirm if everything looks good. Let me know—thanks!         
Screenshot 2025-02-06 184715

PR #2732, opened 2025-03-19
Link to local installation instructions in CONTRIBUTING.md

kurahaupo:

- [X] I agree to license my code under the [MPL 2.0 license](https://www.mozilla.org/en-US/MPL/2.0/).
- [X] I rebased my work on top of the main branch.
- ~[ ] I ran `npm test` and all tests passed.~ (n/a)
- ~[ ] I added test coverages if relevant.~ (n/a)

# Description

"follow the instructions" ... what instructions?

That had been bugging me for a while, then today I came across this page in Mozilla Dev.

## Type of change

*Select all that apply.*

- [X] Meta-documentation improvement
- [ ] Bug fix
- [ ] New feature


CONTRIBUTING.md
```diff
@@ -35,7 +35,7 @@ You can then [open a pull request][pr] on [the l10n repository][l10n].
 ## Tips for contributing
 
 1. Choose [an issue][issues] that you would like to work on.
-2. Fork the repository and follow the instructions for setting it up locally.
+2. Fork the repository and follow the [instructions for setting it up locally](https://extensionworkshop.com/documentation/develop/temporary-installation-in-firefox/).
```
loganrosen, line 38: Would it be better to recommend that contributors run `npm run dev`? That uses `web-ext` to automatically install the add-on in a temporary instance of Firefox.

kurahaupo: `npm: command not found`

I'm sure `npm` will be the simpler option for many, but without a gazillion gigabytes of RAM, running extra instances of Firefox is painful. I'd rather just install it in my normal instance, so that I can test it with my normal workflow using all my other plugins.

So how about mentioning both options?

kurahaupo: I'm aiming to simply add a link to the existing text, not to rewrite the text, since that's a job for someone who understands it much better than me. But in the meantime, having *some* instructions is better than none.

PR #2753, opened 2025-04-22
Fix #2746: sort only ungrouped tabs

### LINKED ISSUE #2746: Sort tabs should keep tabs within their tab groups

- [x] I updated to the latest version of Multi-Account Container and tested if I can reproduce the issue
- [x] I searched for existing reports to see if it hasn't already been reported

### Step to reproduce

1. Enable `browser.tabs.groups.enabled` in about:config to enable Firefox's native tab groups feature (if not enabled)
2. Create 4 tabs.
3. Tabs 0, 1, and 3 should be added to the "Personal" container
4. Tabs 1 and 2 should be added to a new tab group called "test group"
5. Press Ctrl + Comma to sort the tabs by container

### Actual behavior

The tab strip ends up ordered as tabs 2, 0, 1, 3. The tab group "test group" contains tabs 2 and 0 instead of tabs 1 and 2.

### Expected behavior

The tab group "test group" should continue to contain tabs 1 and 2.

### Additional informations

Currently, moving a tab to the position of a grouped tab will add the tab to the tab group. Similarly, moving a tab to the position of an ungrouped tab will remove the tab from any tab group it was previously part of. The Firefox WebExtensions tabs API currently lacks any features for tab groups.

I believe most users who use both Firefox's native tab groups and the Multi-Account Container addon would prefer to keep all tab group membership intact when sorting their tabs. I'd propose something like:

1. Sort all pinned tabs (same as current behavior)
2. Sort all unpinned and ungrouped tabs
3. For each tab group, sort its grouped tabs

Once the Firefox WebExtensions tabGroups API is available, it would also be possible to add another step:

4. For each tab group, place it in the tab strip based on its first tab's container

These changes effectively break the previous guarantees for the "sort tabs" feature -- if we privilege tab groups, then "sort tabs" may not always end up with all tabs in the some container being contiguous. However, I think most users would prefer to maintain their tab group membership in a similar way that most users would prefer to maintain their pinned tabs membership.

```plain text

```

apostrophest:

# Description

Fixes #2746

Firefox 137 introduced tab groups. Tab group web extension support is rolling out in Firefox 138 and later. Tab movements, like the movements done when sorting tabs by container, can inadvertently add or remove tabs from tab groups.

In order to keep users' tab groups intact, this patch only sorts tabs outside of tab groups. Due to the lack of the tabGroups web extensions API as of Firefox 138, this patch cannot move tab groups, so all ungrouped tabs move to the end of the tab strip. That means after sorting, all tab groups will move to the beginning of the tab strip.

## Type of change

*Select all that apply.*

- [x] Bug fix
- [x] New feature

* #2746

### DIFF
--- a/src/js/background/backgroundLogic.js
+++ b/src/js/background/backgroundLogic.js
@@ -3,6 +3,13 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const DEFAULT_TAB = "about:newtab";
+/**
+ * Firefox does not yet have the `tabGroups` API, which exposes this constant
+ * to indicate that a tab is not in a tab group. But the Firefox `tabs` API
+ * currently returns this constant value for `Tab.groupId`.
+ * @see https://searchfox.org/mozilla-central/rev/3b95c8dbe724b10390c96c1b9dd0f12c873e2f2e/browser/components/extensions/schemas/tabs.json#235
+ */
+const TAB_GROUP_ID_NONE = -1;
 
 const backgroundLogic = {
   NEW_TAB_PAGES: new Set([
@@ -343,7 +350,13 @@ const backgroundLogic = {
     let pos = 0;
 
     // Let's collect UCIs/tabs for this window.
+    /** @type {Map<string, {order: string, tabs: Tab[]}>} */
     const map = new Map;
+    /** @type {boolean} */
+    let lastTabIsInTabGroup = tabs.length > 0
+      && tabs[tabs.length - 1].groupId
+      && tabs[tabs.length - 1].groupId !== TAB_GROUP_ID_NONE;
+
     for (const tab of tabs) {
       if (pinnedTabs && !tab.pinned) {
         // We don't have, or we already handled all the pinned tabs.
@@ -353,6 +366,11 @@ const backgroundLogic = {
       if (!pinnedTabs && tab.pinned) {
         // pinned tabs must be consider as taken positions.
         ++pos;
+        continue;
+      }
+
+      if (tab.groupId && tab.groupId !== TAB_GROUP_ID_NONE) {
+        // Skip over tabs in tab groups until it's possible to handle them better.
         continue;
       }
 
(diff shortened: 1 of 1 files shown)


src/js/background/backgroundLogic.js
```diff
@@ -343,7 +350,13 @@ const backgroundLogic = {
     let pos = 0;
 
     // Let's collect UCIs/tabs for this window.
+    /** @type {Map<string, {order: string, tabs: Tab[]}>} */
     const map = new Map;
+    /** @type {boolean} */
+    let lastTabIsInTabGroup = tabs.length > 0
+      && tabs[tabs.length - 1].groupId
+      && tabs[tabs.length - 1].groupId !== TAB_GROUP_ID_NONE;
```
Rob--W, line 358: Although in practice `0` is not used as a value, in theory 0 could also be a valid groupId.

All tab groups have non-negative IDs, and `undefined >= 0` is always false, so you can simplify this logic to:

```suggestion
    let lastTabIsInTabGroup = tabs.at(-1)?.groupId >= 0;
```

src/js/background/backgroundLogic.js
```diff
@@ -356,6 +369,11 @@ const backgroundLogic = {
         continue;
       }
 
+      if (tab.groupId && tab.groupId !== TAB_GROUP_ID_NONE) {
```
Rob--W, line 372:
```suggestion
      if (tab.groupId >= 0) {
```

Similar to above. And with this, you can also drop the `TAB_GROUP_ID_NONE` above.

src/js/background/backgroundLogic.js
```diff
@@ -377,15 +395,24 @@ const backgroundLogic = {
...
-      for (const tab of obj.tabs) {
+    for (const { tabs } of sortMap.values()) {
+      for (const tab of tabs) {
         ++pos;
         browser.tabs.move(tab.id, {
           windowId: windowObj.id,
-          index: pos
+          index: pinnedTabs ? pos : -1
         });
+        if (lastTabIsInTabGroup && browser.tabs.ungroup) {
```
Rob--W, line 405:
```suggestion
        // Pinned tabs are never grouped and always inserted in the front.
        if (!pinnedTabs && lastTabIsInTabGroup && browser.tabs.ungroup) {
```

src/js/background/backgroundLogic.js
```diff
@@ -3,6 +3,13 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const DEFAULT_TAB = "about:newtab";
+/**
+ * Firefox does not yet have the `tabGroups` API, which exposes this constant
```
Rob--W, line 7: FYI `tabGroups` API is coming in 139, see dependencies of: https://bugzilla.mozilla.org/show_bug.cgi?id=1940631

Rob--W (approved):
The logic here works; I have a few small suggestions.

src/js/background/backgroundLogic.js
apostrophest: Looks like the eslint configuration for this project uses the ECMAScript 2018 parser but the optional chaining operator was specified in ECMAScript 2020. I'm going to use `Array.at` and check the result instead of using `?.`. I'll open a separate PR to bump up the parser level of the eslint config.

Rob--W (approved):

Rob--W (approved):

Issue #2735, opened 2025-03-24
Uncaught (in promise) Error: Cannot find menu item with id open-in-this-container (and more) since Firefox 136


Rob--W:
### Before submitting a bug report

- [x] I updated to the latest version of Multi-Account Container and tested if I can reproduce the issue
- [x] I searched for existing reports to see if it hasn't already been reported

### Step to reproduce

1. Install the extension https://addons.mozilla.org/en-US/firefox/addon/multi-account-containers/ (currently latest version is 8.2.0)
2. Look at the Console (Ctrl+Shift+J, Multiprocess), or go to `about:debugging`, This Firefox, Inspect the extension and look at the Console.

### Actual behavior

Logspam, like the following:

```
18:46:13.455 Uncaught (in promise) Error: Cannot find menu item with id open-in-this-container undefined
18:46:13.455 Uncaught (in promise) Error: Cannot find menu item with id remove-open-in-this-container undefined
18:46:13.455 Uncaught (in promise) Error: Cannot find menu item with id separator undefined
18:46:13.456 Uncaught (in promise) Error: Cannot find menu item with id hide-container undefined
18:46:13.456 Uncaught (in promise) Error: Cannot find menu item with id move-to-new-window-container undefined
18:46:13.457 Uncaught (in promise) Error: Cannot find menu item with id open-in-this-container undefined
18:46:13.457 Uncaught (in promise) Error: Cannot find menu item with id remove-open-in-this-container undefined
18:46:13.457 Uncaught (in promise) Error: Cannot find menu item with id separator undefined
18:46:13.457 Uncaught (in promise) Error: Cannot find menu item with id hide-container undefined
18:46:13.457 Uncaught (in promise) Error: Cannot find menu item with id move-to-new-window-container undefined
```

### Expected behavior

No logspam.

### Additional informations

Search for `contextMenus.remove` at https://github.com/mozilla/multi-account-containers/blob/a60f5bb1be87e636241f22600902c9a70cee786a/src/js/background/assignManager.js . Two things stand out: these call `browser.contextMenus.remove(...)` without `await`, so any errors do not interrupt the execution (so it is just logspam). And (2) as the error shows, menus items are removed without even knowing whether the menus exist. A way to fix the logspam is to catch the error and silence it, e.g. `.catch(() => {})`.

[...]

PR #2760, opened 2025-04-29
Avoid logspam: "Cannot find menu item with id ..."

Rob--W:
Fixes https://github.com/mozilla/multi-account-containers/issues/2735


apostrophest (approved):
This fix makes sense to me based on the changes to `menus.remove` documented in Firefox 136. https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/menus/remove

dannycolin (approved):
Works as expected. LGTM

Issue #2827, opened 2025-11-18
We’re launching a survey about Multi-Account Containers 🚀


bakulf:
Hi everyone,

We’re running a survey to better understand how people use Multi-Account Containers and how we can make them more powerful and useful. This is our final study within our Containers-related research arc for this year. Depending on the rate of responses, the survey will be open for a few days to a few weeks, so that we can gather enough feedback from active users.

If you’re an active Containers user, we’d love for you to participate!

To help reach more people, the add-on will show a one-time view inside the browser action popup. This view will include:

* a short description,
* a link to the survey, and
* a button to return to the main popup view.

We hope this won’t be disruptive, but instead a good opportunity for you to tell us how you use Containers, what works well, and where we can improve.

👉 Take the survey [here](https://mozilla.qualtrics.com/jfe/form/SV_2aSQMGyfp2DFLtI)

Thanks a lot for helping us shape the future of Containers!

The Mozilla Team

PR #2828, opened 2025-11-18
Final survey. #2827

bakulf:
This is the implementation of the new survey described in #2827


dannycolin:
@bakulf If the user clicks on "back", they can never see the survey again. What if they simply didn't have the time to take the survey right away. Should we give them a "maybe later" option?

bakulf:
the "maybe later" proposal is interesting but it introduces a next level of complexity.  @lesleyjanenorton  thoughts?

lesleyjanenorton:
I'm inclined to agree with you @bakulf. Also- the initial survey wasn't up for long and it seems likely that someone could click 'maybe later' and then by the time 'later' rolls around the survey is no longer available.

PR #2833, opened 2025-12-02
Migrate to v3

bakulf:
This patch is a starting point to migrate this add-on to manifest v3.


bakulf:
Thank you Rob!

I did a bit of testing and the code does have almost 0 states. The only one I spot so far is related to the isolation key and the 3rd commit is about that bit.

> It is also worth noting that host permissions were not automatically granted on install in MV3 until Firefox 128. This is significant for ESR115 users: Existing users will retain the permissions, but new users will not have any host permissions.

This is what I'm mostly concern about: the backward compatibility of this change.

PR #2835, opened 2025-12-10
icons: Honor context-fill / context-fill-opacity, and dark mode as a fallback in page action icon

emilio:
The only reason this icon doesn't look terrible in dark mode is because there's a hack in Firefox that we want to remove, see bug 2001318.

# Description

The only reason this icon doesn't look terrible in dark mode is because there's a hack in Firefox that we want to remove, see bug 2001318.

This shouldn't make the behavior worse with and without the hack, and allows us to remove it.

## Type of change

*Select all that apply.*

- [x] Bug fix
- [ ] New feature


Rob--W:
For visibility, this patch followed from a report at https://bugzilla.mozilla.org/show_bug.cgi?id=2001318

PR #2841, opened 2026-01-06
Fix 1822: Sync container colors with Firefox container settings

### LINKED ISSUE #1822: No black color available (lost on editing) if editing via popup
- Multi-Account Containers Version: 7.0.2
- Operating System + Version: Windows 10 x64, 1909
- Firefox Version: 78.0.2
- Other installed Add-ons + Version + Enabled/Disabled-Status:
Bookmark Tab Here	1.0	true	{bc21b9a1-3ad6-4b00-bca4-ef5b7e21253b}
Copy Current Page URL with Anchor	1.0	true	jid1-grAPHQfoBoWMdA@jetpack
Copy Link Text	1.6.1	true	{b144be59-6bdc-41e0-9141-9f8d00373d93}
Copy PlainText	1.11	true	copyplaintext@eros.man
Download Manager (S3)	5.12	true	s3download@statusbar
Enhancer for YouTube™	2.0.101	true	enhancerforyoutube@maximerf.addons.mozilla.org
Facebook Container	2.1.1	true	@contain-facebook
Firefox Multi-Account Containers	7.0.2	true	@testpilot-containers
Gesturefy	3.0.5	true	{506e023c-7f2b-40a3-8066-bc5deb40aebe}
Refined GitHub	20.7.16	true	{a4c4eda4-fb84-4a84-b4a1-f7c1cbf2a1ad}
RoboForm Password Manager	9.0.1.6	true	rf-firefox@siber.com
Stylus	1.5.11	true	{7a7a4a92-a2a0-41d1-9fd7-1e92480d612d}
uBlock Origin	1.28.2	true	uBlock0@raymondhill.net
uMatrix	1.4.0	true	uMatrix@raymondhill.net
LastPass: Free Password Manager	4.51.0.1	false	support@lastpass.com
To Google Translate	4.0.6	false	jid1-93WyvpgvxzGATw@jetpack
Worldwide Radio	4.1	false	worldwide@radio
Xdebug-ext	1.2	false	{5483b263-bf97-4858-a2cf-590444554869}

### Actual behavior
There is no black colour when editing/viewing container settings via Multi-Account Containers popup panel. But it is there when managing container via Firefox preferences.

As a result the container will loose black colour and it will be reset to default/1st in the list (which is blue).

### Expected behavior
The black colour is present and remain assigned.

### Steps to reproduce
1. Multi-Account Containers popup panel
  image
2. Click on "Manage Containers" at the bottom
3. Locate desired container (e.g. Facebook) that has assigned black color to it
  image
4. Click on it to bring the editing panel page (e.g. you want to see what domains are assigned to this container)
  image
5. Do not edit anything (or edit, it does not matter) and click "<" button to go back
**Expected:** black colour remains assigned
**Actual:** black is gone, it is now a blue colour instead.
  image

To fix it I have to change the colour via Firefox Options: `about:preferences#containers`

image

nekno:
- [X] I agree to license my code under the [MPL 2.0 license](https://www.mozilla.org/en-US/MPL/2.0/).
- [X] I rebased my work on top of the main branch.
- [X] I ran `npm test` and all tests passed.
- [X] I added test coverages if relevant.

# Description

This PR fixes #1822, which captures the issue well. 

Firefox allows [9 container colors](https://github.com/mozilla-firefox/firefox/blob/e61d59b5c9a651fd7bf28043f87c0dc669833496/browser/components/preferences/dialogs/containers.js#L52), including a `toolbar` color that matches the current text color in light and dark modes.

The extension currently only supports 8 of those colors, omitting the `toolbar` color. That allows users to set a color in the Firefox container settings that doesn't appear or allow editing properly in the extension container settings.

This PR adds support for the `toolbar` color, to sync properly with the color names in the Firefox container settings.

I assume this issue has remained unresolved for ~6 years and marked `Needs: Mozilla Central` because, in lieu of hardcoding two identical collections across codebases, it would be best, architecturally, to expose the array of supported colors and icons on `ContextualIdentityService` in Firefox as the source of truth, and then consume those in the extension, so they would never be out-of-sync. There are also the separate requests to support more icons and any arbitrary RGB color.

Both of those changes would require a lot more support, so I figured it would be better to get this fixed first using the current approach, and those alternatives would be better taken on by a member of the project(s) rather than a random outside contributor.

If those alternatives are something you're looking for someone to take on, let me know.

| Edit Panel Light Theme | Edit Panel Dark Theme |
|--------|--------|
| Edit Panel Light Theme | Edit Panel Dark Theme | 

## Type of change

*Select all that apply.*

- [X] Bug fix
- [ ] New feature

* #1822

### DIFF
--- a/src/css/popup.css
+++ b/src/css/popup.css
@@ -69,7 +69,7 @@
   --icon-button-size: calc(calc(var(--block-line-separation-size) * 2) + 1.66rem); /* 20px */
   --inactive-opacity: 0.3;
   --overflow-size: 1px;
-  --icon-fit: 8;
+  --icon-fit: 9;
 
   background: var(--bgColor);
   margin-block: 0;
@@ -390,6 +390,11 @@ table {
 [data-identity-color="purple"] {
   --identity-tab-color: #af51f5;
   --identity-icon-color: #af51f5;
+}
+
+[data-identity-color="toolbar"] {
+  --identity-tab-color: currentColor;
+  --identity-icon-color: currentColor;
 }
 
 [data-identity-icon="fingerprint"] {


src/css/popup.css
```diff
@@ -69,7 +69,7 @@
   --icon-button-size: calc(calc(var(--block-line-separation-size) * 2) + 1.66rem); /* 20px */
   --inactive-opacity: 0.3;
   --overflow-size: 1px;
-  --icon-fit: 8;
+  --icon-fit: 9;
```
nekno: This seems to just be hardcoded to match the number of colors. Each color swatch is 32 x 32px, and the panel is 320px wide, so you can fit 10 colors in a row without compromising on minimum spacing/padding on anything.

This changes it from 8 to 9 so all colors fit on the same row, and the minimum spacing is still met.

src/js/popup.js
```diff
@@ -1889,7 +1889,7 @@ Logic.registerPanel(P_CONTAINER_EDIT, {
       return Utils.escaped`<input type="radio" value="${containerColor}" name="container-color" id="edit-container-panel-choose-color-${containerColor}" />
      <label for="edit-container-panel-choose-color-${containerColor}" class="usercontext-icon choose-color-icon" data-identity-icon="circle" data-identity-color="${containerColor}">`;
     };
-    const colors = ["blue", "turquoise", "green", "yellow", "orange", "red", "pink", "purple"];
+    const colors = ["blue", "turquoise", "green", "yellow", "orange", "red", "pink", "purple", "toolbar"];
```
nekno: This matches the `toolbar` color named in the [Firefox container settings dialog](https://github.com/mozilla-firefox/firefox/blob/e61d59b5c9a651fd7bf28043f87c0dc669833496/browser/components/preferences/dialogs/containers.js#L52).

achernyakevich-sc:
@nekno As I could understand from you commit you just added support of `toolbar` color...

But maybe it is not a bug but it is the special intention of developers to give no chance to mislead users by providing possibility to select `toolbar` color so that on UI it will be undetectable that you have opened tab in container instead of default (no container)?

Be sure as anybody will select `toolbar` color for some container then it is just a matter of time that it will report issues like:

* something does not work as expected in default container - they just will not distinguish default and their `toolbar`'s containers.
* container identification image is not visible for the menu like below - gray on gray really could be invisible.
image


I'm really not sure. Maybe we need ask somebody like @dannycolin.

nekno:
@achernyakevich-sc — I don't know the history why the color is named `toolbar`, but there isn't an option to change it, as the extension must match the implementation used in the Firefox settings, where the name `toolbar` is used.

The color named `toolbar` isn't actually the background color of the toolbar, it's the primary text color of your current theme, and it matches the color of the line on top of the tab, which is controlled by code in Firefox.

I added screenshots to the description to show it rendered in context.

For the light theme it's #15141A: 
15141A

For the dark theme it's #FBFBFE:
FBFBFE

achernyakevich-sc:
Then it looks reasonable. Thank you for your efforts.

dannycolin:
Sorry for the late response. 

There was some issues/limitations with the *toolbar* color that prevented it to be added in the list. See a previous PR: https://github.com/mozilla/multi-account-containers/pull/2192

nekno:
Thanks, @dannycolin, I didn't find that previous PR in my searching. I wish I would have!

It's funny that we came up with almost identical PRs independently, though I used `currentColor` to match Firefox, rather than `--text-color-primary` in the other PR.

In testing, I see your point still stands, that `currentColor` doesn't work with themes other than light/dark, either.

None of the colors in MAC are expressly themable, so despite the `toolbar` color responding specifically to the light/dark themes, it doesn't to be a requirement for a color to respond to a theme in order to be used in MAC. This missing color is, in fact, the only color that responds to the light/dark themes for contrast accessibility, and it's better than not that it at least does that much.

While I would love to follow your suggestion in the other PR and fix this upstream, the effort and knowledge required to add a custom color picker is much greater than what's required to fix this issue.

If you have any references on how to pull in an existing color picker and incorporate it into the experience, I'm willing to try, but I'm likely not the best person. I will be learning most of it from scratch and the juice is not likely worth the squeeze.

I ask honestly — must we suffer this issue another 5 years until someone with the requisite skills and knowledge finds the time?

dannycolin:
> the only color that responds to the light/dark themes for contrast accessibility

This isn't necessarily true because we can't control third-party themes.

> If you have any references on how to pull in an existing color picker and incorporate it into the experience, I'm willing to try, but I'm likely not the best person. I will be learning most of it from scratch and the juice is not likely worth the squeeze.

This would be the best longterm solution but there are a lot of moving pieces. The mechanism needs to be implemented in Firefox itself. Plus, this change would have an impact on the webextension API so we could potentially break other people addons if their code expect hardcoded color values.

I'm not a big fan of adding more hardcoded colors but I can still ask folks at Mozilla if we couldn't find a short term solution. For example, we could align our color values with the new Tab Group feature which offers 9 predefined colors. It always offers variant of the color palette depending if the theme is set as a dark or light theme. In other words, this solution would be similar to your fix for the end-user but we'd be in control of the visual rendering which would avoid the pitfalls of using the *toolbar* color.

I'll keep you posted.

PR #2854, opened 2026-02-11
RFC for Sync

bakulf:
This PR contains a RFC for Containers + Sync. Please send feedback. The goal is to have a better, secure, safe sync support for containers.


jonathanKingston:
@bakulf seems to make sense to me from a quick glance. I do wonder if all extensions would benefit from this kind of opt-in API though? Something for: https://github.com/w3c/webextensions ?

PR #2868, opened 2026-03-07
Fix #2699: Prevent trashcan from being obscured by scrollbar

### LINKED ISSUE #2699: Trashcan partly obscured by scrollbar when managing site list of a container with "Always show scrollbars" option

### Step to reproduce

In Firefox Multi-Account Containers press Manage Containers, select a container with sufficient sites to require a vertical scrollbar, and then press Manage Site List…

### Actual behavior

When hovering over a site the trashcan icon is partly obscured by the scrollbar

### Expected behavior

Either the trashcan icon or the scrollbar should be moved so that the trashcan remains unobscured

### Additional informations

This was reported in https://bugzilla.mozilla.org/show_bug.cgi?id=1931273 .

[This inline-size style for .truncate-text](https://github.com/mozilla/multi-account-containers/blob/main/src/css/popup.css#L316) is a bit longer with showing the scroll bar.

_No response_

LuciferDono:

- [x] I agree to license my code under the MPL 2.0 license.

# Description

When users have "Always show scrollbars" enabled in their OS settings, the trashcan (delete) and reset buttons on assigned site list rows are partly hidden behind the visible scrollbar.

**Root cause:** `.truncate-text` used `inline-size: calc(100vw - 100px)`, which bases the width on the full viewport — including the area consumed by the scrollbar. This forced the text span to a fixed width that overflows the flex container when a scrollbar is present, pushing the action buttons behind it.

**Fix:** Replace the fixed `calc()` width with `min-inline-size: 0`, which is the standard CSS fix for allowing flex items to shrink below their content size. The parent `<td>` already uses `display: flex` and `.menu-text` already has `flex: 1`, so the text span now naturally fills only the remaining space after the favicon and buttons — regardless of whether a scrollbar is visible.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Major change

Fixes #2699

### DIFF
--- a/src/css/popup.css
+++ b/src/css/popup.css
@@ -309,9 +309,9 @@ table {
   transition: opacity 0s;
 }
 
-/* effect borrowed from tabs in firefox, ensure that the element flexes to the full width */
+/* Allow flex item to shrink below content size for proper text truncation */
 .truncate-text {
-  inline-size: calc(100vw - 100px);
+  min-inline-size: 0;
   overflow: hidden;
   position: relative;
   white-space: nowrap;


bakulf:
Can you please attach a screenshot before and after? Thanks!

PR #2869, opened 2026-03-09
Fix package script

bakulf:
This should be the last commit for 8.3.7


bakulf:
Yes. It's a meta commit... :)

PR #2882, opened 2026-04-07
Add a URL validation for the confirmation page

bakulf:
A validation of the URL was missing.

### DIFF
--- a/src/js/confirm-page.js
+++ b/src/js/confirm-page.js
@@ -1,7 +1,22 @@
+function isSafeUrl(url) {
+  try {
+    const { protocol } = new URL(url);
+    return protocol === "http:" || protocol === "https:";
+  } catch {
+    return false;
+  }
+}
+
 async function load() {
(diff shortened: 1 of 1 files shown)


dannycolin:
Why are we validating the URL on the confirmation page instead of on assignment? AFAIK, the confirmation page is only visible for URLs that have already been assigned so they should already be valid.

If we aren't already validating them on assignment, would it make more sense to do it in   [_setOrRemoveAssignment](https://github.com/mozilla/multi-account-containers/blob/df43ffbf2e08eb4eee3656e2e4da86f3b9201a07/src/js/background/assignManager.js#L590)?

bakulf:
> Why are we validating the URL on the confirmation page instead of on assignment? AFAIK, the confirmation page is only visible for URLs that have already been assigned so they should already be valid.

This is a fix for a claude report. If an attacker manages to find the MAC UUID, they can force non HTTP URLs

lesleyjanenorton (approved):
👍

PR #2903, opened 2026-06-04
Colors and icons from the browser

bakulf:
In firefox 153, ContextualIdentity webext API will (hopefully) export colors and icons. That allows the addon to be fully in sync with what the browser supports.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -18,6 +18,7 @@ module.exports = defineConfig([{
       ...globals.node,
       ...globals.webextensions,
       "Utils": true,
+      "ContainerStyle": true,
       "CustomizableUI": true,
       "CustomizableWidgets": true,
       "SessionStore": true,
--- a/src/css/popup.css
+++ b/src/css/popup.css
@@ -340,107 +340,12 @@ table {
 }
 
 /* Color and icon helpers */
-[data-identity-color="blue"] {
-  --identity-tab-color: #37adff;
-  --identity-icon-color: #37adff;
-}
-
-[data-identity-color="turquoise"] {
-  --identity-tab-color: #00c79a;
-  --identity-icon-color: #00c79a;
-}
-
-[data-identity-color="green"] {
-  --identity-tab-color: #51cd00;
-  --identity-icon-color: #51cd00;
-}
-
-/* We disable no-descending-specificity so we can keep [data-theme="dark"]
- * rules grouped together at the beginning of the file
- */
 /* stylelint-disable no-descending-specificity */
 [data-identity-color="grey"] {
   /* Only used for the edit panel */
   --identity-icon-color: #616161;
 }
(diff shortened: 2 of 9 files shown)


Rob--W:
FYI this depends on an API introduced in https://bugzilla.mozilla.org/show_bug.cgi?id=2044712

dannycolin:
It works on both current Nightly and the autoland branch with the new webextension API. the only scenario I didn't test is using a new color and downgrading Firefox to a version that doesn't have the new API.

dannycolin (approved):

dannycolin (changes requested):
In _Manage Containers_ > [Container Name], the _delete_ button is rendered slightly offcanvas.

image

Alos, Container data is lost/reset to default when opening the same profile in a previous version of Firefox. Is it expected?

bakulf:
> Alos, Container data is lost/reset to default when opening the same profile in a previous version of Firefox. Is it expected?

Yes. There is nothing we can do here, because we should patch previous versions of firefox to ignore unknown colors/icons. This cannot be done. Plus, firefox downgrade is not supported in general.

bakulf:
In the last update, the colors are aligned in 1 single line.

dannycolin:
> > Alos, Container data is lost/reset to default when opening the same profile in a previous version of Firefox. Is it expected?
> 
> Yes. There is nothing we can do here, because we should patch previous versions of firefox to ignore unknown colors/icons. This cannot be done. Plus, firefox downgrade is not supported in general.

This can be an issue even without downgrading if the user has Sync enabled. For example, they can use Firefox ESR on one machine and regular Firefox on another one. Knowing that ESR 153 won't have the new container colors, this is definitely a situation that has a high potential to happen. 

I'm not saying we shouldn't push the new colors but maybe there's something we can do to warn the user. Any idea?

bakulf:
Right. Sync is going to be an issue. About informing the user, yes, this is part of the plan. We must inform the user about the long-term project and inform them that we will slowly migrate functionalities from MAC into firefox. But that is a separate patch/issue.

dannycolin:
> We must inform the user about the long-term project and inform them that we will slowly migrate functionalities from MAC into firefox. But that is a separate patch/issue.

Taking that into consideration, the current patch works as expected now. 

image

dannycolin (approved):

bakulf:
I'll work on Sync in a follow-up

TheoDubus97:
Hi all, I noticed that white was removed as a color option in the latest MAC release. Not a big issue, but is there a reason for this? And are there plans to add more colors in the future?

bakulf:
Since Firefox 153, the lists of colors and icons are retrieved from the browser. So yes! Now we can add/remove/modify icons and colors and MAC will update itself automatically.

PR #2352, opened 2022-05-04
Wildcard subdomains v2 - e.g. *.google.com

mckenfra:
See pull request https://github.com/mozilla/multi-account-containers/pull/1500 and issue https://github.com/mozilla/multi-account-containers/issues/473

This is a second attempt at implementing a wildcard subdomains feature. This time round, I have tried to keep the amount of code that is changing to the bare minimum. This should make code review easier and minimise any risk of introducing regressions. We can potentially cancel the previous pull request if this one looks better.

https://user-images.githubusercontent.com/439593/166697696-78b3d20d-4032-41ac-b030-e017d36a28cb.mov

### How the code works

When you click the `www` part of `www.google.com` in the popup, a `wildcardHostname` property with value `google.com` is added to the stored `siteSettings` object for `www.google.com` in `assignManager.js` [See code here](https://github.com/mckenfra/multi-account-containers/blob/ab400dffdd62e05380b48d5985bd8ce356de5264/src/js/background/assignManager.js#L694)

If you then open a tab and navigate to `mail.google.com`, we need the code to find the matching `google.com` wildcard hostname stored in the `www.google.com` site. This is achieved by storing an additional map of `wildcardHostname` values to `siteStoreKeys`. [See code here](https://github.com/mckenfra/multi-account-containers/blob/ab400dffdd62e05380b48d5985bd8ce356de5264/src/js/background/assignManager.js#L135)
 
So when handling the request for `mail.google.com`, the wildcard hostname map is queried to see if there is a mapping for`mail.google.com`, then `google.com` (and then theoretically `com`). It finds that `google.com` is mapped to `www.google.com`, and so uses the `www.google.com` site settings to handle the request. [See code here](https://github.com/mckenfra/multi-account-containers/blob/ab400dffdd62e05380b48d5985bd8ce356de5264/src/js/background/assignManager.js#L289)

# Syncing

The sync feature should _just work_ as expected, since the new `wildcardHostname` property of the `siteSettings` will be synced along with all the other properties of the `siteSettings` object.

# Try it out

I have put up a temporary build [here](https://github.com/mckenfra/multi-account-containers/releases/tag/8.0.7.1-WildcardSubdomains) for you to try out the feature without having to build it yourself. Let me know if it looks good or if you find any issues.

I have also include unit tests in this pull request.

Issue #473, opened 2017-04-29
“Always Open in This Container” for TLD?


ArchangeGabriel:
Hi there,

I’m new to containers and enjoying the feature, especially with this “Always Open in This Container” feature.

However, I was wondering if it would be feasible to add website in this category by TLD, e.g. *mozilla.org instead of www.mozilla.org, bugzilla.mozilla.org, wiki.mozilla.org…

They are TLD for which I would want that (archlinux.org is another example), and some for which I don’t (likely google.com).

Thanks!

anewuser:
You don't know what a top-level domain is. The TLD of your examples is **org**.

ArchangeGabriel:
Hum yes sorry, you’re right. Edited the title.

groovecoder:
Container isolation is based on origins as [defined for Same Origin Policy](https://developer.mozilla.org/docs/Web/Security/Same-origin_policy). So, we would need to seriously consider the web privacy & security effects if we made it (too) easy to assign (wildcard) subdomains to a container.

jonathanKingston:
For clarity SOP considers port(EG: 80, 300, etc) and protocol(EG: http, https). We however don't.

As mentioned, there are some you won't want this feature for. Deciding on the right UI for this seems a little tricky.

b0urb0n:
I think this would be particularly useful for people who want to ensure that information set by domain-wide resources (behind their company's VPN) don't get disclosed to any other website.  My particular use case would be to have anything that matches *.myworkdomain.com auto-open in the "Work" container.  This would be easier than manually adding each host/URL.

Perhaps a disclaimer: "Beware: this feature, if mis-configured, could defeat the purpose all together."  Or determining the minimum that one can wildcard as to not defeat the point of this feature:

Test | Pass/Fail
---|---
`*`|:x:
`*.*`|:x:
`*.tld`|:x:
`*.domain.tld`|:white_check_mark:
`*.sub.domain.tld`|:white_check_mark:
...|:white_check_mark:

jonathanKingston:
@b0urb0n the problem also is that users want to restrict their search engine traffic from their mail provider.

They for example might want mail.google.com and docs.google.com into a work continer but not www.google.com.

This leads to the conflicting use cases as you mention. I'm not sure if there is an obvious resolution.

b0urb0n:
[...]

Issue #2901, opened 2026-05-22
[linux] container lost all settings on restarting firefox


strider16:
### Before submitting a bug report

- [x] I updated to the latest version of Multi-Account Container and tested if I can reproduce the issue
- [x] I searched for existing reports to see if it hasn't already been reported

### Step to reproduce

User Agent: Mozilla/5.0 (X11; Linux x86_64; rv:150.0) Gecko/20100101 Firefox/150.0

Steps to reproduce:

1 - configure firefox account and firefox sync (NOT the sync option inside m-a-c)
2 - set all containers colors, icons, and credentials while browsing
3 - close the browser
4 - restart the browser reloading all tabs from previous session

### Actual behavior

Actual results:

My containers are lost. Default container reappear.

All tabs are remembered, and even the container color on the tabs that uses them appear for a while. Time enough to type in the password anyway, because in a blink of eye the tabs lose their colored borders and when I click to great a new tab, it reverted to the default 4 types.

### Expected behavior

Expected results:

Expected result: all tabs with respective containers and credentials should have kept their settings.

### Additional informations

tldr: two pcs, one nightly/windows, one firefox/mint, firefox account sync'd, but only F/M loses its container settings very frequently, N/W is rock stable.

I use firefox in 3 instances, 2 windows and 1 linux. Actually, I first started using 2 nightly instances, both windows devices, but then I turned one of them to a linux mint, and since firefox came preinstalled, I kept it instead installing nightly, so one windows instance is on "halt" since some month. Sync works fine, AFAIK firefox sync doesn't sync containers, and this is cool, I use a handful of containers and creating them at a first setup isn't a big deal.

After some weeks, from time to time my firefox on mint started losing its container settings. It wasn't frequent, but it started to accelerate, once a month, than weekly, then twice a week. Now happens each other day. I close firefox and go to work, when I come home and start firefox, it isn't unlikely it will lose the containers I create and the tabs revert to a non-container state. It can happens if I restart the PC too. It is guaranteed to happen if firefox is open while the pc crashes (gaming).

[...]

Issue #2682, opened 2024-10-23
Deleted Container Reappears On Devices Even after Syncing After Deletion


levski:
### Before submitting a bug report

- [X] I updated to the latest version of Multi-Account Container and tested if I can reproduce the issue
- [X] I searched for existing reports to see if it hasn't already been reported

### Step to reproduce

open tab for container 1, open website that was once set to open for container 2 only (in my case it was outlook.com)
it will then ask you which container, you should open the website in (which shouldn't appear since the other container was deleted in the past)

### Actual behavior

randomly, the extension will sync even deleted containers, when it does this issue will happen.

### Expected behavior

syncing extension should sync to the master list and sync to all devices.

### Additional informations

not sure what else to say here.

StormPooper:
I've also had deleted containers reappearing recently.

newptcai:
Yes. Deleting container does not seem to work.

wondercomm:
Same issue here. Is there a fix for this yet?

krc:
Possibly related to or same as #2717 (newer), unclear if that relates to syncing. In my case at least I do have syncing turned on. Would be interesting to hear of a case with no sync.

alexnt85:
> Possibly related to or same as [#2717](https://github.com/mozilla/multi-account-containers/issues/2717) (newer), unclear if that relates to syncing. In my case at least I do have syncing turned on. Would be interesting to hear of a case with no sync.

I recently migrated my firefox profile to waterfox and I'm facing this issue. There is a specific container called "private" that reappears and I don't have sync enabled.

hovancik:
having same issue

wondercomm:
> I had a collection of bookmarks in a Container. I don't need the Container anymore. So I deleted the Container (i.e. Remove). I sync all devices and the Container is gone. A few days later, the Container that I deleted is back. I remove the Container, and sync. Check other devices and all OK. Then a few days later, the Container is back. It feels like Groundhog Day ...
> 
> Maybe I have to manually delete all the bookmarks in that Container???

I simply renamed the container(s) that kept coming back and used it for other purposes instead of creating new container. This change was persistent and for now this is a workaround until Mozilla fixes this.

hovancik:
So syncing does work for you? I have multiple devices but containers are not synced as I would expect.

jacobthemagicalpanda:
[...]

PR #2912, opened 2026-06-22
Feature: catchall domains - e.g. *.google.com

mckenfra:

# Description

Adds the ability to assign all subdomains of a domain to a container.

The list of domains is automatically updated as new sites are assigned to a container. This feature is made possible by the new [`publicSuffix`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/publicSuffix) API that is due to be released in [Firefox 153](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/153).

https://github.com/user-attachments/assets/5e0a9933-5c3d-4fd0-921b-9604f654caf9

## Type of change

*Select all that apply.*

- [ ] Bug fix
- [x] New feature
- [x] Major change (fix or feature that would cause existing functionality to work differently than in the current version)

* #473 
* #2352

### DIFF
--- a/src/js/background/assignManager.js
+++ b/src/js/background/assignManager.js
@@ -46,6 +46,18 @@ window.assignManager = {
       return this.getByUrlKey(siteStoreKey);
     },
 
+    async match(pageUrl) {
+      const domainMatchKeys = domainManager.getDomainMatchKeys(pageUrl);
+      const siteStoreKey = this.getSiteStoreKey(pageUrl);
+      const storageResponse = await this.area.get([...domainMatchKeys, siteStoreKey]);
+      const domainStoreKey = domainMatchKeys.find(key => key in storageResponse);
+      const site = storageResponse[domainStoreKey ?? siteStoreKey];
+      if (domainStoreKey) {
+        site.isDomain = true;
+      }
+      return [domainStoreKey ?? siteStoreKey, site];
+    },
+
     async getSyncEnabled() {
       const { syncEnabled } = await browser.storage.local.get("syncEnabled");
       return !!syncEnabled;
@@ -118,9 +130,15 @@ window.assignManager = {
             continue;
           }
           const site = siteConfigs[urlKey];
-          // In hindsight we should have stored this
-          // TODO file a follow up to clean the storage onLoad
-          site.hostname = urlKey.replace(/^siteContainerMap@@_/, "");
+          const domainName = domainManager.getDomainNameFromStoreKey(urlKey);
+          if (domainName) {
+            site.hostname = domainName;
+            site.isDomain = true;
+          } else {
+            // In hindsight we should have stored this
+            // TODO file a follow up to clean the storage onLoad
+            site.hostname = urlKey.replace(/^siteContainerMap@@_/, "");
+          }
           sites[urlKey] = site;
         }
       }
@@ -172,11 +190,11 @@ window.assignManager = {
 
       // If we have existing data and for some reason it hasn't been
       // deleted etc lets update it
-      this.storageArea.get(pageUrl).then((siteSettings) => {
+      this.storageArea.match(pageUrl).then(([siteStoreKey, siteSettings]) => {
         if (siteSettings) {
           siteSettings.neverAsk = true;
           siteSettings.userContextId = backgroundLogic.getUserContextIdFromCookieStoreId(m.cookieStoreId);
-          this.storageArea.set(pageUrl, siteSettings);
+          this.storageArea.set(siteStoreKey, siteSettings);
         }
       }).catch((e) => {
         throw e;
@@ -186,8 +204,8 @@ window.assignManager = {
 
   // We return here so the confirm page can load the tab when exempted
   async _exemptTab(m) {
-    const pageUrl = m.pageUrl;
-    await this.storageArea.setExempted(pageUrl, m.tabId);
+    const [siteStoreKey] = await this.storageArea.match(m.pageUrl);
+    await this.storageArea.setExempted(siteStoreKey, m.tabId);
     return true;
   },
 
@@ -224,9 +242,9 @@ window.assignManager = {
       return {};
     }
     this.removeContextMenu();
-    const [tab, siteSettings] = await Promise.all([
+    const [tab, [siteStoreKey, siteSettings]] = await Promise.all([
       browser.tabs.get(options.tabId),
-      this.storageArea.get(options.url)
+      this.storageArea.match(options.url)
     ]);
     let container;
     try {
@@ -238,7 +256,7 @@ window.assignManager = {
 
     // The container we have in the assignment map isn't present any
     // more so lets remove it then continue the existing load
-    if (siteSettings && !container) {
+    if (siteSettings && !siteSettings.isDomain && !container) {
       this.deleteContainer(siteSettings.userContextId);
       return {};
     }
@@ -267,7 +285,7 @@ window.assignManager = {
     if (!siteIsolatedReloadInDefault) {
       if (!siteSettings
           || userContextId === siteSettings.userContextId
-          || this.storageArea.isExempted(options.url, tab.id)) {
+          || this.storageArea.isExempted(siteStoreKey, tab.id)) {
         return {};
       }
     }
@@ -585,6 +603,16 @@ window.assignManager = {
   },
 
   async _setOrRemoveAssignment(tabId, pageUrl, userContextId, remove) {
+    const assignmentStoreKey = this.storageArea.getSiteStoreKey(pageUrl);
+    await this.setOrRemoveAssignmentOrDomain(tabId, assignmentStoreKey, userContextId, remove, false);
+  },
+
+  async _setOrRemoveDomain(domainName, userContextId, remove) {
+    const domainStoreKey = domainManager.getDomainStoreKeyFromName(domainName);
+    await this.setOrRemoveAssignmentOrDomain(false, domainStoreKey, userContextId, remove, true);
+  },
+
+  async setOrRemoveAssignmentOrDomain(tabId, storeKey, userContextId, remove, isDomain) {
     let actionName;
     // https://github.com/mozilla/testpilot-containers/issues/626
     // Context menu has stored context IDs as strings, so we need to coerce
@@ -593,11 +621,17 @@ window.assignManager = {
 
     if (!remove) {
       const tabs = await browser.tabs.query({});
-      const assignmentStoreKey = this.storageArea.getSiteStoreKey(pageUrl);
       const exemptedTabIds = tabs.filter((tab) => {
-        const tabStoreKey = this.storageArea.getSiteStoreKey(tab.url);
+        let isSameHostname;
+        if (isDomain) {
+          const tabDomainMatchKeys = domainManager.getDomainMatchKeys(tab.url);
+          isSameHostname = !!tabDomainMatchKeys.find(key => key === storeKey);
+        } else {
+          const tabStoreKey = this.storageArea.getSiteStoreKey(tab.url);
+          isSameHostname = tabStoreKey === storeKey;
+        }
         /* Auto exempt all tabs that exist for this hostname that are not in the same container */
-        if (tabStoreKey === assignmentStoreKey &&
+        if (isSameHostname &&
             this.getUserContextIdFromCookieStore(tab) !== userContextId) {
           return true;
         }
@@ -606,14 +640,14 @@ window.assignManager = {
         return tab.id;
       });
 
-      await this.storageArea.set(pageUrl, {
+      await this.storageArea.set(storeKey, {
         userContextId,
         neverAsk: false
       }, exemptedTabIds);
       actionName = "assigned site to always open in this container";
     } else {
       // Remove assignment
-      await this.storageArea.remove(pageUrl);
+      await this.storageArea.remove(storeKey);
 
       actionName = "removed from assigned sites list";
 
@@ -651,13 +685,15 @@ window.assignManager = {
     // Ensure we have a cookieStore to assign to
     if (cookieStore
         && this.isTabPermittedAssign(tab)) {
-      return this.storageArea.get(tab.url);
+      const [, siteSettings] = await this.storageArea.match(tab.url);
+      return siteSettings;
     }
     return false;
   },
 
-  _getByContainer(userContextId) {
-    return this.storageArea.getAssignedSites(userContextId);
+  async _getByContainer(userContextId) {
+    const sites = await this.storageArea.getAssignedSites(userContextId);
+    return domainManager.getDomainsAndAssignments(sites);
   },
 
   removeContextMenu() {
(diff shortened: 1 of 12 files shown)


Ignotum77:
Does the code have I fallback when `publicSuffix` is not supported for older FF versions? Because with the current code it will only work on FireFox 153 or above.

Furthermore, thank for this updated version. It solves the problem discussed by @dannycolin in https://github.com/mozilla/multi-account-containers/pull/2352#issuecomment-1137466167.

mckenfra:
> Does the code have I fallback when `publicSuffix` is not supported for older FF versions? Because with the current code it will only work of FireFox 153 or above.

No the current implementation does not provide a fallback. Instead, the minimum required Firefox version is increased to `153.0.0`.

See the change to the min version here: https://github.com/mozilla/multi-account-containers/pull/2912/changes#diff-6bc2c0b5164076a1b57b067398be19a40d2b8efa3428b03504562ea88593866cR37

Rob--W:
Instead of bumping the minimum required version to 153, would it be feasible to hide the UI and feature when the feature is unavailable? That enables users of older Firefox versions to use other features of MAC.

In particular, Firefox ESR 115 continues to be supported for a while for older Windows versions (until at least the start of the next year according to https://whattrainisitnow.com/calendar/).

Firefox ESR 140 is also still supported, albeit only for a few months more. Fortunately the latest ESR is Firefox ESR 153, which is not too coincidentally the release that includes publicSuffix support.

mckenfra:
Makes sense, I will look at updating this PR to disable the catchall domains functionality on earlier Firefox versions.

Ignotum77:
> In particular, Firefox ESR 115 continues to be supported for a while for older Windows versions (until at least the start of the next year according to https://whattrainisitnow.com/calendar/).

However if we wait until March 2027 a "disable the catchall domains functionality on earlier Firefox versions" isn't needed.
So question is, are we willing to wait for this feature?

mckenfra:
I have updated this PR as follows:

1. Change the minimum required Firefox version to `115.0.0` instead of `153.0.0`.
2. Hide the catchall-domain UI and functionality if`browser.publicSuffix` is unavailable.

Rob--W:
> I have updated this PR as follows:
> Change the minimum required Firefox version to `115.0.0` instead of `153.0.0`.

Should be `115.0`, there is no `115.0.0`.

FYI: Although the patch "bumps" the version from 91.1.0 to 115.0, there is effectively no change in supported versions, because the addition of `data_collection_permissions` in a previous commit already made the extension incompatible with versions before 115 due to https://bugzilla.mozilla.org/show_bug.cgi?id=1996829 . And even if it were not incompatible, Firefox versions before 115 cannot receive extension updates anyway due to https://support.mozilla.org/en-US/kb/root-certificate-expiration (both of these issues actually affect Firefox 127 and earlier, but Firefox ESR 115 has received patched to fix the issues, and there is no way to say "Firefox 128+ and Firefox ESR 115", so setting the min version to `115.0` is the closest alternative).

src/js/background/domainManager.js
```diff
@@ -0,0 +1,110 @@
...
+  isEnabled() {
+    return "publicSuffix" in browser;
+  },
+
+  getDomainNameFromHostname(hostname) {
+    if (!this.isEnabled()) { return null; }
+    try {
+      const domainName = browser.publicSuffix.getDomain(
+        hostname,
+        { allowUnknownSuffix: true, encoding: "display" },
```
Rob--W, line 11: "display" should be used for display only, not as a storage key. Otherwise changes to display rules (by browser updates) would change the storage key and cause associations to be lost.

src/js/background/assignManager.js
```diff
@@ -46,6 +46,18 @@ window.assignManager = {
       return this.getByUrlKey(siteStoreKey);
     },
 
+    async match(pageUrl) {
+      const domainMatchKeys = domainManager.getDomainMatchKeys(pageUrl);
+      const siteStoreKey = this.getSiteStoreKey(pageUrl);
+      const storageResponse = await this.area.get([...domainMatchKeys, siteStoreKey]);
+      const domainStoreKey = domainMatchKeys.find(key => key in storageResponse);
```
Rob--W, line 53: Shouldn't you try to select the longest matching domain instead of an arbitrary one? Especially with the specified lookup order on the previous line (domainMatchKeys before siteStoreKeys), the implementation would end up preferring catchall domains over specific domains.

src/js/background/assignManager.js
```diff
@@ -172,11 +190,11 @@ window.assignManager = {
 
       // If we have existing data and for some reason it hasn't been
       // deleted etc lets update it
-      this.storageArea.get(pageUrl).then((siteSettings) => {
+      this.storageArea.match(pageUrl).then(([siteStoreKey, siteSettings]) => {
         if (siteSettings) {
           siteSettings.neverAsk = true;
```
Rob--W, line 195: Not sure if I'm reading the code right, but could it be that the UI is prompting for a specific site, and then ends up saving the decision for the catchall? That would be surprising if it is the case.

src/js/background/assignManager.js
Rob--W, line 187: Does this need to account for catchalls as well?

src/js/background/domainManager.js
```diff
@@ -0,0 +1,110 @@
...
+    const labels = domainName.split(".");
+    let suffix = labels.pop();
+    return labels
+      .reverse()
+      .map(label      => suffix = `${label}.${suffix}`)
+      .map(domainName => this.getDomainStoreKeyFromName(domainName))
+      .reverse();
+  },
+
+  getDomainNameFromStoreKey(domainStoreKey) {
```
Rob--W, line 56: I'd expect the  `getDomainNameFromStoreKey` implementation to be completely neutral if `publicSuffix` is unavailable, but there is no `isEnabled` check here. Shouldn't we add an `isEnabled` check here?

src/js/background/assignManager.js
```diff
@@ -238,7 +256,7 @@ window.assignManager = {
 
     // The container we have in the assignment map isn't present any
     // more so lets remove it then continue the existing load
-    if (siteSettings && !container) {
+    if (siteSettings && !siteSettings.isDomain && !container) {
```
Rob--W, line 259: Why is the "clean up when container isn't present anymore" logic conditional on `!siteSettings.isDomain`? With the proposed logic you'd fall through below, to `reloadPageInContainer` with an invalid `cookieStoreId`.

src/js/background/domainManager.js
```diff
@@ -0,0 +1,110 @@
...
+    } catch {
+      return null;
+    }
+  },
+
+  getDomainStoreKey(pageUrlorUrlKey) {
+    if (pageUrlorUrlKey.includes("siteContainerMap@@_")) return pageUrlorUrlKey;
+    const url = new window.URL(pageUrlorUrlKey);
+    const domainName = this.getDomainNameFromHostname(url.hostname);
+    return domainName ? this.getDomainStoreKeyFromHostname(domainName) : null;
```
Rob--W, line 25: `getDomainStoreKeyFromHostname` is not defined anywhere. In fact... this whole function `getDomainStoreKey` does not have any callers, so it is dead code.

src/js/background/domainManager.js
```diff
@@ -0,0 +1,110 @@
...
+  getDomainMatchKeysFromName(domainName) {
+    // E.g. "mysite.s3.us-west-1.amazonaws.com" ===>
+    // [
+    //   "siteContainerMap@@_*.mysite.s3.us-west-1.amazonaws.com",
+    //   "siteContainerMap@@_*.s3.us-west-1.amazonaws.com",
+    //   "siteContainerMap@@_*.us-west-1.amazonaws.com",
+    //   "siteContainerMap@@_*.amazonaws.com",
+    // ]
+    const labels = domainName.split(".");
+    let suffix = labels.pop();
```
Rob--W, line 48: This logic here appears to return keys that are not valid domain names. There are public suffices with more than 3 parts, e.g. `act.edu.au` and `edu.au` and `au` are all public suffixes. Is it OK to return such domains here? (I haven't checked other places, would be good to double check as well).

src/manifest.json
```diff
@@ -33,7 +34,7 @@
   "browser_specific_settings": {
     "gecko": {
       "id": "@testpilot-containers",
-      "strict_min_version": "91.1.0",
+      "strict_min_version": "115.0.0",
```
Rob--W, line 37: As noted at https://github.com/mozilla/multi-account-containers/pull/2912#issuecomment-5117716173 this should be 115.0, not 115.0.0

```suggestion
      "strict_min_version": "115.0",
```

Rob--W (commented):
I glanced over the code and left some comments. I haven't worked extensively on the code base before, so I'll defer to @bakulf for an in depth review.

src/js/background/domainManager.js
mckenfra: Agreed. This will require converting the stored domains from `punycode` to `display` when showing the domains in the UI. Since there is currently no API available for doing this type of encoding conversion, we will need to call `publicSuffix.getDomain(domain, { encoding: "display" })` again on each stored domain. See a relevant discussion on this topic here: https://github.com/w3c/webextensions/pull/1012#issuecomment-4574797111

src/js/background/assignManager.js
mckenfra: This is intentional, to avoid showing the user the "Open this site in your assigned Container?" screen every time they browse to a new subdomain with an enabled catchall domain. I.e. if the user clicks "Remember my decision for this site", it then applies to all subdomains. We could change the wording in the prompt to "Remember my decision for all sites in this domain" when a catchall domain is in effect.

The alternative is to:

1. Show the prompt each time a new subdomain of an enabled catchall domain is visited (this may become annoying?)
2. If the user then clicks "Remember my decision for this site", a new site entry will need to be added in storage for that subdomain, to hold the "never ask" setting

src/js/background/assignManager.js
mckenfra: This is intentional, so that the user is only prompted "Open this site in your assigned Container?" once for an enabled catchall domain, and not every time a new subdomain is visited. See my comment here https://github.com/mozilla/multi-account-containers/pull/2912#discussion_r3679203875

src/js/background/assignManager.js
mckenfra: Well spotted, I had overlooked this. Just as in https://github.com/mozilla/multi-account-containers/pull/2912#discussion_r3679203875 , the same consideration applies here: do we re-prompt the user for every subdomain of an enabled catchall domain, or do we change the wording to "Remember my decision for all sites in this domain"?

If we re-prompt the user for every subdomain, then the implementation here becomes more complicated: we will need to store a list of subdomains to be explicitly *excluded* from an enabled catchall domain.

src/js/background/domainManager.js
mckenfra: We still need this, even when catchall-domains are disabled, to distinguish storage entries that are catchall-domains from those that are plain sites.

It is theoretically possible for a user to use different browser versions on different devices: one browser may support catchall-domains, the other may not. Due to syncing, a device that does not support catchall-domains may still have those catchall-domain entries in storage. Those entries need to be detectable so that they can be filtered out, otherwise they will end up being displayed in the UI as if they are plain sites. Your comment here has highlighted a bug in my implementation that this filtering-out is currently not done, and so I will need to update the `getDomainsAndAssignments()` method here: https://github.com/mozilla/multi-account-containers/pull/2912/changes#diff-1ffae9dff098547503a2394dc0bede4c2b4bcbbbdcea33cbbd9aea05d2a8d70fR66

FYI - the storage key format for catchall-domains and plain sites is intentionally identical, so that syncing of the storage entries is done automatically for both catchall-domains and plain sites. Storage key examples:

| Storage Key | Type |
| -:| - |
| `siteContainerMap@@_www.amazon.com` | Plain site |
| `siteContainerMap@@_*.amazon.com` | Catchall domain |

src/js/background/domainManager.js
mckenfra: Agreed, I will remove this method.

src/manifest.json
mckenfra: Noted, I will change this.

src/js/background/assignManager.js
mckenfra: Agreed: as with plain sites, this should abort if a container is missing. I will make the change.

However, in my view this existing code is dangerous: deleting assignments silently while a user is browsing. What if the container is temporarily unavailable due to a syncing issue or some other glitch?

It may be better to keep the assignments to the phantom container, and show the container in the UI with special highlighting, allowing the user to manually delete it, or else reinstate it.

Note that user assignments suddenly going missing is a commonly-reported bug with MAC, e.g.: https://github.com/mozilla/multi-account-containers/issues/2901 and also https://chat.mozilla.org/#/room/#containers:mozilla.org This silent deletion of assignments while browsing may be contributing to that issue.

However, I do note other bug reports about deleted containers reappearing, e.g. https://github.com/mozilla/multi-account-containers/issues/2682 While fixing that is important, my view is that doing so silently by deleting missing containers in `onBeforeRequest` is not the best solution.

src/js/background/domainManager.js
mckenfra: `getDomainMatchKeysFromName()` is used when determining if the user has enabled a catchall-domain for a site.

The problem this method addresses is that a site may have more than one possible public suffix, and therefore more than one possible catchall-domain. When determining if the user has enabled a catchall-domain for a site, we need to try all possible public suffixes.

E.g. the user browses to site `www.mysite.s3.us-west-1.amazonaws.com`:

| Public Suffix | Catchall Domain Key to Try |
| -:| -:|
| `s3.us-west-1.amazonaws.com` | `*.mysite.s3.us-west-1.amazonaws.com` |
| `com` | `*.amazonaws.com` |

If we don't check for all possible catchall-domains, a bug would manifest if the user:

1. Browses to `banana.amazonaws.com`
2. Adds it to container `Amazon`
3. Uses the UI to enable catchall-domain `*.amazonaws.com` in the same container
4. Browses to `www.mysite.s3.us-west-1.amazonaws.com` in a new tab
5. The longest catchall-domain for this site is `*.mysite.s3.us-west-1.amazonaws.com`, therefore a lookup of this key in storage would not find enabled catchall-domain `*.amazonaws.com`

The [publicSuffix API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/publicSuffix) provides a method for obtaining the _longest_ registrable domain for a site, but not _all possible_ registrable domains.

Therefore as a workaround, this method creates an array of all possible catchall-domain-keys by removing leading domain labels one-at-a-time. This array is then used for a single storage lookup, and the first match is used.

Note that the returned catchall-domain-key array may indeed contain redundant candidates such as `*.act.edu.au` and `*.edu.au`, however these will never match anything in storage since the UI provides no way for the user to actually add these.

Issue #2847, opened 2026-01-23
Site reappears in container site list a few hours after deleting it


devurandom:
### Before submitting a bug report

- [x] I updated to the latest version of Multi-Account Container and tested if I can reproduce the issue
- [x] I searched for existing reports to see if it hasn't already been reported

### Step to reproduce

1. On my work computer, I open the multi-account container settings > Manage Containers > {select my work container} > Manage Site List > {find accounts.google.com} > {click delete icon}
2. Wait a day or so
3. Go to "Manage Site List" again and find that accounts.google.com has magically reappeared

### Actual behavior

Something keeps adding accounts.google.com back to my work computer every time I delete it from that list / unassign it from containers.

### Expected behavior

Deleting a site from a container's list should keep it unassigned to that container.

### Additional informations

My problem could be related to these earlier reports:
* https://github.com/mozilla/multi-account-containers/issues/2110
* https://github.com/mozilla/multi-account-containers/issues/2600

But for me the site does not reappear in the list when I press the "sync" button in Firefox.  Something else appears to cause this to happen, but I'm unable to figure out what it is.

On my home computer I do not assign accounts.google.com to any container either.  The two computers' Firefoxes are synced (as well as my phone).

### Provide a copy of Troubleshooting Information page (optional)

```plain text
Application Basics
------------------

Name: Firefox
Version: 147.0.1
Build ID: 20260120081829
Distribution ID: fedora
User Agent: Mozilla/5.0 (X11; Linux x86_64; rv:147.0) Gecko/20100101 Firefox/147.0
OS: Linux 6.18.5-200.fc43.x86_64 #1 SMP PREEMPT_DYNAMIC Sun Jan 11 17:09:32 UTC 2026
OS Theme: Breeze / Adwaita
Multiprocess Windows: 8/8
Fission Windows: 8/8 Enabled by default
Remote Processes: 30
Enterprise Policies: Inactive
Google Location Service Key: Found
Google Safebrowsing Key: Found
Mozilla Location Service Key: Found
Safe Mode: false
Memory Size (RAM): 61,9 GB
Disk Space Available: 530 GB
```

greenaussie:
I'll add that the time period I see is much shorter than days - usually hours. Sync is enabled. May have started when FF profiles were re-vamped.

Issue #2098, opened 2021-08-07
Deleted containers keep coming back


hmlendea:
- Multi-Account Containers Version: 7.4.0
- Operating System + Version: Arch Linux
- Firefox Version: 90
- Other installed Add-ons + Version + Enabled/Disabled-Status:

### Actual behavior

Delete a container.
Wait anywhere between a few seconds and a few days, and they will be added back.

### Expected behavior

Delete a container.
It will never come back unless I manually add it again from scratch.

### Steps to reproduce
1. Open Manage Conteiners
2. Open a container
3. Delete that container
4. Confirm that it's gone
5. Check back later to see it is there again

### Notes

I've had more than a dozen containers in the past. And at some point a few months after I deleted all of them, they kept coming back, haunting me.
I've deleted them at least 10 times already. I even tried manually deleted all associated sites from them before deleting the containers themselves.
Now, just before writing this issue, I was deleting those containers, and just as I've been doing it, they've started reappearing.

I think this needs to be addressed quickly. It's very annoying and affects my work. It happened multiple times while sharing the screen with my coworkers that I had to stop what I was doing and disable the extension, because those containers interfered and caused pages to loop between containers, or in milder cases just had me have to log back into everything again.

hmlendea:
Note: Just when I pressed the submit button for this issue, the page opened in a new tab in a container I've just deleted before writing it, and had me log back into GitHub.

This is getting out of hand.

JR-bva:
I have the same problem. In addition, if I don't manually delete the container, a clone of it appears and then another and another. I left a container intentionally undeleted for about 3 days and got to about 70 identical containers. Which wasn't wise, because now I delete them manually :(

hmlendea:
@JR-bva the workaround I applied was to disable sync for the extension (although this makes it mighty annoying as I use Firefox across 3 different systems, sometimes 4).

And in regards to avoiding the manual deletion of dozens of containers, you can edit (Linux) `~/.mozilla/firefox/[profile-id]/containers.json` and make your changes there quickly and easily with a text editor. The catch is that there is surely another file that has those containers (e.g. storing the linked domains and other settings) but I can't find it.

kdmcguire:
Also experiencing this. Unfortunately I also used to use Temporary Containers, so I have 50+ old, deleted temporary containers (and duplicates) being brought back from the graveyard by synchronisation.

Have disabled container synchronisation for now.

[...]

Issue #1732, opened 2020-05-05
Containers persistently come back after deleting them


RudDemons:
- Multi-Account Containers Version: 6.2.5
- Operating System + Version: Windows 10 Pro 1909 (64 bits)
- Firefox Version: 75.0 (64 bits)
- Other installed Add-ons + Version + Enabled/Disabled-Status:

![Extensions](https://i.imgur.com/cav69Qr.png)

### Actual behavior
Containers keep reappearing after deleting them, in any device, making my container list gigantic and repetitive.

### Expected behavior
When deleting a container, keep them deleted and sync that information with my other devices.

### Steps to reproduce
1. Delete a container
2. Wait for sync to kick in
3. It comes back

kvannotten:
I seem to be having the same issue.

Solid-Ice8:
Repeat issue.  > #1708

jtagcat:
Duplicate of #1696

PR #2924, opened 2026-07-31
Fix deletion persistence bug (#2847, #1732, #2098, #2076, #2682) (AI)

ndrewtl:
# Description

I am a longtime user and fan of this project who has been haunted by issue #2847 for a long time. It also appears that issues #1732, #2098, #2076, and #2682 have the same root cause. I am a dev, but do not have experience with Firefox browser extensions. In a fit of frustration, I asked an LLM (Gemini 3.1 Pro) to attempt to diagnose and solve the issue. Gemini 3.1 diagnosed an issue– improper iteration over the `deletedSiteList` variable in `src/js/background/sync.js` that causes a key to remain in `browser.storage.sync`, while also being in `deletedSites`, which causes the previous configuration to 'resurrect' after syncing. I ran another LLM (GPT 5.6-Luna) over the codebase independently and it diagnosed the same issue.

The LLM proposed a five-line fix, which should be reviewable by any of the maintainers. I freely admit that I don't understand how this extension works, but I believe there is a good chance that this fix is correct, because

1. Two separate AI's provided very similar patches without collaborating. If they were confabulating, I would expect them to suggest different changes, not similar changes in the same part of the codebase.
2. The first AI provided a workaround that allowed me to successfully fix the bug via the Dev console, which prevented the rules from resurrecting. This strongly signals that the AI understands the root cause of the issue.

I have not run this code myself, as I'm not sure I understand even how to build this project. Therefore, it is possible that this patch is incorrect, which is why I would ask for any existing project maintainers to read it over and understand. I have also attached a longform transcript of my conversation with Gemini, for transparency's sake.

If this patch is genuinely incorrect, a reviewer should be able to reject it quite easily. On the other hand, if it is correct, it will ease the burden of many people who have struggled with the above-mentioned bugs, not least myself. I understand that project maintainers might not want to deal with AI slop. However I think the reward / tradeoff here is such that a very small amount of dev time might save many users from frustration.

I asked Gemini to describe its diagnosis and fix. What follows below is its summary of the content and change it applied.

# AI Summary

This PR addresses a long-standing and highly reported synchronization bug (the "Zombie Rules" bug) that causes deleted containers and site assignments to permanently resurrect themselves across devices, despite repeated deletions by the user.

### 1. The Root Cause of the Bug

The underlying cause is a state inconsistency triggered by the way the `deletedSiteList` is reconciled against active sync storage during concurrent or stale device updates.           
                                                                                                                                                                                        
When a user deletes a site assignment on Device A:                                                                                                                                    
                                                                                                                                                                                        
1. The site is added to the `deletedSiteList` on the sync server.                                                                                                                       
2. The site is removed from the active `browser.storage.sync` area.                                                                                                                     
                                                                                                                                                                                        
However, if Device B (which hasn't pulled this deletion yet) assigns or modifies a different rule before syncing:                                                                     
                                                                                                                                                                                        
1. Device B triggers `sync.storageArea.backup()`, which calls `updateSyncSiteAssignments()`.                                                                                              
2. `updateSyncSiteAssignments()` iterates through all of Device B's local storage and blindly re uploads its entire local state to browser.storage.sync.                                
3. Because Device B has not yet removed the original deleted rule locally, it accidentally re-uploads the deleted rule back to the sync server's active storage.                      
                                                                                                                                                                                        
At this point, the sync server is corrupted: The rule exists in BOTH the `deletedSiteList` and the active sync storage simultaneously.                                                  
                                                                                                                                                                                        
When `reconcileSiteAssignments()` runs on the next sync cycle (e.g., on browser restart):                                                                                               
                                                                                                                                                                                        
1. It processes `deletedSiteList` first, deleting the rule from local storage.                                                                                                          
2. In the very next block, it iterates over the active sync storage, finds the re-uploaded rule, and immediately restores it to local storage.                                        
3. This triggers another backup, keeping the rule alive forever.                                                                                                                      
                                                                                                                                                                                        
To make matters worse, `deleteSite()` had a premature early return: if a rule was already present in the `deletedSiteList`, the function aborted before calling      `sync.storageArea.area.remove()`. Because the rule was already trapped in the `deletedSiteList`, any subsequent attempts by the user to delete the resurrected rule would abort early and fail to remove the rule from active sync storage, cementing its zombie status.                                                                                                        

### 2. How this Fix works                                                                                                                                                             
                                                                                                                                                                                        
This PR introduces two logic changes to break the resurrection loop:                                                                                                                  
                                                                                                                                                                                        1. Explicit skipping in the restoration loop:                                                                                                                                         
   Inside `reconcileSiteAssignments()`, the loop that iterates over active sync storage now explicitly checks if the site is present in the `deletedSiteList`. If it is, the restoration is skipped. This ensures that even if a stale device accidentally re-uploads a deleted rule, the `deletedSiteList` explicitly overrides it, and the rule dies rather than propagating.     
2. Fixing the `deleteSite` early return:
   The early return in `deleteSite()` was modified. Even if a site is already in the `deletedSiteList`, the function now continues execution to ensure await `sync.storageArea.area.remove(siteStoreKey)` is called. This guarantees that deleting a zombie rule actually wipes it from the active sync storage, allowing the user to permanently kill rules that were previously trapped in this state.                                                                                                                                                     

### 3. Why this fixes the bug:

By enforcing that the `deletedSiteList` takes strict precedence over conflicting active assignments, we prevent stale devices from resurrecting deleted rules. By fixing the deletion function, we allow users to permanently clear corrupted rules from sync storage regardless of their historical presence in the deleted list.   

## Type of change

*Select all that apply.*

- [x] Bug fix
- [ ] New feature

* #2847
* #1732
* #2098 
* #2076 
* #2682

### DIFF
--- a/src/js/background/sync.js
+++ b/src/js/background/sync.js
@@ -43,9 +43,10 @@ const sync = {
     async deleteSite(siteStoreKey) {
       const deletedSiteList = 
         await sync.storageArea.getDeletedSiteList();
-      if (deletedSiteList.find(element => element === siteStoreKey)) return;
-      deletedSiteList.push(siteStoreKey);
-      await sync.storageArea.set({ deletedSiteList });
+      if (!deletedSiteList.find(element => element === siteStoreKey)) {
+        deletedSiteList.push(siteStoreKey);
+        await sync.storageArea.set({ deletedSiteList });
+      }
       await sync.storageArea.area.remove(siteStoreKey);
     },
 
@@ -533,6 +534,8 @@ async function reconcileSiteAssignments() {
   }
 
   for(const urlKey of Object.keys(assignedSitesFromSync)) {
+    if (deletedSiteList.find(element => element === urlKey)) continue;
+
     const assignedSite = assignedSitesFromSync[urlKey];
     try{
       if (assignedSite.identityMacAddonUUID) {


ndrewtl:
[transcript_full.jsonl.zip](https://github.com/user-attachments/files/30587524/transcript_full.jsonl.zip)

Full transcript

ndrewtl:
I am also attaching the workaround / fix that I applied without updating the code. This allowed me to identify and "hard delete" URL's that were locked to a particular site, and it did succeed. This might be useful to @devurandom as the originator of the bug I am experiencing. The following is all AI output.

### Diagnostic Script                                                                                                                                                                 

Open the Extension Console (about:debugging#/runtime/this-firefox -> Inspect on Multi-Account Containers -> Console tab) and paste this snippet:     
                           
```js                                                                                                                                                                      
    (async () => {                                                                                                                                                                      
      const syncData = await browser.storage.sync.get();                                                                                                                                
      const deletedSites = syncData.deletedSiteList || [];                                                                                                                              
      const zombies = [];                                                                                                                                                               
                                                                                                                                                                                        
      for (const key of Object.keys(syncData)) {                                                                                                                                        
        if (key.startsWith("siteContainerMap@@_") && deletedSites.includes(key)) {                                                                                                      
          zombies.push(key);                                                                                                                                                            
        }                                                                                                                                                                               
      }                                                                                                                                                                                 
                                                                                                                                                                                        
      if (zombies.length === 0) {                                                                                                                                                       
        console.log("✅ State is clean. No state inconsistencies detected.");                                                                                                           
      } else {                                                                                                                                                                          
        console.warn(`⚠️ Found ${zombies.length} inconsistent (zombie) rules!`);                                                                                                        
        zombies.forEach(zKey => {                                                                                                                                                       
            console.log("Zombie URL:", zKey.replace("siteContainerMap@@_", ""));                                                                                                        
        });                                                                                                                                                                             
                                                                                                                                                                                        
        console.log("\nTo fix all of them automatically, copy and run the following command:");                                                                                         
        console.log(`await browser.storage.sync.remove(${JSON.stringify(zombies)})`);                                                                                                   
      }                                                                                                                                                                                 
    })();
```                                                                                                                                                                          
                                                                                                                                                                                        
### How it works                                                                                                                                                                      

1. It downloads your entire browser.storage.sync database.                                                                                                                            
2. It cross-references every active site rule against the deletedSiteList.                                                                                                            
3. If it finds an overlap, it identifies it as a "zombie" rule that will continually resurrect itself due to the bug.                                                                 
4. If it finds any, it dynamically generates the exact command you need to run to fix all of them at once in a single batch operation.                                                
                                                                                                                                                                                        
You can run this snippet anytime you suspect an assignment didn't delete properly to ensure your sync state remains clean until the upstream bug fix is released.

### Fix: Delete the rule via Extension Console                                                                                                                                 
                                                                                                                                                                                        
1. In Firefox, open a new tab and navigate to about:debugging#/runtime/this-firefox                                                                                                   
2. Scroll down to find Firefox Multi-Account Containers in your list of extensions.                                                                                                   
3. Click the Inspect button next to it. This will open a Developer Tools window for the extension.                                                                                    
4. Go to the Console tab in the Developer Tools.                                                                                                                                      
5. You will need to know the exact domain name of the site (e.g., www.example.com or github.com). Run the following two commands in the console, pressing Enter after each (replace `example.com` with your actual domain):                                                                                                                                                 

```js                                                                                                                                                    
    await browser.storage.sync.remove("siteContainerMap@@_example.com")                                                                                                                 
                                                                                                                                                                                        
    await browser.storage.local.remove("siteContainerMap@@_example.com") 
```

PR #2928, opened 2026-08-19
Per-site assignment migration

bakulf:

# Description

Soon, in firefox we are going to have the per-site container assignment as a built-in feature. This patch migrates the existing association from MAC to firefox. Any new change will be kept in sync bidirectionally.

## Type of change

*Select all that apply.*

- [ ] Bug fix
- [x] New feature
- [x] Major change (fix or feature that would cause existing functionality to work differently than in the current version)

### DIFF
--- a/src/js/background/siteAssociation.js
+++ b/src/js/background/siteAssociation.js
@@ -1,0 +1,172 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/*
+ * Keeps the add-on's per-site assignments and Firefox's own per-site container
+ * associations (bug 2052136) in step. Firefox owns the truth: we import what it
+ * holds at every startup, hand over the assignments it doesn't have yet, and
+ * from then on the changes travel both ways as they happen.
+ */
+window.siteAssociation = {
+  STORAGE_PREFIX: "siteContainerMap@@_",
+
+  get supported() {
+    return !!(browser.contextualIdentities &&
+              browser.contextualIdentities.setSiteAssociation);
+  },
+
+  init() {
+    this.load().catch((e) => {
+      console.error("Could not load the site associations", e);
+    });
+  },
+
+  // Importing first is what makes Firefox win every disagreement.
+  async load() {
+    if (!this.supported) {
+      return;
+    }
+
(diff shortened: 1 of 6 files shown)


src/js/background/siteAssociation.js
```diff
@@ -0,0 +1,172 @@
...
+  // cookieStoreId is left out when the association was removed.
+  async onChanged({site, cookieStoreId}) {
+    if (cookieStoreId) {
+      await this.applyFromFirefox(site, cookieStoreId);
+      return;
+    }
+    await this.removeFromFirefox(site);
+  },
+
+  // neverAsk stays off for a site we don't know yet: the user never went
```
dannycolin, line 78: This makes sense in the current implementation. However, if a user installs MAC after using the builtin feature, they might be annoyed to see the confirmation page on many websites depending on the number of assignment they created before installing the addon.

Should we consider setting neverAsk to true for the site imported from Firefox to MAC in this case? An alternatie would be to move the confirmation page to the builtin implementation but that'd require more thinking (UX work). Thoughts?

src/js/background/siteAssociation.js
```diff
@@ -0,0 +1,172 @@
...
+    await assignManager.storageArea.remove(
+      siteStoreKey,
+      true, // shouldSync
+      true // fromFirefox
+    );
+  },
+
+  /*
+   * getSiteStoreKey concatenates hostname and port without a separator, so an
+   * assignment for https://example.com:8080 becomes "example.com8080" and can
```
dannycolin, line 122: We actually have a contributor who submitted a patch to fix this issue in MAC. https://github.com/mozilla/multi-account-containers/pull/2934 

Should we consider landing it alongside this one?

dannycolin:
LGTM after testing multiple scenarios to trigger the syncing mechanism between Firefox and MAC. 

The only thing I couldn't test is Firefox Sync to keep in sync the list of assignments across different devices/profiles. It is known to already create issues already so I'm not sure how much we want to investigate on that front right now.

dannycolin (approved):

src/js/background/siteAssociation.js
bakulf: The confirmation page is visible only if we decide to disable the per-site assignment feature in firefox. So, true/false does not impact users.

src/js/background/siteAssociation.js
bakulf: sounds good! Let me land this first and I review #2934 and merge it if ready.

PR #2931, opened 2026-08-23
Fix: Add "Default Container" to "Always open in..." menu to clear assignments

offvault:

# Description

Fixes a major UX friction point where users can easily get stuck in cross-container OAuth flows (such as Google Sign-In) because there is no explicit way to clear an "Always open in..." assignment from the popup or page action menu.

This PR adds a "Default Container" option to both the popup `always-open-in` menu and the `pageAction` menu. Clicking this option clears the site's container assignment and safely reloads the tab in the default container. 

**Implementation Details:**
- Added a `removeAlwaysOpenInContainer()` utility in `utils.js`.
- Prevented a race condition where the `webRequest` listener intercepts the newly created default tab before the assignment is cleared from storage. The rule is now cleanly removed *before* the new tab is created, and then the success message is routed to the new tab.
- Rendered the "Default Container" menu item identically to how it is handled in the "Reopen This Site in..." menu.
- Updated `assignment.test.js` and `940.test.js` to account for the new Default Container menu item to ensure tests continue passing.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Major change


dannycolin:
> This PR adds a "Default Container" option

By "default container", are you referring to the cookies & site data for regular tabs that aren't in a container?

